PR #3, opened 2019-07-13
add circleci config

tsloughter:
Circle really needs a way to set an executor for every job in a workflow... but for now, this is it.


tsloughter:
@SergeyKanzhelev I don't have permissions to enable CircleCI, are you able to?

PR #4, opened 2019-07-25
inital api impl

tsloughter:
Opening as a draft PR to get the conversation started (also related to the convo @garthk started in https://github.com/opencensus-beam/opencensus_elixir/issues/23)

There are currently two tracers here: `oc_tracer_sdk` and oc_tracer_sdk_dyn`. The former uses the pdict context and ets storage spans directly while the latter can have the context and storage set with configuration.

I'm not married to any of this.

### DIFF
--- a/include/opentelemetry.hrl
+++ b/include/opentelemetry.hrl
@@ -1,0 +1,137 @@
+%%%------------------------------------------------------------------------
+%% Copyright 2019, OpenTelemetry Authors
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
+%%
+%% http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing, software
+%% distributed under the License is distributed on an "AS IS" BASIS,
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+%%%------------------------------------------------------------------------
+
+%% These records are based on protos found in the opentelemetry-proto repo:
+%% src/opentelemetry/proto/trace/v1/trace.proto
+%% They are not exact translations because further processing is done after
+%% the span has finished and can be vendor specific. For example, there is
+%% no count of the number of dropped attributes in the span record. And
+%% an attribute's value can be a function to only evaluate the value if it
+%% is actually used (at the time of exporting).
+
+%% for use in guards: sampling bit is the first bit in 8-bit trace options
+-define(IS_SPAN_ENABLED(X), (X band 1) =/= 0).
+
+-define(MESSAGE_EVENT_TYPE_UNSPECIFIED, 'TYPE_UNSPECIFIED').
+-define(MESSAGE_EVENT_TYPE_SENT, 'SENT').
+-define(MESSAGE_EVENT_TYPE_RECEIVED, 'RECEIVED').
+
+-define(SPAN_KIND_UNSPECIFIED, 'SPAN_KIND_UNSPECIFIED').
+-define(SPAN_KIND_INTERNAL, 'INTERNAL').
+-define(SPAN_KIND_SERVER, 'SERVER').
+-define(SPAN_KIND_CLIENT, 'CLIENT').
+-define(SPAN_KIND_PRODUCER, 'PRODUCER').
+-define(SPAN_KIND_CONSUMER, 'CONSUMER').
+
+-record(span_ctx, {
+          %% 128 bit int trace id
+          trace_id          :: opentelemetry:trace_id() | undefined,
+          %% 64 bit int span id
+          span_id           :: opentelemetry:span_id() | undefined,
+          %% 8-bit integer, lowest bit is if it is sampled
+          trace_options = 1 :: integer() | undefined,
+          %% Tracestate represents tracing-system specific context in a list of key-value pairs.
+          %% Tracestate allows different vendors propagate additional information and
+          %% inter-operate with their legacy Id formats.
+          tracestate        :: opentelemetry:tracestate() | undefined,
+          %% IsValid is a boolean flag which returns true if the SpanContext has a non-zero
+          %% TraceID and a non-zero SpanID.
+          is_valid          :: boolean() | undefined
+         }).
+
+-record(span, {
+          %% name of the span
+          name                                    :: unicode:unicode_binary(),
+
+          %% 128 bit int trace id
+          trace_id                                :: opentelemetry:trace_id() | undefined,
+
+          %% 64 bit int span id
+          span_id                                 :: opentelemetry:span_id() | undefined,
+          %% 64 bit int parent span
+          parent_span_id                          :: opentelemetry:span_id() | undefined,
+
+          tracestate                              :: opentelemetry:tracestate() | undefined,
+
+          %% 8-bit integer, lowest bit is if it is sampled
+          trace_options = 1                       :: integer() | undefined,
+
+          kind = ?SPAN_KIND_UNSPECIFIED           :: opentelemetry:span_kind() | undefined,
+
+          start_time                              :: wts:timestamp(),
+          end_time                                :: wts:timestamp() | undefined,
+
+          %% A set of attributes on the span.
+          %% Kept as a list so ets:select_replace/2 can be used to add new elements
+          attributes = []                         :: opentelemetry:attributes() | undefined,
+
+          %% optional stacktrace from where the span was started
+          stack_trace                             :: opentelemetry:stack_trace() | undefined,
+
+          %% A time-stamped annotation or message event in the Span.
+          time_events = []                        :: opentelemetry:time_events(),
+
+          %% links to spans in other traces
+          links = []                              :: opentelemetry:links(),
+
+          %% An optional final status for this span.
+          status                                  :: opentelemetry:status() | undefined,
+
+          %% An optional resource that is associated with this span. If not set, this span
+          %% should be part of a batch that does include the resource information, unless resource
+          %% information is unknown.
+          resource                                :: opentelemetry:resource() | undefined,
+
+          %% A highly recommended but not required flag that identifies when a trace
+          %% crosses a process boundary. True when the parent_span belongs to the
+          %% same process as the current span.
+          same_process_as_parent_span = undefined :: boolean() | undefined,
+
+          %% An optional number of child spans that were generated while this span
+          %% was active. If set, allows implementation to detect missing child spans.
+          child_span_count = undefined            :: integer() | undefined
+         }).
+
+-record(link, {
+          trace_id                  :: opentelemetry:trace_id(),
+          span_id                   :: opentelemetry:span_id(),
+          attributes                :: opentelemetry:attributes(),
+          tracestate                :: opentelemetry:tracestate()
+         }).
+
+-record(message_event, {
+          %% type of MessageEvent. Indicates whether the RPC message was sent or received.
+          type = 'TYPE_UNSPECIFIED' :: opentelemetry:message_event_type(),
+
+          %% identifier for the message, which must be unique in this span.
+          id                        :: integer(),
+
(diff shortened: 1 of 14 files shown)


tsloughter:
Related docs:

* [Library Guidelines](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/library-guidelines.md)
* [OpenTelemetry Specification](https://github.com/open-telemetry/opentelemetry-specification/blob/5935d5193ea77e2f91951726bb723aafe3e11437/specification/tracing-api.md#span-creation)

include/opentelemetry.hrl
```diff
@@ -0,0 +1,140 @@
...
+%% the span has finished and can be vendor specific. For example, there is
+%% no count of the number of dropped attributes in the span record. And
+%% an attribute's value can be a function to only evaluate the value if it
+%% is actually used (at the time of exporting).
+
+%% for use in guards: sampling bit is the first bit in 8-bit trace options
+-define(IS_SPAN_ENABLED(X), (X band 1) =/= 0).
+
+
+-define(SPAN_TAB, otel_span_table).
```
hauleth: Should this table be public? Wouldn't it be better to have additional header file within `src` directory to contain private macros?

src/opentelemetry.erl
```diff
@@ -0,0 +1,124 @@
...
+%% @end
+%%--------------------------------------------------------------------
+-spec generate_span_id() -> span_id().
+generate_span_id() ->
+    uniform(2 bsl 63 - 1). %% 2 shifted left by 63 == 2 ^ 64
+
+%% Before OTP-20 rand:uniform could not give precision higher than 2^56.
+%% Here we do a compile time check for support of this feature and will
+%% combine multiple calls to rand if on an OTP version older than 20.0
+-ifdef(OTP_RELEASE).
```
hauleth: Which versions of OTP will we support?

src/opentelemetry_app.erl
```diff
@@ -21,10 +21,23 @@
...
 start(_StartType, _StartArgs) ->
+    maybe_init_ets(),
     opentelemetry_sup:start_link().
 
 stop(_State) ->
     ok.
 
 %% internal functions
+
+maybe_init_ets() ->
```
hauleth: Isn't it it discouraged to make supervisor process containing ETS table? Wouldn't it be better to create supervision tree for each segment of the application?

src/ot_tracer.erl
```diff
@@ -0,0 +1,66 @@
...
+         start_span/2,
+         with_span/1,
+         current_span_ctx/0,
+         finish/0,
+         get_binary_format/0,
+         get_http_text_format/0]).
+
+-include("opentelemetry.hrl").
+
+-define(tracer, (persistent_term:get({opentelemetry, tracer}, ot_tracer_sdk))).
```
hauleth: As there is usage of `persistent_term` then we support only OTP 21+, which mean that earlier part with fallback for old OTPs.

hauleth: Will there always be only one tracer? I haven't checked Java implementation yet.

src/opentelemetry_app.erl
tsloughter: Yea, I'm going to move this into an `init` under a supervisor that will make using ets span storage essentially optional. Will do that soon to clear that up.

src/ot_tracer.erl
tsloughter: That is the default tracer. There will only be one default tracer.

hauleth:
There is heavy use of `persistent_term`, shouldn't most of that be moved to application configuration?

include/opentelemetry.hrl
tsloughter: This will go into a module actually once I do the supervisor ets thing I mention in another comment.

tsloughter:
I use persistent term because it needs to have as low overhead as possible. I'm even uneasy using persistent term for this as it also still entails a slow function call because of not being a static module name in the code.

We can have it compile to an application config if on earlier OTPs and just document this fact.

hauleth:
I think that it will be a pretty low overhead when comparing to other languages anyway.

src/opentelemetry_app.erl
tsloughter: The supervisor stuff isn't complete but I did do a quick move of the init to a process for `ot_span_ets`.

garthk:
Quite. We're wrapping spans around calls with 10²–10⁵ms latency at work. No need to worry about per-span overheads in the 10⁻²ms  range, e.g. `Application.get_env/3` like I once wasted too much time trying to optimise around.

Not having lerned me some Erlang, can I assume from Elixir we can `@behaviour :opentelemetry.ot_tracer`? If so: _nice_, you've given us one configurable behaviour from which our users can sync with Logger metadata, rummage through `Process.get(:"$callers")`, muck about with `:seq_trace`, whatever, and in any order they like.

src/opentelemetry.erl
garthk: OTP20 will cover Elixir programmers going back to 1.5 as long as the reason they're using an old Elixir isn't because they need an even older OTP.

Use of the `:seq_trace` label to track spans will be tricky in OTP20 without hoping that the birthday paradox won't give us collisions: it's limited to 28 bits, giving us a 17% chance of a collision per 10K spans. Might not hurt as long as they expire fast, but it'll be a lot easier in OTP21.

garthk:
Re-reading your comment in the intro, and the code: is the idea that the default tracer won't be pluggable, so it can go _extra_ fast, but users will be able to plug in the `_dyn` version instead, and it'll let them plug in alternative implementations for the explicit `ot_ctx` behaviour and implicit `ot_span_ets` behaviour? Feels a little convoluted, so I hope I'm wrong.

src/ot_ctx.erl
```diff
@@ -0,0 +1,22 @@
...
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+%%
+%% @doc
+%% @end
+%%%-------------------------------------------------------------------------
+-module(ot_ctx).
+
+-callback get(any()) -> any().
```
garthk: Q: if the key will always be `?SPAN_CTX`, do we need the `Key` arguments?

A: it'll provide more or less backward-compatible rummaging through the process dictionary (`ot_span_ctx_key` vs `oc_span_ctx_key`) with `ot_ctx_pdict`, and give people a plausible way of sharing the `seq_trace` label with `ot_ctx_seqtrace`.

src/ot_ctx.erl
```diff
@@ -0,0 +1,22 @@
...
+-callback with_value(any(), any()) -> ok.
+-callback with_value(any(), any(), fun()) -> ok.
```
garthk: `-> any().`, surely, as it'll return the result of `fun/0`?

src/ot_ctx.erl
garthk: Tried writing against this API. Straight away, needed to special-case `:oc_span_ctx_key` so I  knew to try other means than just `Process.get/1`. Strikes me it'd be simpler for API consumers if `ot_ctx` didn't take keys, and if the `seq_trace` implementation relied on a lower level API to accomplish plausible sharing.

garthk: Tripped over this trying to wrap up the work above: 

Special-casing the span context key to try unreliable but useful ways of recovering it, e.g. wading through `$callers`, could also cause problems for the context restoration feature of the `get_current_ctx` … `start_span` … `with_span_ctx` … `finish_span` … `with_span_ctx` pattern, which would end up putting back a context that wasn't really there.

If our goal is to let users track the current span context however they need to, strikes me the behaviour should have `put_span_ctx/1`, `get_span_ctx/0`, and `recover_span_ctx/0`. I'll try it out and report back.

Nearly forgot `Logger.metadata` sync et al, which'd need a separate behaviour `sync_current_span_ctx/2` called every time we change the current span context or its attributes.

tsloughter:
> Re-reading your comment in the intro, and the code: is the idea that the default tracer won't be pluggable, so it can go _extra_ fast, but users will be able to plug in the `_dyn` version instead, and it'll let them plug in alternative implementations for the explicit `ot_ctx` behaviour and implicit `ot_span_ets` behaviour? Feels a little convoluted, so I hope I'm wrong.

That was the original idea but I think you are right that it is too convoluted.

src/ot_ctx.erl
tsloughter: I think the `ot_tracer_sdk_dyn` will become `ot_tracer_sdk`. Then ctx will be dynamic like tracer. This should resolve your issues, right?

tsloughter:
Pushed updates. I moved the `_dyn` tracer to be the main sdk module and thus the default as well. 

It also now tracks the current tracer used in the pdict: https://github.com/open-telemetry/opentelemetry-erlang/pull/4/files#diff-98920f865a632de900785a223c5e42ccR53

This way if you create a span with a different tracer from the default it will use that tracer instead of the default when the span is finished (or is updated or a child is created).

tsloughter:
I've marked this is ready for review so we can get it merged in before even more gets added in and it gets more of a pain to follow.

tsloughter:
@hauleth @garthk any objections to me merging this? It doesn't mean anything is set in stone but I think it is a good start and allows for people to work on and send PRs for various features (like the seqtrace context isn't implemented yet and samplers are needed) separately.

src/opentelemetry_sup.erl
```diff
@@ -37,11 +37,12 @@ start_link() ->
...
 %%                  modules => modules()}   % optional
-init([]) ->
-    SupFlags = #{strategy => one_for_all,
+init([Opts]) ->
+    SupFlags = #{strategy => one_for_one,
                  intensity => 0,
                  period => 1},
-    ChildSpecs = [],
+    ChildSpecs = [#{id => ot_span_ets,
+                    start => {ot_span_ets, start_link, [Opts]}}],
```
hauleth: As for sure we will expand this I would create `ot_span_sup` that would handle just spans and would manage everything related to them. This would simplify `opentelemetry_sup` in future.

tsloughter: Do you mean having `opentelemetry_sup` start `ot_span_sup` which starts `ot_span_ets`? Do you expect there to be other span related children?

src/ot_ctx.erl
```diff
@@ -0,0 +1,23 @@
...
+%%
+%% @doc
+%% @end
+%%%-------------------------------------------------------------------------
+-module(ot_ctx).
+
+-callback get(term()) -> term().
+-callback get(term(), term()) -> term().
+-callback with_value(term(), term()) -> ok.
+-callback with_value(term(), term(), fun()) -> ok.
```
hauleth: I think that all calls to the callbacks should be handled there. Elvis will be disappointed when making dynamic call to callbacks not defined in current module.

tsloughter: It will? That doesn't make sense, it is just defining a behaviour. 

I guess instead of `?ctx:get` we can have it be `ot_ctx:get` which is implemented as `get(Term) -> ?ctx:get(Term)`, just seemed like overkill.

src/opentelemetry_sup.erl
hauleth: @tsloughter I do not know how we want to implement dispatching spans to the handler(s). Will there be another process like in OC?

tsloughter: Ah, I guess I was considering reporters and such as separate from span management.

src/ot_ctx.erl
hauleth: Elvis do not like dynamic calls to the behaviour callbacks outside given behaviour module. I am 100% sure about that. Also such `ot_ctx:get/1` IMHO would make other modules less "magical" and would not require sharing `?ctx` macro.

tsloughter: I've updated it. And yea, it wasn't shared before because it was only done in `ot_tracer_sdk`. Definitely wouldn't want it having to be shared. But this is good now I think.

src/opentelemetry_sup.erl
tsloughter: There is the sweeper too. Yea, I'll make this change. One sec.

PR #6, opened 2019-08-15
add benchmarking example

tsloughter:


tsloughter:
This required some fixes to rebar_mix and rebar3 on my setup. It may work as is for others, it is something complaining about my elixir that caused a crash.

samples/ot_benchmarks.erl
```diff
@@ -0,0 +1,29 @@
...
+
+-export([run/0]).
+
+run() ->
+    Iterations = 100,
+    PDictCtxFun = fun() ->
+                      [ot_tracer:start_span(<<"span-", (integer_to_binary(X))/binary>>)
+                       || X <- lists:seq(1, 100)]
+                  end,
+    'Elixir.Benchee':run(#{<<"pdict_ctx">> => PDictCtxFun}),
```
hauleth: Benchee defines also [`benchee` module](https://github.com/bencheeorg/benchee/blob/bd5185bcfdc8320c237959505e63a89d31e044a9/lib/benchee.ex#L16), so you can use `benchee:run(#{<<"pdict_ctx">> => PDictCtxFun})` to be more Erlangy.

tsloughter: Ah, nice.

PR #7, opened 2019-08-17
refactor span modules and add span impl setup callback

tsloughter:


tsloughter:
I've updated this to configure the span and context implementations through the setup of `ot_tracer_sdk`.

One other change I'm considering is rename `ot_span` and ot_ctx` to `ot_span_api` and `ot_ctx_api` respectively to signal they are just apis and not implementations.

src/opentelemetry.app.src
```diff
@@ -8,7 +8,8 @@
     stdlib,
     wts
    ]},
-  {env,[]},
+  {env, [{ot_tracer, #{ot_span_impl => {ot_span_ets, []},
```
hauleth: Why use `ot_` prefix for configuration option? I would use just `tracer`. Also that `_impl` suffix leaking to the end user is less that perfect IMHO.

src/opentelemetry_app.erl
```diff
@@ -23,7 +23,14 @@
 
 start(_StartType, _StartArgs) ->
     Opts = application:get_all_env(opentelemetry),
-    opentelemetry_sup:start_link(Opts).
+    {ok, Pid} = opentelemetry_sup:start_link(Opts),
+
+    %% if the span impl needs to have a process supervised it must be
+    %% setup after the supervision tree has started.
+    {ot_tracer, TracerOpts} = lists:keyfind(ot_tracer, 1, Opts),
+    ot_tracer_sdk:setup(TracerOpts),
```
hauleth: Shouldn't this be ran **before** starting supervisor itself?

src/ot_ctx.erl
```diff
@@ -17,30 +17,7 @@
...
-get(Key, Default) ->
-    ?ctx:get(Key, Default).
-
--spec with_value(term(), term()) -> ok.
-with_value(Key, Value) ->
-    ?ctx:with_value(Key, Value).
-
--spec with_value(term(), term(), fun()) -> ok.
-with_value(Key, Value, Fun) ->
-    ?ctx:with_value(Key, Value, Fun).
```
hauleth: Are we removing that functions once again?

src/ot_span_ets.erl
```diff
@@ -18,16 +18,24 @@
 %%%-------------------------------------------------------------------------
 -module(ot_span_ets).
 
+-behaviour(ot_span).
 -behaviour(gen_server).
```
hauleth: Do we need both behaviours? Wouldn't it be better to hide from the implementer the fact that there is any process involved?

src/ot_span_utils.erl
```diff
@@ -0,0 +1,111 @@
...
+    Links = maps:get(links, Opts, []),
+    Kind = maps:get(kind, Opts, ?SPAN_KIND_UNSPECIFIED),
+
+    %% TODO: support overriding the sampler
+    _Sampler = maps:get(sampler, Opts, undefined),
+
+    new_span_(Name, Parent, Kind, Attributes, Links).
+
+%% if parent is undefined, first run sampler
+new_span_(Name, undefined, Kind, Attributes, Links) ->
```
hauleth: I do not know how about others, but I find it a little bit confusing to have `_` suffix for private functions. Especially when there is no public function with the same name and arty.

tsloughter: Oh, that is a mistake since there isn't a function of the same name/arity that is public.

src/opentelemetry.app.src
tsloughter: Yea, `ot_` prefix is wrong. As for `_impl`, probably better to just go with `ot_span`, true.

src/opentelemetry_app.erl
tsloughter: It can't if it is to be allowed to have supervision by the application.

src/ot_span_ets.erl
tsloughter: The process is specific to only this implementation.

src/ot_ctx.erl
tsloughter: Yes, I talk about it in the issue I created. If the context and span are configurable outside the scope of the tracer it is confusing. The tracer is what gets configured to use specific context and span implementations.

src/opentelemetry_app.erl
tsloughter: Hm, though the return for the tracer setup function could be a list of children to start...

seancribbs: Yeah, I would do that... use the configuration to detemine whether to start children of the top-level sup.

src/ot_span_ets.erl
```diff
@@ -68,6 +75,34 @@ finish_span(#span_ctx{span_id=SpanId,
 finish_span(_) ->
     ok.
 
+-spec get_ctx(opentelemetry:span()) -> opentelemetry:span_ctx().
+get_ctx(_Span) ->
+    #span_ctx{}.
```
seancribbs: I'm guessing these are just stubs?

tsloughter:
I think I need to rename some modules.

The reason there is any `_sdk` module is that technically there should be both an `opentelemetry_api` and `opentelemetry_sdk` application. Instead of doing that my plan was to have them both in the same application but identified through the suffix.

But it definitely causes confusion for anyone looking at the code that isn't coming directly from the OTel spec -- and maybe even causes confusion in that case.

Instead I'll add to the readme an explanation of how the OTel concepts of API and SDK map to our codebase and for example `ot_tracer_sdk` will become `ot_tracer_default`.

src/opentelemetry_app.erl
tsloughter: Yea, issue was you should be able to enable tracers at runtime. But I can add a function that handles that for the user allowing the setup function to still return specs.

src/ot_span_ets.erl
tsloughter: Right.

tsloughter:
@hauleth @seancribbs I made a number of updates. I think it is getting closer...

src/ot_span_ets.erl
ferd: Could the stubs return anything significant?
I'm looking at the utils module and all that data seems like it should be available anyway?

src/ot_tracer_default.erl
```diff
@@ -15,48 +15,67 @@
...
+-define(span, (persistent_term:get(?SPAN_IMPL_KEY))).
 
 -type pdict_trace_ctx() :: {opentelemetry:span_ctx(), pdict_trace_ctx() | undefined}.
 
+-spec setup(map()) -> [supervisor:child_spec()].
+setup(Opts) ->
+    lists:filtermap(fun({ConfigKey, PersistentKey}) ->
+                        {Module, Args} = maps:get(ConfigKey, Opts),
+                        persistent_term:put(PersistentKey, Module),
+                        case erlang:function_exported(Module, start_link, 1) of
```
ferd: Risks of the module not being loaded? I guess it doesn't matter for releases.

src/ot_span_utils.erl
```diff
@@ -0,0 +1,111 @@
...
+update_trace_options(should_sample, #span_ctx{trace_id=_TraceId,
+                                              span_id=_ParentSpanId,
+                                              trace_options=_ParentTraceOptions}) ->
+    1.
+    %% case oc_sampler:should_sample(TraceId, ParentSpanId, ?IS_ENABLED(ParentTraceOptions)) of
+    %%     true ->
+    %%         1;
+    %%     false ->
+    %%         0
+    %% end.
```
ferd: what's with the commented code?

tsloughter: This is code from OpenCensus. I had left it commented because I wasn't sure if it would be this or a later PR which would implement this (thus uncommenting the code once the necessary arguments were passed through). It is handled in the sampler PR.

tsloughter:
@ferd hm, what behaviour is "stubbed out"? Do you mean like `ot_span_ets` and other placeholders?

ferd:
> @ferd hm, what behaviour is "stubbed out"? Do you mean like `ot_span_ets` and other placeholders?

Yeah. They have like the create/stop span functions defined, the rest is just ignoring all arguments and returning `ok`.

tsloughter:
Yea, just placeholders to have the behaviour callbacks covered but haven't yet implemented the logic.

tsloughter:
Ugh, it seems to not be in sync with circleci which now has all jobs in the config green.

src/ot_tracer_default.erl
hauleth: Still, it can be problematic during development.

tsloughter:
Closing and reopening because circle and github won't sync properly.

PR #10, opened 2019-08-31
samplers

tsloughter:
Also depends on #7 so to view just what the sampler changes are look at the last commit in this PR.

Unlike in opencensus this is using funs for samplers and not modules. I think this makes storing state, like the `IdUpperBound` in the probability sampler, simpler and is more efficient to call a fun than if we did a variable module function call.

### DIFF
--- a/include/opentelemetry.hrl
+++ b/include/opentelemetry.hrl
@@ -22,7 +22,7 @@
 %% is actually used (at the time of exporting).
 
 %% for use in guards: sampling bit is the first bit in 8-bit trace options
--define(IS_SPAN_ENABLED(X), (X band 1) =/= 0).
+-define(IS_SPAN_ENABLED(TraceOptions), (TraceOptions band 1) =/= 0).
 
 -define(MESSAGE_EVENT_TYPE_UNSPECIFIED, 'TYPE_UNSPECIFIED').
 -define(MESSAGE_EVENT_TYPE_SENT, 'SENT').
@@ -41,14 +41,20 @@
           %% 64 bit int span id
           span_id           :: opentelemetry:span_id() | undefined,
           %% 8-bit integer, lowest bit is if it is sampled
-          trace_options = 1 :: integer() | undefined,
+          trace_flags = 1 :: integer() | undefined,
           %% Tracestate represents tracing-system specific context in a list of key-value pairs.
           %% Tracestate allows different vendors propagate additional information and
           %% inter-operate with their legacy Id formats.
           tracestate        :: opentelemetry:tracestate() | undefined,
           %% IsValid is a boolean flag which returns true if the SpanContext has a non-zero
           %% TraceID and a non-zero SpanID.
-          is_valid          :: boolean() | undefined
+          is_valid          :: boolean() | undefined,
+          %% true if the span context came from a remote process
+          is_remote         :: boolean() | undefined,
+          %% this field is not propagated and is only here as an implementation optimization
+          %% If true updates like adding events are done on the span. The same as if the
+          %% trace flags lowest bit is 1 but simply not propagated.
+          is_recorded       :: boolean() | undefined                                
          }).
 
 -record(span, {
@@ -101,7 +107,12 @@
 
           %% An optional number of child spans that were generated while this span
           %% was active. If set, allows implementation to detect missing child spans.
-          child_span_count = undefined            :: integer() | undefined
+          child_span_count = undefined            :: integer() | undefined,
+
+          %% this field is not propagated and is only here as an implementation optimization
+          %% If true updates like adding events are done on the span. The same as if the
+          %% trace flags lowest bit is 1 but simply not propagated.
+          is_recorded       :: boolean() | undefined                                
          }).
 
 -record(link, {
--- a/src/opentelemetry_app.erl
+++ b/src/opentelemetry_app.erl
@@ -26,10 +26,15 @@ start(_StartType, _StartArgs) ->
 
     %% if the span impl needs to have a process supervised it must be
     %% setup after the supervision tree has started.
-    {tracer, {Tracer, TracerOpts}} = lists:keyfind(tracer, 1, Opts),
-    Children = ot_tracer:setup(Tracer, TracerOpts),
+    {sampler, {Sampler, SamplerOpts}} = lists:keyfind(sampler, 1, Opts),
+    SamplerFun = ot_sampler:setup(Sampler, SamplerOpts),
 
-    opentelemetry_sup:start_link(Children, Opts).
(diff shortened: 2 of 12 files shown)


deadtrickster:
Hi, thanks for taking care of this. How to configure custom sampler then?

tsloughter:
@deadtrickster right now you can provide any module name and it'll call `setup` on it which must return a fun https://github.com/open-telemetry/opentelemetry-erlang/pull/10/files#diff-8e30ccebd796115d6bb0634079ad39b0R108

I can, and probably should, add an option to just set a function in the config, so that if the `Sampler` is a function of appropriate arity it uses that.

src/opentelemetry_app.erl
```diff
@@ -23,7 +23,18 @@
 
 start(_StartType, _StartArgs) ->
     Opts = application:get_all_env(opentelemetry),
-    opentelemetry_sup:start_link(Opts).
+
+    %% if the span impl needs to have a process supervised it must be
+    %% setup after the supervision tree has started.
```
ferd: I think this whole comment is off.

Should be something like

```erlang
%% Any sampler implementation must be passed to the tracer chosen.
%% since the tracer is started after the rest of the supervision tree has started
%% and we only get its children, we must instantiate the sampler before anything else.
%% The sampler turns out to be instantiated before any supervision tree is started.
```
or something to that effect?

src/ot_sampler.erl
```diff
@@ -0,0 +1,126 @@
...
+                        opentelemetry:links(),
+                        opentelemetry:span_name(),
+                        opentelemetry:kind(),
+                        opentelemetry:attributes()) -> sampling_result()).
+-export_type([sampling_result/0,
+              sampling_decision/0,
+              sampler/0]).
+
+-define(MAX_VALUE, 9223372036854775807).
+-define(DEFAULT_PROBABILITY, 0.5).
```
ferd: Any logic behind these choices?

src/ot_sampler.erl
```diff
@@ -0,0 +1,126 @@
...
+                    case do_probability_sample(TraceId, IdUpperBound) of
+                        ?RECORD_AND_PROPAGATE ->
+                            {?RECORD_AND_PROPAGATE, []};
+                        ?NOT_RECORD ->
+                            %% sampling hint could still be ?RECORD
+                            maybe_sampling_hint(SamplingHint, IgnoreHints)
+                    end;
+                false ->
+                    %% sampling hint could still be ?RECORD
+                    maybe_sampling_hint(SamplingHint, IgnoreHints)
```
ferd: Sounds like you could keep:

```erlang
...
false ->
    case OnlyRoot =:= false andalso IsRemote =:= true
             andalso do_probability_sample(...) of
        ?RECORD_AND_PROPAGATE -> ...
        _ -> maybe_sampling_hint(...)
    end
end
```
Essentially cheating to use the do_probability_sample/2 in the trailing position of shortcircuiting-booleans allows to group the two negative clauses into a single nexting and keep them in sync.

Not sure if that's something you feel would be clearer.

src/ot_sampler.erl
```diff
@@ -0,0 +1,126 @@
...
+setup(probability, Opts) ->
+    IdUpperBound = 
+        case maps:get(probability, Opts, ?DEFAULT_PROBABILITY) of
+            P when P =:= 0.0 ->
+                0;
+            P when P =:= 1.0 ->
+                ?MAX_VALUE;
+            P when P >= 0.0 andalso P =< 1.0 ->
+                P * ?MAX_VALUE
+        end,
```
ferd: so if I get it right, the span IDs are random, and this is just setting a probability value. Since a RNG is used for the span IDs, this method allows to do sampling without having to call the RNG more than once per span? Clever.

test/otel_samplers_SUITE.erl
```diff
@@ -0,0 +1,106 @@
+-module(otel_samplers_SUITE).
```
ferd: any reason for using otel_ rather than just ot_ for this?

include/opentelemetry.hrl
```diff
@@ -41,14 +41,20 @@
...
           %% IsValid is a boolean flag which returns true if the SpanContext has a non-zero
           %% TraceID and a non-zero SpanID.
-          is_valid          :: boolean() | undefined
+          is_valid          :: boolean() | undefined,
+          %% true if the span context came from a remote process
+          is_remote         :: boolean() | undefined,
+          %% this field is not propagated and is only here as an implementation optimization
+          %% If true updates like adding events are done on the span. The same as if the
+          %% trace flags lowest bit is 1 but simply not propagated.
+          is_recorded       :: boolean() | undefined
```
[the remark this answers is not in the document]
tsloughter: False means that the span is basically ignored.

src/opentelemetry_app.erl
tsloughter: Hm, yea, I think that comment was from an old version. Will fix.

src/ot_sampler.erl
tsloughter: Max is the largest 64 bit integer, I guess I can put a comment to that effect. 50% I just made up. I'll check if there is a default in the spec.

src/ot_sampler.erl
tsloughter: Problem is that case statement returns a boolean and not one of the 3 possible record values.

tsloughter: Oh, derp, no it doesn't, forgot it returns the result of the last expression. Perfect.

src/ot_sampler.erl
tsloughter: Right. Though trace id not span id. That is also why it has to hold some extra state, the calculated value.

test/otel_samplers_SUITE.erl
tsloughter: Hm, no, not sure how that happened. Though I have considered using `otel` as the prefix everywhere.

src/ot_sampler.erl
hauleth: There is no such sampler defined in specs. The specs define only 2 samplers:

- always
- never

All others aren't specified at all.

tsloughter: There is, the rfc was approved https://github.com/open-telemetry/oteps/blob/master/text/0006-sampling.md#default-samplers

hauleth: I really need to catch up with OT…

bogdandrutu:
@tsloughter I pressed the update branch to make sure you pass tests :)

bogdandrutu:
@tsloughter and now I am forcing you to fix the builds 👍

PR #15, opened 2019-09-12
Enable Elvis linting

hauleth:
It is easier to start with linting when we still have not-that-much code.


tsloughter:
Approved in principle, obv can't merge until it passes. Some of getting it to pass will involve configuring elvis rules. Like the max line length should be 120.

tsloughter:
Still needs the linting fixed and CLA signed so we can merge.

hauleth:
I signed it

PR #16, opened 2019-09-13
Move remote calls to handler to the behaviour modules

hauleth:
This will make changing the scope of the handlers behaviours easier.
I.e. currently this removes 2 required behaviour callbacks from
the ot_ctx as it can be easily implemented as shared implementation.

### DIFF
--- a/src/ot_ctx.erl
+++ b/src/ot_ctx.erl
@@ -22,19 +22,27 @@
          with_value/3,
          with_value/4]).
 
--callback get(term(), term()) -> term().
--callback with_value(term(), term()) -> ok.
+-type key() :: term().
+-type instrumented_fun() :: fun(() -> term()).
 
--spec get(Impl :: module(), term()) -> term().
+-export_type([key/0, instrumented_fun/0]).
+
+%% Get value of `Key' from current context and return `Default' if none.
+-callback get(Key :: key(), Default :: term()) -> term().
+
+%% Set value of `Key' in current context to `Value'.
+-callback with_value(Key :: key(), Value :: term()) -> ok.
+
+-spec get(Impl :: module(), key()) -> term().
 get(Module, Key) -> get(Module, Key, undefined).
 
--spec get(Impl :: module(), term(), term()) -> term().
+-spec get(Impl :: module(), key(), term()) -> term().
 get(Module, Key, Default) -> Module:get(Key, Default).
 
--spec with_value(Impl :: module(), term(), term()) -> term().
+-spec with_value(Impl :: module(), key(), term()) -> term().
 with_value(Module, Key, Value) -> Module:with_value(Key, Value).
 
--spec with_value(Impl :: module(), term(), term(), fun()) -> term().
+-spec with_value(Impl :: module(), key(), term(), instrumented_fun()) -> term().
 with_value(Module, Key, Value, Fun) ->
     Orig = get(Module, Key),
     try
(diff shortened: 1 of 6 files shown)


src/ot_ctx.erl
```diff
@@ -17,7 +17,29 @@
...
+
+-spec get(Impl :: module(), term(), term()) -> term().
+get(Module, Key, Default) -> Module:get(Key, Default).
+
+-spec with_value(Impl :: module(), term(), term()) -> term().
+with_value(Module, Key, Value) -> Module:with_value(Key, Value).
+
+-spec with_value(Impl :: module(), term(), term(), fun()) -> term().
+with_value(Module, Key, Value, Fun) ->
+    Orig = get(Module, Key),
```
tsloughter:
```suggestion
    Orig = ?MODULE:get(Module, Key),
```

I'm not sure if this is actually required when there is a locally available `get/2` function but it is clearer.

tsloughter:
I'm personally not a fan of adding another function call to the callstack, though I shouldn't care, the cost of the `Module:fun` will way out way the cost of the new function around it. So this is probably fine.

Interested in @ferd  and @deadtrickster thoughts.

And the CLA needs to be signed, can you do that so this can pass CI and be merged?

src/ot_ctx.erl
```diff
@@ -17,7 +17,29 @@
...
+         get/3,
+         with_value/3,
+         with_value/4]).
+
 -callback get(term(), term()) -> term().
 -callback with_value(term(), term()) -> ok.
--callback with_value(term(), term(), fun()) -> ok.
+
+-spec get(Impl :: module(), term()) -> term().
+get(Module, Key) -> get(Module, Key, undefined).
```
ferd: This is unsafe as an interface since you can't know for sure whether the value returned is `undefined` or missing.

If you do an API with a default value, it either needs to be set explicitly, crash when the value is missing, or return `{ok, Value} | undefined`.

I know this is just repeating the older API, but if we're breaking its location for something new, might as well fix it?

src/ot_ctx.erl
```diff
@@ -17,7 +17,29 @@
...
+
+-spec get(Impl :: module(), term(), term()) -> term().
+get(Module, Key, Default) -> Module:get(Key, Default).
+
+-spec with_value(Impl :: module(), term(), term()) -> term().
+with_value(Module, Key, Value) -> Module:with_value(Key, Value).
+
+-spec with_value(Impl :: module(), term(), term(), fun()) -> term().
```
ferd: there should really be types defined for these, like:

``` 
-type key() :: term().
-type overridden_value() :: term().
-type instrumented_fun() :: fun(() -> term()).

-spec with_value(module(), key(), overridden_value(), instrumented_fun()) -> term().
``` 
Right now I'm unsure what these types mean and there's no docs, so I'm just guessing the meaning.

hauleth: @ferd it was just migration of old callbacks, but I agree. There should be more meaningful types used instead of vague `term()`.

tsloughter: They are `term()` because a context is just a box that can hold anything.

src/ot_ctx.erl
tsloughter: `undefined` means the same thing whether it is explicitly put there or because there is no value set for the context.

tsloughter:
@hauleth can you sign off on the CLA?

hauleth:
I signed it

PR #18, opened 2019-10-17
Reporters

tsloughter:
Opening as a draft because sweeper isn't done and I want that in along with this.

But it is a major change on how reporters are run that I wanted to open it as soon as possible to get feedback.

Note that it still uses 2 ets tables and swaps between them. I think we should consider doing an `ets:delete(Table)` in the process that runs the actual reporters and then creating a new table (with the same name as the deleted one).

I'd be curious to find out if there is any issue with creating/deleting tables that often and using the same name. Like there is a hash table of some sort that stores the mapping of table name to table reference, I don't know the implementation behind that yet and don't want to find out later it ends up collecting garbage if you use it in such a pattern.


tsloughter:
I've updated the changes to delete the table after it finishes with reporting and create a new one.

src/ot_reporter.erl
```diff
@@ -0,0 +1,250 @@
...
+                                       send_interval_ms=SendInterval}) ->
+    {OldTableName, RunnerPid} = report_spans(Data),
+    {keep_state, Data#data{runner_pid=RunnerPid,
+                           handed_off_table=OldTableName},
+     [{state_timeout, ReportingTimeout, reporting_timeout},
+      {{timeout, report_spans}, SendInterval, report_spans}]};
+reporting(state_timeout, reporting_timeout, Data) ->
+    %% kill current reporting process because it is taking too long
+    Data1 = kill_runner(Data),
+    {repeat_state, Data1};
```
ferd: So if I get this right, whenever we hit a state timeout, it's because the runner took too long. Since the table has been transferred to it, this interrupts reporting and flushes the old table via killing the process, then runs again and automatically reports on the new table since the swap has been done on the initial setup?

You risk however not having a new reporting table ready because the new table is created upon report completion and it has never been set up in this case. So you'll start reporting on the new current table with no replacement, and if that one times out too, you end up with no tables available at all?

test/ot_sweeper_SUITE.erl
```diff
@@ -0,0 +1,175 @@
...
+            ?assertMatch({ST, O} when is_integer(ST)
+                                      andalso is_integer(O), S#span.start_time),
+            ?assertMatch({ST, O} when is_integer(ST)
+                                      andalso is_integer(O), S#span.end_time)
+    after
+      1000 -> ct:fail("Do not received any message after 1s")
+    end,
+
+    %% sleep long enough that the reporter would have run again for sure
+    timer:sleep(10),
```
ferd: you can drop that sleep by using it in the `after` of the `receive` instead and probably speed up tests

tsloughter: It won't speed it up since the "ok" clause in receive is the `after 0`.

src/ot_reporter.erl
tsloughter: Fixed in latest comment.

test/ot_sweeper_SUITE.erl
ferd: It would speed up the happy path if you get the result in 1ms (no need to sleep 10ms)

tsloughter: It would speed up the bad path. The happy path is `after 0`. But yea, may as well change the happy path to `after 10` I guess.

PR #20, opened 2019-11-07
wip: restructure tracer around registry

tsloughter:
This is to be split into separate OTP applications to make the
distinction clearer. For now the SDK registry is started when
opentelemetry application is started. The registry is responsible
for tracking the tracers, the sampler and span processors.

The registry also currently sets a ctx and span module to use. I
think this will change when context is split out.  I'd also like
to consider not having the span module be configurable in the
SDK but always use ot_span_ets and if the user needs a fancy form
of span handling they can replace the SDK if they can't get by
with using span processors.

### DIFF
--- a/src/opentelemetry.app.src
+++ b/src/opentelemetry.app.src
@@ -8,11 +8,11 @@
     stdlib,
     wts
    ]},
-  {env, [{tracer, {ot_tracer_default, #{span => {ot_span_ets, []},
-                                        ctx => {ot_ctx_pdict, []}}}},
-         {sampler, {always_on, #{}}},
-         {exporter, [{exporters, []},
-                     {scheduled_delay_ms, 30000}]}]},
+  {env, [{sampler, {always_on, #{}}},
+         {processors, [%% #{id => my_processor,
+                       %%   module => ot_batch_processor,
+                       %%   config => #{}}
+                        ]}]},
   {modules, []},
 
   {licenses, ["Apache 2.0"]},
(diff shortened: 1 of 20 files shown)


tsloughter:
Currently getting a tracer, done for the user in `ot_tracer` if they don't do it themselves, returns a module name by calling the registry -- a registry is what is currently called a factory in the api https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-tracing.md#obtaining-a-tracer

The sampler is only pulled from storage when/if needed, same with span processors.

Another option is to have `get_tracer` return a map or record like:

```erlang
#{module => module(),
   processors => [module()],
   sampler => fun()}
```

Updates to processor and sampler should be rare, if ever, after startup (hence why they are currently persistent_terms), so that should be fine, it just means data to copy around instead of just looking up what we need when we need it.

tsloughter:
The map/record might make more sense once resources are added.. The tracer should have an associated resource https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/sdk-resource.md

tsloughter:
Also might want to get rid of any persistent_term usage that isn't an atom, just to be safe. Move configuration like processors and sampler to ETS.

Though I worry the "free-ness" of an atom persistent_term may cause me to overuse them as well. Part of why I sort of want to do something else with the ctx and span modules.

tsloughter:
Closing. I'll open a new PR that use the new separate app for API.

PR #21, opened 2019-11-10
split the API into a separate application

tsloughter:
This patch moves the API modules to another application, opentelemetry_api.

This repo is now the SDK implementation and when booted will start
the ot_tracer_server which is used to serialize configuration of
the tracer and set the default_tracer through the API.


tsloughter:
How the context is used still needs to be updated but I wanted to open this for feedback since it is already a large patch to move shit out to the API repo.

PR #22, opened 2019-11-12
fix probability sampler

44044:
the probability sampler does not work as expected, because when starting a new span, a span_ctx is passed to sampler as parent instead of undefined.


tsloughter:
Thanks. Can you also fix the typespec line 32 should be `opentelemetry:span_ctx() | undefined` now. Other option would have been to have the probability sampler recognize a span_ctx with span_id of undefined means no parent but this is good too.

Just need you to sign off on the CLA so this can be merged.

tsloughter:
Thanks. This was merged as part of #24

PR #24, opened 2019-11-20
fix probability sampler

tsloughter:
a parent span context with span id of 'undefined' must be passed
as the atom 'undefined' to the sampler for the logic to work. This
patch updates the typespec for samplers to reflect this.


tsloughter:
This is a copy of change from #22 which the author hasn't signed the CLA and I'd like to get this merged so I can cut a first release.

PR #27, opened 2019-11-30
W3C interop tests

tsloughter:
Currently not a part of the CI. I was planning to do something like `rebar3 as interop shell -s w3c_trace_context_interop run` but that isn't an option currently, so it is manual.

Second commit moves the creation of the on_span/on_end function to the creation of the tracer instead of doing it every time they are needed.

### DIFF
--- a/src/ot_propagation_http_w3c.erl
+++ b/src/ot_propagation_http_w3c.erl
@@ -24,13 +24,18 @@
 
 -include_lib("opentelemetry_api/include/opentelemetry.hrl").
 
--define(VERSION, "00").
+-define(VERSION, <<"00">>).
 
 -define(ZERO_TRACEID, <<"00000000000000000000000000000000">>).
 -define(ZERO_SPANID, <<"0000000000000000">>).
 
 -define(HEADER_KEY, <<"traceparent">>).
 -define(STATE_HEADER_KEY, <<"tracestate">>).
+
+-define(KEY_MP, element(2, re:compile("^[a-z0-9][a-z0-9_*/-]{0,255}$|^([a-z0-9_*/-]{1,241})(@[a-z0-9_*/-]{1,14})$"))).
+-define(VALUE_MP, element(2, re:compile("^[ -~]{0,256}$"))).
+
+-define(MAX_TRACESTATE_PAIRS, 32).
 
 -spec inject(ot_propagation:http_headers(),
                  {opentelemetry:span_ctx(), opentelemetry:span_ctx() | undefined} | undefined)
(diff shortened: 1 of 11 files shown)


src/ot_propagation_http_w3c.erl
```diff
@@ -90,44 +99,73 @@ combine_headers(Key, Headers) ->
...
         error:badarg ->
             undefined
-    end;
-decode(_) ->
+    end.
+
+tracestate_decode(Value) ->
+    parse_pairs(string:lexemes(Value, [$,])).
+
+parse_pairs(Pairs) when length(Pairs) < 33 ->
```
ferd: is 32 like a max depth limit? If so might be useful to stick it as a macro.

test/opentelemetry_SUITE.erl
```diff
@@ -121,17 +121,20 @@ propagation(_Config) ->
     EncodedTraceId = io_lib:format("~32.16.0b", [TraceId]),
     EncodedSpanId = io_lib:format("~16.16.0b", [SpanId]),
 
-    ?assertListsMatch([{<<"X-B3-TraceId">>, EncodedTraceId},
-                       {<<"X-B3-SpanId">>, EncodedSpanId},
-                       {<<"X-B3-Sampled">>, "1"},
+    ?assertListsMatch([%% {<<"X-B3-TraceId">>, EncodedTraceId},
+                       %% {<<"X-B3-SpanId">>, EncodedSpanId},
+                       %% {<<"X-B3-Sampled">>, "1"},
```
ferd: worth dropping commented-out code?

test/opentelemetry_SUITE.erl
```diff
@@ -121,17 +121,20 @@ propagation(_Config) ->
...
+                       {<<"traceparent">>,
+                        [<<"00">>, "-", EncodedTraceId,"-", EncodedSpanId, "-", <<"01">>]},
```
ferd: so 00 appears to be the version, and 01 are the opts, but what do opts mean in this case?

src/ot_propagation_http_w3c.erl
```diff
@@ -90,44 +99,73 @@ combine_headers(Key, Headers) ->
...
   when TraceId =:= ?ZERO_TRACEID orelse SpanId =:= ?ZERO_SPANID ->
     undefined;
-decode(<<?VERSION, "-", TraceId:32/binary, "-", SpanId:16/binary, "-", Opts:2/binary, _/binary>>) ->
+decode(<<Version:2/binary, "-", TraceId:32/binary, "-", SpanId:16/binary, "-", Opts:2/binary>>)
+  when Version >= ?VERSION andalso Version =/= <<"ff">> ->
+    to_span_ctx(Version, TraceId, SpanId, Opts);
+%% future versions could have more after Opts, so allow for a trailing -
+decode(<<Version:2/binary, "-", TraceId:32/binary, "-", SpanId:16/binary, "-", Opts:2/binary, "-", _/binary>>)
+  when Version > ?VERSION andalso Version =/= <<"ff">> ->
+
```
ferd: whitespace messed up

src/ot_propagation_http_w3c.erl
```diff
@@ -90,44 +99,73 @@ combine_headers(Key, Headers) ->
...
+        %% verify version is hexadecimal
+        _ = binary_to_integer(Version, 16),
         #span_ctx{trace_id=binary_to_integer(TraceId, 16),
                   span_id=binary_to_integer(SpanId, 16),
-                  trace_flags=case Opts of <<"01">> -> 1; _ -> 0 end}
+                  trace_flags=case Opts of <<"01">> -> 1; <<"00">> -> 0; _ -> error(badarg) end}
     catch
         %% to integer from base 16 string failed
         error:badarg ->
             undefined
```
ferd: you also get this exceptions for options that are neither 00 nor 01. Is this also still safe?

tsloughter: Yea, that is why I had it error with badarg, either way needs to result in an `undefined` context.

src/ot_propagation_http_w3c.erl
tsloughter: Yea, good point, will move to a macro.

test/opentelemetry_SUITE.erl
tsloughter: 01 opts just mean it is an enabled trace.

PR #32, opened 2020-01-09
remove baggage. it was removed from otep-66

tsloughter:


rebar.lock
```diff
@@ -1,7 +1,7 @@
 {"1.1.0",
 [{<<"opentelemetry_api">>,
   {git,"https://github.com/open-telemetry/opentelemetry-erlang-api",
-       {ref,"48da6489cc45d59a42a7d9c68616144c10b36e3b"}},
```
hauleth: We still need to find a way to store these two in single monorepo. AFAIK Mix supports sparse Git checkouts, maybe it would be worth adding it to the Rebar3 as well.

tsloughter: Yea, and that is how I want to support it in rebar3. I've briefly looked into doing it but it is going to be a decent bit of work.

PR #35, opened 2020-01-13
use record for tracing tracer context instead of 2-tuple

tsloughter:


src/ot_propagation_http_b3.erl
```diff
@@ -21,19 +21,19 @@
...
 
--spec inject(ot_propagation:http_headers(),
-             {opencensus:span_ctx(), opentelemetry:span_ctx() | undefined} | undefined)
+-spec inject(ot_propagation:http_headers(), tracer_ctx() | undefined)
             -> ot_propagation:http_headers().
-inject(_, {#span_ctx{trace_id=TraceId,
-                     span_id=SpanId}, _}) when TraceId =:= 0
-                                               ; SpanId =:= 0 ->
+inject(_, #tracer_ctx{active=#span_ctx{trace_id=TraceId,
+                                       span_id=SpanId}}) when TraceId =:= 0
```
hauleth: Any reason why not use `#span_ctx{trace_id=0,span_id=0}`?

tsloughter: Hmm... weird, hehe, I can't think of one, must have been some refactoring mistake :). Fixing now.

tsloughter: Oh, nevermind, because it is `or` and not `and`.

hauleth: Just in case, there is also second place that could use it.

hauleth: Oh, yeah. The Erlang approach with using `;` as `or` and `,` as `and` in guards is confusing.

PR #38, opened 2020-01-15
Updates for OTEP 66

tsloughter:
This brings the implementation of OTEP 66 (https://github.com/open-telemetry/oteps/pull/66) into line with the latest, and soon to be merged, design. Now propagated context is kept in a separate variable. This is especially good so there isn't the possibility to end local span and end up then updating/ending what is actually a remote span: https://github.com/open-telemetry/oteps/blob/3f5ef37f28a0a38a252c3c6b669f42c2dcb7ef79/text/0066-separate-context-propagation.md#extract

### DIFF
--- a/src/ot_tracer_default.erl
+++ b/src/ot_tracer_default.erl
@@ -30,6 +30,7 @@
 %% tracer access functions
 -export([span_module/1]).
 
+-include_lib("opentelemetry_api/include/opentelemetry.hrl").
 -include("ot_tracer.hrl").
 
 %% the context namespace key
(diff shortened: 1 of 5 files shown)


src/ot_tracer_default.erl
```diff
@@ -30,42 +30,46 @@
 %% tracer access functions
 -export([span_module/1]).
 
+-include_lib("opentelemetry_api/include/opentelemetry.hrl").
```
ferd: Just a thing I noticed:

since the record definitions are done in the `opentelemetry_api` OTP app, and that we're going to see that one included into all kinds of dependencies far more often than the current app, we should expect far more frequent conflicts in versions if the internal structure is to ever change if anyone else is importing the .hrl file.

I don't know if it'd be interesting to find a way to generate a warning when someone who is not opentelemetry is including the file and sneaking a peek at the records' structure somehow. I.e. maybe defining a macro in erl_opts for this lib and the API, and if that macro isn't defined, we cause some error or issue through preprocessor macros and bad declarations (write an unused function like `this_include_file_not_intended_for_non_opentelemetry_apps() -> ok.` conditionally)

tsloughter: Argh, I was about to say, "no, it shouldn't be needed", but it is here `SPAN_KIND_*` macros are defined.

We may just need to split this header.

I don't believe anyone but a SDK should have to include this header. And as I type that I realize, also exporters. Grr.

But regular instrumented apps should not use the records directly, ever.

tsloughter: Oh, but you want to warn against including it since it shouldn't be included?

I misread that as you worrying about how people had to include it but versions of it could change.

That would be interesting if we could warn not to use internal structures. I'd also wonder if we could move the records to the SDK to be safe. I think in the API they are currently just used for patching in function headers to know if an argument is of that record type.

ferd: That at least reduces the scope, yeah. You shouldn't have 50 apps depending on various exporters, and as long as the record definition is macro'd and not literal, then the compiler will work it fine. The problem is any explicit matching on records at the call site can cause major compatibility issues.

ferd: Yeah, whichever, the idea is to make sure nobody hardcodes a record's structure in a way that will cause incompatibilities in a top-level project including it all. External function calls are similar but would crash at runtime only, but we can't do much for these.

tsloughter: Yea, it is probably fine because of the limited use it is required in. But never know what people will try to do in their apps.

PR #42, opened 2020-01-22
lookup and update only the events or attributes changing in ot_span_ets

tsloughter:
We can save on copying by only looking up and updating the particular
element being changed, instead of looking up the whole span and doing
a compare and swap.

Spans are referenced through the current active span context and thus
modified only by a single process, so concurrent updates of the same
field are not a real concern.


tsloughter:
Ah, yup, will add that comment to each of the functions.

Issue #28, opened 2019-12-01
Implement Resources


tsloughter:
Resources can be associated with a tracer and can be added to individual spans. We need functions for creating/merging the resource maps and for retrieving them for a tracer/span which gets used mostly in exporters.

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/sdk-resource.md

PR #43, opened 2020-01-25
otel resource implementation, including resources from OS vars

tsloughter:


tsloughter:
I still need to actually have it tie the resource to the tracer provider on startup, as well as prepare a way for third party resource dependencies (like an application that creates resource entries based on kubernetes metadata).

But I wanted to open this partly because of the way I used `{resource, map()}` as the resource return and an opaque type for resource.

The idea being in part so that calls to `merge` must be with `create`ed resources instead of maps that haven't yet been verified for size and type of entries.

tsloughter:
Closes #28

hauleth:
Cannot that be fetched from the application environment? Or the specs require environment variable?

tsloughter:
It is part of the spec.

We can and should add one that reads from the application env, but this ensures simply setting a environment variable on any environment running an otel instrumented application will get the resource values.

src/ot_resource.erl
```diff
@@ -0,0 +1,64 @@
...
+
+%%
+
+%% all resource strings, key or value, must be latin1 with length less than 255
+check_string(S) ->
+    length(S) =< ?MAX_LENGTH andalso io_lib:deep_latin1_char_list(S).
+
+%% a resource key must be a non-empty latin1 string
+check_key(K) ->
+    is_list(K) andalso length(K) > 0 andalso check_string(K).
```
ferd: with this and the check_string() call we'll be going over the list twice for the length.
```
check_key(K = [_|_]) -> check_string(K);
check_key(_) -> false.
```
would save a bunch of iterations.

src/ot_resource_env_var.erl
```diff
@@ -0,0 +1,44 @@
...
+%%
+
+parse_resource_labels(false) ->
+    #{};
+parse_resource_labels(RawLabels) ->
+    Labels = string:split(RawLabels, ?LABEL_LIST_SPLITTER, all),
+    lists:foldl(fun(Label, Acc) ->
+                        case string:split(Label, ?LABEL_KEY_VALUE_SPLITTER, all) of
+                            [K, V] ->
+                                V1 = re:replace(string:trim(V), "^\"|\"$", "", [global, {return, list}]),
```
ferd: can the value contain any characters that would otherwise confuse the splitting:

```
1> string:split("a=\"b=c\"", "=", all).
["a","\"b","c\""]
```

src/ot_resource.erl
tsloughter: Good point, will change. But wonder if the compiler is able to do anything smart with length comparisons to not have to count to the to Nth element if the result of length isn't used...

src/ot_resource_env_var.erl
tsloughter: I thought not but now I can't find the spec and the Java implementation contradicts itself. It says it can have `=` in the value but then the implementation doesn't support that https://github.com/open-telemetry/opentelemetry-java/blob/24e470522d744c1249cbf818deab4a26d580b3e4/sdk/src/main/java/io/opentelemetry/sdk/resources/EnvVarResource.java#L59

I'll have to verify and fix this up if it is true it can have other characters in the value.

tsloughter:
Updated and added a module for getting resource attributes from the application env.

tsloughter:
Will fix dialyzer in the morning.

src/ot_resource_env_var.erl
tsloughter: After talking to some people the allowed chars in a value is still not defined anywhere, so nothing we can do about it for now. But may have to update to not naively split.

PR #52, opened 2020-03-13
metrics sdk first iteration

tsloughter:
aggregation is instrument type specific until views are added to the spec.

releasing bound metrics does nothing and won't until the active instrument
table tracks if updates have occured within the last collection interval
for each active instrument and deletes them after 2 intervals of no
updates. At that point binding and releasing will need to be tracked
along with if there have been updates so it is known if an active instrument
is actually inactive and should be removed from the table or not.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -1,7 +1,8 @@
 {erl_opts, [debug_info]}.
 {deps, [{wts, "~> 0.4"},
-        {opentelemetry_api, {git, "https://github.com/tsloughter/opentelemetry-erlang-api",
-                             {branch, "metrics-api"}}}]}.
+        {opentelemetry_api,
+         {git, "https://github.com/open-telemetry/opentelemetry-erlang-api",
+          {branch, "master"}}}]}.
 
 {shell, [{apps, [opentelemetry]},
          {config, "config/sys.config"}]}.
--- a/rebar.config
+++ b/rebar.config
@@ -1,7 +1,7 @@
 {erl_opts, [debug_info]}.
 {deps, [{wts, "~> 0.4"},
-        {opentelemetry_api, {git, "https://github.com/open-telemetry/opentelemetry-erlang-api",
-                             {branch, "master"}}}]}.
+        {opentelemetry_api, {git, "https://github.com/tsloughter/opentelemetry-erlang-api",
+                             {branch, "metrics-api"}}}]}.
 
 {shell, [{apps, [opentelemetry]},
          {config, "config/sys.config"}]}.
--- a/src/ot_meter_default.erl
+++ b/src/ot_meter_default.erl
@@ -132,6 +132,11 @@ observe(_, _, _) ->
     ok.
 
 init(_Opts) ->
+    %% TODO: we do not want to lose instrument and observer tables ever
+    %% eventually need to have an heir to take them if this process crashes.
+    %% Another option is to just use persistent_term since these things
+    %% don't change after creation.
+
     %% ets table is required for other parts to not crash so we create
     %% it in init and not in a handle_continue or whatever else
     case ets:info(?TAB, name) of
--- a/src/ot_meter_default.erl
+++ b/src/ot_meter_default.erl
@@ -1,0 +1,184 @@
+%%%------------------------------------------------------------------------
+%% Copyright 2019, OpenTelemetry Authors
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
+%%
+%% http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing, software
+%% distributed under the License is distributed on an "AS IS" BASIS,
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+%%
+%% @doc
+%% @end
+%%%-------------------------------------------------------------------------
+-module(ot_meter_default).
+
+-behaviour(ot_meter).
+-behaviour(gen_server).
+
+-export([start_link/1,
+         new_instruments/2,
+         lookup_instrument/1,
+         record/4,
+         record_batch/3,
+
+         %% functions used for bound instruments
+         record/3,
+         bind/3,
+         release/2,
+
+         %% observer functions
+         observer_tab/0,
+         register_observer/3,
+         set_observer_callback/3,
+         observe/3]).
+
+-export([init/1,
+         handle_call/3,
+         handle_cast/2]).
+
+-include("ot_meter.hrl").
+
+-define(OBSERVER_TAB, ot_metric_accumulator_observers).
+
+-define(TAB, ?MODULE).
+
+-record(state, {}).
+
+start_link(Opts) ->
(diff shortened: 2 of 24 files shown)


tsloughter:
This is the simplest version of the metrics SDK and I believe fulfills the spec.

The spec does not yet have views, so aggregations are not configurable.

src/ot_meter_default.erl
```diff
@@ -0,0 +1,184 @@
...
+    {ok, #state{}}.
+
+handle_call({new, List}, _From, State) ->
+    Result = ets:insert_new(?TAB,
+      [#instrument{name=Name,
+                   description=maps:get(description, I, <<>>),
+                   kind=MetricKind,
+                   input_type=maps:get(input_type, I, integer),
+                   unit=maps:get(unit, I, one),
+                   label_keys=maps:get(label_keys, I, [])} || I=#{name := Name,
```
ferd: the instrument record defines 8 fields with no `undefined` in the typespec, but this only initializes 6. From an eye check this will piss dialyzer off.

src/ot_metric_accumulator.erl
```diff
@@ -0,0 +1,179 @@
...
+init(_Opts) ->
+    %% ets table is required for other parts to not crash so we create
+    %% it in init and not in a handle_continue or whatever else
+    case ets:info(?ACTIVE_TAB, name) of
+        undefined ->
+            _ = ets:new(?ACTIVE_TAB, [named_table,
+                                      public,
+                                      {keypos, #active_instrument.key},
+                                      {write_concurrency, true},
+                                      ordered_set]);
```
ferd: since this is a named process, I would expect the named table to always succeed since its creation relies on the process being able to register its name already. You probably don't need to do a check for liveness there in practice, and if you could, then you would risk running in a race condition between the check and creation anyway.

src/ot_metric_aggregator_mmsc.erl
```diff
@@ -0,0 +1,96 @@
...
+
+-spec update(ets:tab(), ot_meter:key(), ot_meter:input_type(), number()) -> boolean().
+update(Tab, Key, _, Number) ->
+    update_mmsc(Tab, Key, Number).
+
+-spec checkpoint(ets:tab(), ot_meter:key()) -> boolean().
+checkpoint(Tab, NameLabelSet) ->
+    MS = ets:fun2ms(fun(A=#active_instrument{key=Key,
+                                             aggregator=Aggregator,
+                                             current=Current}) when Key =:= NameLabelSet ,
```
ferd: you may need to add the `_ = '_'` key for the rest not to turn into default values

src/ot_metric_aggregator_mmsc.erl
```diff
@@ -0,0 +1,96 @@
...
+    {infinity, 0.0, 0.0, 0};
+initial_value(_) ->
+    {infinity, 0, 0, 0}.
+
+%%
+
+update_mmsc(Tab, NameLabelSet, Number) ->
+    %% can't use min/max functions in matchspec body so stuck with this
+    MS = ets:fun2ms(fun(A=#active_instrument{key=Key,
+                                             current={infinity, 0, 0, 0}}) when Key =:= NameLabelSet ->
```
ferd: same for all of these

src/ot_metric_aggregator_sum.erl
```diff
@@ -0,0 +1,85 @@
...
+    select_replace(Tab, Key, Number);
+update(_, _, _, _) ->
+    false.
+
+-spec checkpoint(ets:tab(), {ot_meter:name(), ot_meter:label_set()}) -> boolean().
+checkpoint(Tab, NameLabelSet) ->
+    MS = ets:fun2ms(fun(A=#active_instrument{key=Key,
+                                             aggregator=Aggregator,
+                                             instrument=#instrument{input_type=integer},
+                                             current=Current}) when Key =:= NameLabelSet ->
```
ferd: again here

src/ot_metric_exporter_stdout.erl
```diff
@@ -0,0 +1,26 @@
...
+%% @doc
+%%
+%% @end
+%%%-------------------------------------------------------------------------
+-module(ot_metric_exporter_stdout).
+
+-export([export/1]).
+
+export(Records) when map_size(Records) > 0 ->
+    io:format("Records ~p~n", [Records]);
```
ferd: that's a sad format!

src/ot_meter_default.erl
tsloughter: 2 fields were removed in my related API PR.

I should really finish my rebar3 sparse checkout PR :(, this is a pain having these in separate repos.

rebar.config
```diff
@@ -1,7 +1,7 @@
 {erl_opts, [debug_info]}.
 {deps, [{wts, "~> 0.4"},
-        {opentelemetry_api, {git, "https://github.com/open-telemetry/opentelemetry-erlang-api",
-                             {branch, "master"}}}]}.
+        {opentelemetry_api, {git, "https://github.com/tsloughter/opentelemetry-erlang-api",
+                             {branch, "metrics-api"}}}]}.
```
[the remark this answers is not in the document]
tsloughter: Yea, once the other PR is merged.

src/ot_metric_accumulator.erl
tsloughter: Yea. I left it there with the intent of deciding if it needed to set an heir and handoff/return. But I think for this table its fine to drop if a crash happens instead of putting a lot of effort into trying to save metrics that would be lost, so I'll remove this.

src/ot_metric_exporter_stdout.erl
tsloughter: Hah, yea, eventually want to copy the format they are doing in Go.

src/ot_metric_aggregator_mmsc.erl
tsloughter: Oh, hm, I thought it automatically handled that because of the `A=` followed by use of `A#`.

tsloughter: Testing I don't see it setting them to defaults.

ferd: I didn't think the assignment played any role on this and am surprised this is apparently done by default now. I guess it's fine that way and it's for match_object that it would have been a problem?

tsloughter: Yea, I think you are thinking of match_object where it is the case.

tsloughter:
Dialyzer won't pass until https://github.com/open-telemetry/opentelemetry-erlang/pull/56 is approved and merged.

tsloughter:
@ferd is this good enough for now and can be +1'ed? I won't be merging until https://github.com/open-telemetry/opentelemetry-erlang/pull/56 is approved and merged so I can update the lock file (and then dialyzer will pass).

tsloughter:
@ferd yea, I can add an annotation. But are you saying like something might be expecting to use this output as input to shovel metrics? This exists for testing purposes, so locally someone can say, "yup, its outputting, now I can figure out why my actual exporter isn't working since I know thats not the issue."

ferd:
> @ferd yea, I can add an annotation. But are you saying like something might be expecting to use this output as input to shovel metrics? This exists for testing purposes, so locally someone can say, "yup, its outputting, now I can figure out why my actual exporter isn't working since I know thats not the issue."

yeah. Or someone building 3rd party tooling around that stuff. If it's intended for debug purposes it's not as bad.

PR #56, opened 2020-03-15
Instrumentation info attributes from named tracers

tsloughter:
Based on OTEP-83 https://github.com/open-telemetry/oteps/blob/master/text/0083-component.md

### DIFF
--- a/include/ot_span.hrl
+++ b/include/ot_span.hrl
@@ -18,8 +18,13 @@
 
 %% Holds information about the instrumentation library specified when
 %% getting a Tracer from the TracerProvider.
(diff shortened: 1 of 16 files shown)


tsloughter:
Just realized this needs changes after recent updates to the spec. So this isn't completely ready, I'll be updating it to not keep instrumentation library name/version as attributes but as their own term that will be used directly in the exporter.

tsloughter:
Now it is ready.

tsloughter:
"Ready for review", still requires PR 40 in API to be merged before this can be updated and pass tests.

PR #61, opened 2020-03-26
updates use of api to version removing wts

tsloughter:


test/opentelemetry_SUITE.erl
```diff
@@ -168,7 +168,7 @@ update_span_data(Config) ->
                        span_id=SpanId} = ?start_span(<<"span-1">>, #{links => Links}),
     ?set_attribute(<<"key-1">>, <<"value-1">>),
 
-    Events = opentelemetry:events([{opentelemetry:timestamp(),
+    Events = opentelemetry:events([{erlang:system_time(nanosecond),
```
hauleth: Why `erlang:system_time/1` there instead of `opentelemetry:timestamp/0`?

PR #65, opened 2020-03-30
trying github actions

tsloughter:


tsloughter:
Ugh, I thought disabling circleci building for PRs from forked repos would mean they wouldn't run in this PR but they do, so this attempt is pointless.

PR #72, opened 2020-04-15
drop attributes that aren't a list

tsloughter:
to guard against ++ blowing up if NewAttributes or NewEvents are
not a list we simply drop anything that isn't a list in
set_attributes/2 and add_events/2


ferd:
That's not very fail-fast and it's overly defensive but I guess there's a good reason for it?

tsloughter:
Do you mean its not fail fast because it doesn't crash or because the check is not earlier?

We don't want it to fail/crash, its lossy telemetry.

But my original reasoning for doing this here instead of in the API I think now is wrong. Since the type in the API is a list we should check there and not here. Closing and reopening on the API.

PR #78, opened 2020-06-23
Set default span kind to internal

bryannaegele:
According to the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#spankind) the default span kind should be `internal`


bryannaegele:
Signed

bryannaegele:
I signed it

PR #81, opened 2020-06-29
replace parent sampler with parent_or_else

tsloughter:
0.4 Otel spec replaced the parent sampler with parent_or_else (https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#parentorelse).
parent_or_else is configured with a sampler to run if there is
no parent span (a root span). The default sampler, always_on,
is used if none is given.

Example configuring sampler to use parent_or_else:

{parent_or_else, #{delegate_sampler => {Sampler, SamplerOpts}}}

### DIFF
--- a/include/ot_sampler.hrl
+++ b/include/ot_sampler.hrl
@@ -18,4 +18,4 @@
 
 -define(NOT_RECORD, not_record).
 -define(RECORD, record).
--define(RECORD_AND_PROPAGATE, record_and_propagate).
+-define(RECORD_AND_SAMPLED, record_and_sampled).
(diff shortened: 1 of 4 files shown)


include/ot_sampler.hrl
```diff
@@ -18,4 +18,4 @@
 
 -define(NOT_RECORD, not_record).
 -define(RECORD, record).
--define(RECORD_AND_PROPAGATE, record_and_propagate).
+-define(RECORD_AND_SAMPLED, record_and_sampled).
```
hauleth, line 21: Does the mix of present and past tense is intended there?

tsloughter: Here, yes. In the spec, I don't know :).

tolbrino:
LGTM, the configuration could use some documentation.

PR #85, opened 2020-07-15
update metrics to work with 0.4.0 metrics api

tsloughter:


src/ot_meter_default.erl
```diff
@@ -187,3 +189,45 @@ handle_cast(_Msg, State) ->
...
+insert_new_instrument(Name, InstrumentKind, Opts) ->
+    ToInsert = instrument(Name, InstrumentKind, Opts),
+    insert_new(ToInsert).
+
+insert_new(ToInsert) ->
+    try
+        ets:insert_new(?TAB, ToInsert)
+    catch
+        C:T ->
+            ?LOG_INFO("Unable to create instruments because of error ~p:~p at ets:insert_new(~p, ~p)", [C, T, ?TAB, ToInsert]),
```
ferd, line 202: Would it make sense to go structured?

src/ot_meter_default.erl
```diff
@@ -187,3 +189,45 @@ handle_cast(_Msg, State) ->
...
+            false
+    end.
+
+insert_new_instruments(List) when is_list(List) ->
+    ToInsert = lists:filtermap(fun({Name, InstrumentKind, Opts}) ->
+                                       instrument(Name, InstrumentKind, Opts);
+                                  (_) ->
+                                       false
+                               end, List),
+    insert_new(ToInsert);
```
ferd, line 212: For reporting sake, would it make sense to call `insert_new` on each individual instrument? Otherwise if you submit many and you get a failure you gotta figure which one is problematic in the whole list? I assume creating them is rare enough that the performance hit of multiple calls is worth it compared to the clarity in errors

src/ot_meter_default.erl
tsloughter: Yup, I have a `TODO` marked somewhere about doing that. I guess I should just get it done if its just passing it as a map as the first argument.

tsloughter: Does the default stdout logger include the metadata? Or do I need to add to the readme/docs how to not miss important log information when setting up a release.

src/ot_meter_default.erl
tsloughter: Hm, that is a good point. They should only be created once at the beginning so you are right that it may as well be an insert for each. I'll update to that.

src/ot_meter_default.erl
tsloughter: Updated.

src/ot_meter_default.erl
tsloughter: Updated.

PR #86, opened 2020-07-23
Child span count is deprecated

bryannaegele:
The corresponding `local_child_span_count` field in the protos has been removed in open-telemetry/opentelemetry-proto/pull/107


tolbrino:
LGTM. Are these protocol changes documented somewhere which also needs to be updated?

PR #90, opened 2020-08-01
Format propagation header values to binary

bryannaegele:
The `traceparent` header value is currently being returned as a mixed list vs a single binary value. This should be returned as a single binary.

Current behavior:

```elixir
[
  {"traceparent",
   ["00", '-', 'ef7ad0605733c5b943f4602eddac6adb', '-', '81fb003c23ab543b', '-',
    "01"]}
]
```

Expected behavior:

```elixir
[
  {"traceparent", "00-ef7ad0605733c5b943f4602eddac6adb-81fb003c23ab543b-01"}
]
```

### DIFF
--- a/src/ot_propagation_http_b3.erl
+++ b/src/ot_propagation_http_b3.erl
@@ -38,11 +38,11 @@ inject(_, #tracer_ctx{active=#span_ctx{trace_id=TraceId,
 inject(_, #tracer_ctx{active=#span_ctx{trace_id=TraceId,
                                        span_id=SpanId,
                                        trace_flags=TraceOptions}}) ->
-    Options = case TraceOptions band 1 of 1 -> "1"; _ -> "0" end,
+    Options = case TraceOptions band 1 of 1 -> <<"1">>; _ -> <<"0">> end,
     EncodedTraceId = io_lib:format("~32.16.0b", [TraceId]),
     EncodedSpanId = io_lib:format("~16.16.0b", [SpanId]),
-    [{?B3_TRACE_ID, EncodedTraceId},
-     {?B3_SPAN_ID, EncodedSpanId},
+    [{?B3_TRACE_ID, iolist_to_binary(EncodedTraceId)},
+     {?B3_SPAN_ID, iolist_to_binary(EncodedSpanId)},
      {?B3_SAMPLED, Options}];
 inject(_, undefined) ->
     [].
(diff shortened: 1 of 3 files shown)


tsloughter:
The reason is to keep anything that can be an iolist as an iolist. The result of the encoding/injection should be a value for sending over the wire, so why does it need to be a single binary instead of an iolist?

bryannaegele:
Because an iolist is not what is expected by http client libraries for a header value. At least in Elixir, all the clients I know require headers to be `[{binary(), binary()}]`. So if we're calling this function `inject_headers` and accepting headers as an input, it should return values in a usable format.

tsloughter:
grrr, for a request they should accept iolists. I thought only `httpc` was like this (it actually requires lists and doesn't even except binaries for headers).

I guess we have to merge this then, once the tests work. But going to also go poke around http clients and see if they actually don't work or if they just set their typespecs to require binary.

bryannaegele:
Mint was the offender in this case (has a binary guard), which is maybe a
little surprising given the perf focus.

bryannaegele:
Tesla enforces `is_binary/1` as well, so I'm going to get this PR fixed up. I think this is OK given they'll have to be converted at some point and it's efficient up to this point.

tsloughter:
Shouldn't have to be converted. The HTTP request should be constructed as an iolist and sent as is to gen_tcp.

Only have to be converted since these clients force use of binary for some reason.

bryannaegele:
OK. I'm going to leave this as is while what needs to be done gets figured out. 

I don't know that updating all of these client libs is or telling users to convert it themselves is a great experience towards adoption. We can hide some of that in bridge libraries but it's not a great solution.

tsloughter:
Oh, no, I was just pointing out there should be no reason. But we should merge this patch after tests are fixed because currently too many clients require it.

bryannaegele:
I couldn't figure out a test strategy that could be quite the same as before since the decoded values aren't binary and shouldn't be.

ferd:
Seconding Tristan here. Converting to binary is in theory useless because we can skip the cost altogether with the VM just sending iovecs to the underlying C code, which ends up forwarding traffic . Mint and Tesla force the caller to pay an entirely optional IO cost. Agreeing as well that we can't do much but work around it though at this point.

bryannaegele:
Would it make more sense to just wrap this in the Elixir portion of the SDK? Sounds like this might only be an issue in that language.

ferd:
Yeah wrapping in the elixir client might actually be an interesting compromise. Doesn't fix the `httpc` issue side of it (which is in both langs and lists-only) but works around problems with most other Erlang clients and should reduce memory pressure for them.

tsloughter (approved):

PR #98, opened 2020-09-19
Single repo

tsloughter:
This contains other commits, including support for https://github.com/open-telemetry/opentelemetry-specification/pull/875

Such changes really show how badly we need to be in one repo so I'm opening this now.

The tests currently fail because of a bug in rebar3 where tests can't find the headers in the apps.


ferd:
I'm checking out the path stuff.

ferd:
I'm checking and `?current_meter` is just not a macro that appears to be defined anywhere. It's not that the include files aren't included, it's that it's not defined in any of them. This and the `?new_instruments/1` and `?counter_add` macros need to be prefixed with `?ot_` to work.

tsloughter:
Hah, yup, thanks. Forgot I renamed them so thought existing tests that worked before the restructure were just suddenly failing.

hauleth:
Why not use `git subtree` to maintain the edit history?

tsloughter:
I just tried it to see what the history would look like and it doesn't appear to keep any history. It creates a single commit (even when not using `--squash`).

tsloughter:
Or maybe you have to go through the history a different way? Some special `log` commands for viewing it?

hauleth:
I will test it in the evening here, and create additional PR if I succeed.

tsloughter:
Using #99 instead.

PR #104, opened 2020-09-23
Run Elixir tests in GitHub Actions

hauleth:


hauleth:
For some reason `rebar3 dialyzer` didn't catch all the issues in the project (some types weren't exported and were reported ok anyway). I did't dig much more, but someone experienced with Dialyzer should take closer look on it.

tsloughter (approved):

PR #106, opened 2020-09-25
Rename application folders

hauleth:
In my opinion it makes it clearer what is what. On the other hand older
naming system was clearer about package names. So it is up to the
discussion whether we want this change.


tsloughter:
Not sure why this was closed but to be clear we can't have a different name for the apps because of how rebar3 `git_subdir` works. They must be the name of the app or they won't work as git dependencies in rebar3.

tsloughter:
Well, actually because of how Erlang's code server works, but rebar3 enforces it at the dep level.

hauleth:
I closed it because I noticed it will not work.

PR #109, opened 2020-09-25
replace ot_ prefix with otel_ as required by the spec

tsloughter:


bryannaegele (approved):
Seeing as this is the most breaking change we've undertaken, how do you want to coordinate the update/release of the downstream packages? I'd rather do this right away with another release and get the two big changes (this and ctx) knocked out before updating the protos in the exporter in a subsequent release.

tsloughter:
Agreed.

I also wonder if we shouldn't consider moving shit into this repo. The libraries are small and could be added to this repo either under `apps/` or in a separate directory.

tsloughter:
Think I'll merge this and then get a PR open to finalize support for https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/context/api-propagators.md#propagators-api

bryannaegele:
@tsloughter I'd be happy to at least get the exporter in here, if not the others. I think it would make all of our lives easier.

tsloughter:
Ah yea, the OTLP exporter would be good to merge in here.

PR #110, opened 2020-09-25
support user defined text_map propagators

tsloughter:
As Bogdan pointed out in this issue, https://github.com/open-telemetry/opentelemetry-specification/issues/1019#issuecomment-699209055 we don't have a way for a user to define a propagator.

This is just HTTP (Text) propagators for the moment and pretty sure I need to rename it to `text`, it is based off PR #109 so waiting on that to merge before it can be merged anyway, giving me time to bring it up to date with the latest spec and add the binary, or a more generic, propagator support.


bogdandrutu:
@tsloughter two suggestions:
1. Can you write maybe in the examples directory, one "dummy" propagator to prove this API? 
2. Why can I not write a propagator with only depending on the `API` repo? Or maybe I am completely lost in the erlang world.

tsloughter:
1. Yup, will add that to this PR.

2. I don't remember how they ended up in the SDK. Looks like the spec now has them under Context and all in the API. I'll move at least what is needed for defining them over the API.

Soooo glad we are in a single repo for the SDK and API now, much easier to do this.

bogdandrutu:
>Soooo glad we are in a single repo for the SDK and API now, much easier to do this.

I still see two repos, not sure if you are joking here :)

tsloughter:
@bogdandrutu hah, no, not a joke. The https://github.com/open-telemetry/opentelemetry-erlang-api has been merged into this repo, SDK and API being under the `apps/` directory. We need to have the old API repo archived.

tsloughter:
Oi, there were a lot of additions to the propagators specs:

* https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/context/api-propagators.md
* https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/baggage/api.md#baggage-propagation

The [Composite Propagator](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/context/api-propagators.md#composite-propagator) we already support through the propagators being a list, I don't know if there is good reason to change this to being a propagator itself.

Another difference is the lack of getter and setter arguments. These might be needed because of what I mention next (keys like `uberctx-{user-defined-key}` but right now there is only 1 key per propagator and getting/setting in the Carrier and Context are handled by their respective modules and not by the propagators themselves, making getter/setter currently unnecessary as far as I can tell.

To still be done: the `keys` function https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/context/api-propagators.md#keys to support stuff like `uberctx-{user-defined-key}`.

tsloughter:
Right now I'm going to add a "custom" propagator and a test for it before adding docs for how to implement and use the builtin and custom propagators.

tsloughter:
In commit https://github.com/open-telemetry/opentelemetry-erlang/pull/110/commits/0188ee39790f91df8a3a106beea9cf2fddd5af5e I add a `custom_propagator` module that implements the propagator behaviour.

apps/opentelemetry_api/src/otel_propagator.erl
```diff
@@ -0,0 +1,66 @@
...
+-callback inject(term()) -> carrier().
+-callback extract(carrier(), term()) -> term().
+
+-type text_map() :: [{binary(), binary()}].
+
+%% TODO add binary carrier when it is included in the otel spec
+-type carrier() :: text_map().
+
+%% T is a carrier()
+-type extractor(T) :: {fun((T, term(), fun()) -> ok), term()}.
```
ferd, line 32: just so I'm clear, since the return value is `ok`, the expectation is that the extractor gets the values from `T` and then does something with it in the environment or some stateful component somewhere?

apps/opentelemetry_api/test/custom_propagator.erl
```diff
@@ -0,0 +1,65 @@
...
+    [];
+inject(Something) ->
+    [{?SOMETHING_TEXT_ID, Something}].
+
+extract(TextMap, _) when is_list(TextMap) ->
+    case lists:search(fun({Key, _Value}) ->
+                              string:equal(Key, ?SOMETHING_TEXT_ID, true, none)
+                      end, TextMap) of
+        {value, {_, Value}} ->
+            Value;
```
ferd, line 59: if this is an extractor, the value is just discarded in the fold that handles its output, doesn'T it?

tsloughter: It isn't, it is an `extract` functions. Yes, very confusing and I need to clean that up but not sure of a better name.

The "extractor" is what is stored in the list of propagators. It will populate the Context, the 3rd argument `fun()` is what the `extract` function is. I'm updating that to at least be a real type and not just `fun()` now.

ferd (approved):

PR #114, opened 2020-09-30
Import OTLP exporter

### LINKED ISSUE #113: Move exporter app to this repo
At least the OTLP exporter, https://github.com/opentelemetry-beam/opentelemetry_exporter, should be moved to this repo.

hauleth:
Close #113

@tsloughter - remember to merge it "clasically" not with squash to preserve history.


hauleth:
@gugahoa we will need you to sign CLA

tsloughter:
Hm, compilation fails but it shouldn't. Maybe the rebar3 version being used is older than 3.14.

hauleth:
@tsloughter it failed due to usage of old `ot_*` prefix instead of new one `otel_*`, it also required some movement in the `rebar.config` and submodules.

tsloughter:
Aah, ok, thought it couldn't find the header file.

hauleth:
Yes, it couldn't find header file, as header file also uses `ot_*` prefix (IMHO needlessly as header files are already namespaced).

tsloughter:
> IMHO needlessly as header files are already namespaced

Yea, I can never decide how best to name headers, esp in public projects, and just went with being consistent across modules and headers.

And you won't believe what people will do with adding include file paths to their builds and not using the application path in `include_lib`, hehe, so the "namespacing" is fragile.

hauleth:
Ok, I have no idea how to fix the remaining issue. It seems like something in `grpcbox` is failing and I cannot find why.

tsloughter:
Ok, I can give it a look in a bit.

tsloughter:
@hauleth looks like an issue with connecting to the otel collector, I can't reproduce locally. But can you first update the subtree, I just merged fixes for the protocol talking to the collector I forgot to merge before https://github.com/opentelemetry-beam/opentelemetry_exporter/pull/14

hauleth:
@tsloughter managed, the problem was that `network: host` will not work on macOS, so you need to use "regular", and in such case you need to listen on all interfaces (`0.0.0.0`), not on `localhost`.

hauleth:
@gugahoa ping

tsloughter (approved):

PR #115, opened 2020-10-01
remove context modification in start and end_span

tsloughter:
This removes the need for #tracer_ctx and instead only stores the current span ctx.


tsloughter:
It is a draft because things to still do:

* Finalize if this is the direction we want to go. It does match the official spec
* I think update Elixir macros and tests
* Docs
* Pretty sure I'm missing at least one that I already had in mind but isn't coming to me at the moment...

GA is soon and I think we should lock in with either the existing API or this new one. A benefit of the new one is it is closer to the spec.

apps/opentelemetry_api/test/opentelemetry_api_SUITE.erl
```diff
@@ -70,30 +72,37 @@ noop_tracer(_Config) ->
...
+    ?assertMatch(SpanCtx2, ?current_span_ctx),
 
     %% start another child of the 2nd span
     SpanCtx4 = ?start_span(<<"span-4">>),
-    ?assertMatch(SpanCtx4, ?current_span_ctx()),
 
-    %% end 4th span and 2nd should be current
-    ?end_span(),
-    ?assertMatch(SpanCtx2, ?current_span_ctx()),
+    ?set_current_span(SpanCtx4),
```
ferd, line 86: so something switched and it's all explicit now?

tsloughter: Yes. It is possible to do it without explicit setting by using `with_span`. Doing it this way means we can get rid of our odd tracking of `previous` and `active` in `#tracer_ctx{}` that makes it easy for a user to accidentally act on a child span they weren't meaning to.

tsloughter: If people really want we can add a `start_as_current_span` that sets it to the current span, overwriting the previous Context and not tracking it, but I suspect it won't be needed.

ferd (approved):

PR #117, opened 2020-10-03
move all functions that take a Tracer to the tracer module

tsloughter:
The first commit here is actually PR #115 so can be ignored. The second commit is where the changes for this PR are found.

Thsi change is the recognition that the functions in `otel_span` that use a `Tracer` should really be in `otel_tracer`. It might make sense to also remove passing `SpanCtx` from these functions and have the functions (instead of the macros that call them) handle looking up the current span. 

I think removing the `SpanCtx` argument to the `otel_tracer` functions makes the most sense *if* we could offer `otel_span` functions that accepted the `SpanCtx` and updated it. This is because looking up the `SpanCtx` on each operation makes less sense than if you can just pass it in and not do a context lookup every single time. 

The problem is our Spans are stored in ETS and updated through the SDK. The API needs a way to know what module to call. So to not need the Tracer would mean adding the module to the SpanCtx.

Instead I think the best bet might be this change, plus adding functions in the API to `otel_span` that accept a full Span and return a Span:

```erlang
set_attribute(Span=#span{attributes=Attributes}, K, V) ->
  Span#span{attributes=[{K,V} | Attributes]}.
```

Thoughts?


tsloughter:
After discussing in the meeting today I'm going to close this and open a new PR that does parts of what this does but not the major change of moving functions like `set_attribute` to Tracer.

tsloughter:
I think superseded by #126 but leaving open for right now.

PR #121, opened 2020-10-10
add Elixir modules for baggage and context

### LINKED ISSUE #120: Add Elixir Context module
Expose `otel_ctx` functions in an Elixir module `Ctx` or `Context`.

tsloughter:
Resolves #120

### DIFF
--- a/apps/opentelemetry/test/opentelemetry_SUITE.erl
+++ b/apps/opentelemetry/test/opentelemetry_SUITE.erl
@@ -208,7 +208,7 @@ propagation(Config) ->
     ?assertEqual(undefined, ?current_span_ctx()),
 
     %% clear our baggage from the context to test extraction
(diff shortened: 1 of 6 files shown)


hauleth:
Out of interest. Why do we implement such modules that are fully `defdelegate`. I would understand situation when we need to have some changes to be made to make it usable with Elixir, like defining Elixir macros or changing order of arguments. `in this case I feel that it is a little bit unneeded in most cases, as calling Erlang functions is pretty straightforward.

tsloughter:
Because modules like Tracer and the metrics modules exist that do do more than just defdelegate. It was discussed at some point that it is better than just having some modules with Elixir counterpart to have all.

The SDK has no Elixir modules because they would all just be defdelegates, but the API feels more consistent if you aren't mixing different types of module calls.

ferd (approved):

PR #134, opened 2020-10-31
update sampler to take a context and update some names to match spec

tsloughter:


apps/opentelemetry/src/otel_span_ets.erl
```diff
@@ -47,24 +47,30 @@ start_link(Opts) ->
...
+    case otel_span_utils:start_span(Ctx, Name, Opts) of
+        {SpanCtx=#span_ctx{is_recording=true}, Span=#span{}} ->
+            Span1 = Span#span{instrumentation_library=InstrumentationLibrary},
+            Span2 = Processors(Ctx, Span1),
+            _ = storage_insert(Span2),
+            SpanCtx;
+        {SpanCtx, Span=#span{}} ->
+            %% span isn't recorded so don't run processors
+            %% but we do insert to ets table?
+            _ = storage_insert(Span),
```
ferd, line 62: Is this a spec thing or just wondering?

ferd (approved):

apps/opentelemetry/src/otel_span_ets.erl
tsloughter: Yea. It is something we need to figure out next and I gave up on completely solving in this PR.

I'm unsure if we need an actual `#span{}` in the ets table for a recorded but not sampled span.

PR #138, opened 2020-11-02
Add RecordException to Elixir's Span module

gugahoa:
Reference for [`RecordException`](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#record-exception)

`RecordException` has been implemented only on Elixir side as it has a
more uniform way to deal with exceptions as a value.

The attribute `exception.escaped` is left up to the user to supply
through the `attributes` parameter in `record_exception/4`.


gugahoa:
I would like some comments on how to approach this from Erlang side

I researched about exceptions in Erlang, and they don't differ from actual values at all. It’s hard to settle on a good interface for `record_exception`

I read [here](https://erlang.org/workshop/2004/exception.pdf) about how the `try` came to be, and the only way I could think of a good interface for `record_exception` would be the triple `{class, term, trace}` mentioned in Figure 8, but I couldn't extract that triple from the `try` as a user.

gugahoa:
I also increased the minimum Elixir version to `1.11`, so that `is_exception` could be used as a guard, but if it's not desirable I can refactor to remove that

tsloughter:
Is "message" required or an optional field? The type in Erlang would be the "exception" (2nd part of `C:T:S`). But we'd want the "class" as well. So maybe it'd be include the class and exception in the same "type" field of the event? Other option would be to use type as the class and message for the exception.

I'd prefer "message" to be the result of a call to `Module:format_error(T)` passing the exception. The question there would be if we relied on the stacktrace to know the `Module` to call or require the exception to be of the form `{Module, T}`.

gugahoa:
@tsloughter at least one of message or type should be present in the event.

My whole issue with the Erlang side is about which arguments to accept.

Should the function require that the user know what to pass? What if they mistakenly pass a incorrect argument? How would we know?

tsloughter:
It is on them to pass the right arguments, that is always true. But we can also optionally handle it for them in the `with_span` function since we already have a `try`.

Hm, this might be reason to replace the Elixir call of `with_span` with its own implementation so it can use whatever exception stuff it has?

For Erlang I guess we need to offer a couple options, like `otel_span:record_exception(C, T, S)` and `otel_span:record_exception(C, T, Message, S)`.

The `record_exception` function should then always combine C and T into a string `<<"C:T">>` and `Message` is passed by the user, so if they have a `Moudle:format_error(Reason)` they can pass that themselves.

@ferd any thoughts on recording exceptions in Erlang?

tsloughter:
Based on what Python does https://github.com/open-telemetry/opentelemetry-python/blob/master/opentelemetry-sdk/src/opentelemetry/sdk/trace/__init__.py#L698-L700 maybe `type` should just be the Erlang `class` of the exception.

Need to check what Java uses for that field too.

ferd:
Feels fair to use the type as a class. The direct matched value is the reason (which is often a string in other languages and therefore it's alright to use tuples or strings there too for us), and stacktrace has direct matches as well.

An interesting question will be to plan for stuff in EEP-54: https://github.com/erlang/otp/pull/2849 -- this uses Args hidden in the stacktrace to add context to the exception's reason

gugahoa:
That PR introduces a `format_exception/3`, which we could use for `exception.stacktrace`

Until then, would it be ok to use `erlang:display/1` to populate `exception.stacktrace`?

ferd:
No, `erlang:display/1` is a debug-only BIF sending data to stdout, that can't be the thing you're looking for.

tsloughter:
It also prints to stdout, it doesn't return a string. You just want `io_lib:format("~p", [Stacktrace])` or maybe a depth limited form of that.

tsloughter:
@gugahoa are you able to update this to add the Erlang version? I'd like to get it merged before making a release soon.

apps/opentelemetry/test/opentelemetry_SUITE.erl
```diff
@@ -455,6 +455,45 @@ record_but_not_sample(Config) ->
...
+    try throw(my_error) of
+        _ ->
+        ok
+    catch
+        Class:Term:Stacktrace ->
+            otel_span:record_exception(SpanCtx, Class, Term, "My message", Stacktrace, [{"some-attribute", "value"}]),
+            ?end_span(SpanCtx),
+            [Span] = assert_exported(Tid, SpanCtx),
+            [Event] = Span#span.events,
+            ?assertEqual(Event#event.name, <<"exception">>),
```
tsloughter, line 489: Should swap the order of all the `?assertEqual` calls, first argument should be the expected value.

It just makes the error message make sense when they do fail since the output will say "expected: X" and use the first argument.

apps/opentelemetry/test/opentelemetry_SUITE.erl
```diff
@@ -455,6 +455,45 @@ record_but_not_sample(Config) ->
...
+        _ ->
+        ok
+    catch
+        Class:Term:Stacktrace ->
+            otel_span:record_exception(SpanCtx, Class, Term, Stacktrace, [{"some-attribute", "value"}]),
+            ?end_span(SpanCtx),
+            [Span] = assert_exported(Tid, SpanCtx),
+            [Event] = Span#span.events,
+            ?assertEqual(Event#event.name, <<"exception">>),
+            ?assertEqual(Event#event.attributes, [{<<"exception.type">>, ["throw",58,"my_error"]},
```
tsloughter, line 471: The reason you have `58` is because a binary string isn't used when it probably should be. I believe an attribute value can technically be a list/array of integers means we need to only use binaries for strings in attributes.

gugahoa: I'm not sure I understood this.
I understood that attribute value can be a list/array of integer, but we need to use binaries for strings in attributes. So I have to change something somewhere to get a binary from that list, is that correct?

I made a fix with what I understood (https://github.com/open-telemetry/opentelemetry-erlang/pull/138/commits/3b03145c51123eca241e0667a48464cca1e0f75d) but if I misunderstood I can revert and properly fix it

apps/opentelemetry_api/src/otel_span.erl
```diff
@@ -105,6 +107,30 @@ add_events(SpanCtx=#span_ctx{span_sdk={Module, _}}, Events) when ?is_recording(S
...
     false.
 
+-spec record_exception(SpanCtx, Class, Term, Stacktrace, Attributes) -> boolean() when
+      SpanCtx :: opentelemetry:span_ctx(),
+      Class :: atom(),
+      Term :: term(),
+      Stacktrace :: list(any()),
+      Attributes :: opentelemetry:attributes().
+record_exception(SpanCtx, Class, Term, Stacktrace, Attributes) ->
+    ExceptionAttributes = [{<<"exception.type">>, iolist_to_binary(io_lib:format("~p:~p", [Class, Term]))},
```
tsloughter, line 117: I'm not sure that `Term` should be included in the type? @ferd thoughts?

ferd: That could indeed be very large.

```
1> maps:get(key, <<0:1000>>).
** exception error: {badmap,<<0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,
                              0,0,0,0,...>>}
     in function  maps:get/2
        called as maps:get(key,
                           <<0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,
                             ...>>)
2> catch maps:get(key, <<0:1000>>).
{'EXIT',{{badmap,<<0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,
                   0,0,0,0,0,...>>},
         [{maps,get,
                [key,<<0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,...>>],
                []},
          {erl_eval,do_apply,6,[{file,"erl_eval.erl"},{line,684}]},
          {erl_eval,expr,5,[{file,"erl_eval.erl"},{line,437}]},
          {shell,exprs,7,[{file,"shell.erl"},{line,686}]},
          {shell,eval_exprs,7,[{file,"shell.erl"},{line,642}]},
          {shell,eval_loop,3,[{file,"shell.erl"},{line,627}]}]}}
```

For example, a map error like that has the whole bad term in the reason. If you were to pass a 200mb file, you'll have a 200mb type key.

tsloughter: Well I was going to suggest moving `Term` to the `message` field, so that issue would still exist, just in a different field.

ferd: That would probably be safer. If you have a limit to what can be sent, then I do suggest truncating the log output the way the erlang logger (or lager) do it.

tsloughter: Yea, adding truncation in the `io_lib:format` call makes sense.

gugahoa: What char limit would be reasonable?

tsloughter: I was trying to find in other implementations (like Java's) what they use but haven't found their defaults.. Maybe 50, I think thats what Fred's flatlog defaults to.

tsloughter: @gugahoa any update? are you going with 50?

gugahoa: @tsloughter yes, I'll get around to it this weekend

gugahoa: It's ready for review now @tsloughter

tsloughter:
Great, looking good. But now I'm not sure about requiring 1.11. At work we have systems still on 1.10 and 1.9 that aren't easy upgrades. Would at least 1.10 be not much work to support for this?

gugahoa:
The only thing that makes 1.11 required is the is_exception/1 guard, we can copy paste that into the module that uses it and remove the requirement on 1.11

Does that sound good to you? @tsloughter

tsloughter:
@gugahoa ah, yea, that sounds good.

gugahoa:
@tsloughter done :top:

apps/opentelemetry_api/src/otel_span.erl
```diff
@@ -105,6 +107,30 @@ add_events(SpanCtx=#span_ctx{span_sdk={Module, _}}, Events) when ?is_recording(S
...
     false.
 
+-spec record_exception(SpanCtx, Class, Term, Stacktrace, Attributes) -> boolean() when
+      SpanCtx :: opentelemetry:span_ctx(),
+      Class :: atom(),
+      Term :: term(),
+      Stacktrace :: list(any()),
+      Attributes :: opentelemetry:attributes().
+record_exception(SpanCtx, Class, Term, Stacktrace, Attributes) ->
+    ExceptionAttributes = [{<<"exception.type">>, iolist_to_binary(io_lib:format("~p:~p", [Class, Term], [{chars_limit, 50}]))},
```
tsloughter, line 117: Instead of `~p` I think it should be `~0tP`, in which case a second argument of the term depth is needed. So like:

`io_lib:format("~0tp:~0tP", [Class, Term, 10], [{chars_limit, 50}]))`

The `0` means to not add newlines and the `10` means the depth of the term is limited to 10. I'm not sure if 10 is the best depth but seems reasonable.

What do you think @ferd ?

ferd: Yeah a depth of 10 should be acceptable and it's easier to expand than it is to clamp down usually

apps/opentelemetry_api/src/otel_span.erl
```diff
@@ -105,6 +107,30 @@ add_events(SpanCtx=#span_ctx{span_sdk={Module, _}}, Events) when ?is_recording(S
...
+
+-spec record_exception(SpanCtx, Class, Term,  Message, Stacktrace, Attributes) -> boolean() when
+      SpanCtx :: opentelemetry:span_ctx(),
+      Class :: atom(),
+      Term :: term(),
+      Message :: unicode:unicode_binary(),
+      Stacktrace :: list(any()),
+      Attributes :: opentelemetry:attributes().
+record_exception(SpanCtx, Class, Term, Message, Stacktrace, Attributes) ->
+    ExceptionAttributes = [{<<"exception.type">>, iolist_to_binary(io_lib:format("~0tP:~0tP", [Class, Term, 10], [{chars_limit, 50}]))},
```
tsloughter, line 129:
```suggestion
    ExceptionAttributes = [{<<"exception.type">>, iolist_to_binary(io_lib:format("~0tP:~0tP", [Class, 10, Term, 10], [{chars_limit, 50}]))},
```

apps/opentelemetry_api/src/otel_span.erl
```diff
@@ -105,6 +107,30 @@ add_events(SpanCtx=#span_ctx{span_sdk={Module, _}}, Events) when ?is_recording(S
...
     false.
 
+-spec record_exception(SpanCtx, Class, Term, Stacktrace, Attributes) -> boolean() when
+      SpanCtx :: opentelemetry:span_ctx(),
+      Class :: atom(),
+      Term :: term(),
+      Stacktrace :: list(any()),
+      Attributes :: opentelemetry:attributes().
+record_exception(SpanCtx, Class, Term, Stacktrace, Attributes) ->
+    ExceptionAttributes = [{<<"exception.type">>, iolist_to_binary(io_lib:format("~0tP:~0tP", [Class, Term, 10], [{chars_limit, 50}]))},
```
tsloughter, line 117:
```suggestion
    ExceptionAttributes = [{<<"exception.type">>, iolist_to_binary(io_lib:format("~0tP:~0tP", [Class, 10, Term, 10], [{chars_limit, 50}]))},
```

tsloughter (approved):

PR #139, opened 2020-11-06
update protos

tsloughter:


ferd (approved):
Shouldn't this come with a git submodule update though? I guess that stuff isn't locked anyway.

tsloughter:
Huh, yea, it also doesn't seem to have updated it to the commit I wanted. I'm going to figure this out and include fix to status code that comes with the latest protos.

ferd (approved):

PR #152, opened 2020-11-20
Final fixups for 0.5.0

tsloughter:
The packages have been published but these are a couple patches I had to make to fixup links and make the changelog.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ _build
 *.iml
 rebar3.crashdump
 *~
+doc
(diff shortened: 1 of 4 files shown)


ferd (approved):

ferd:
Small gotcha, the pr2relnote script only works on merge commits so we have to be careful (or maybe disable the other merging methods).

tsloughter:
Yup, true true, I will restrict those.

Would love a script that could do it without merge commits, but I guess the only option for that would be maybe squash commits that modified the commit message to be like, "PR #: ..." so it had something to match on, and github doesn't let you automate something like that.

PR #157, opened 2020-11-26
bind needs to return {{key, labels}, instrument}

robashton:


tsloughter:
Does this actually work? I'm trying to follow the code to record on a bound instrument and it looks to me that it won't work even with this change.

I'll add a test for `bind` to sort this out.

tsloughter:
Ah, I see it :). Once I added a test it was obvious.

tsloughter (approved):

PR #172, opened 2020-11-28
remove reference to opentelemetry-erlang-api in the README

tsloughter:
Fixes #171

Issue #171, opened 2020-11-28
Link to archived project in the README


CrowdHailer:
https://github.com/open-telemetry/opentelemetry-erlang/blob/master/README.md#design

The section linked above says `If you are instrumenting a project your application should only depend on the OpenTelemetry API application.` However following that link takes you to an archived project that links back around to the original project.

Unclear what the correct course of action is

tsloughter:
Damn, thanks, I'll make a PR.

PR #176, opened 2020-11-29
fix docs and links

### LINKED ISSUE #174: Broken link to named Tracer
On this section a link to `Named Tracer` is broken https://hexdocs.pm/opentelemetry_api/OpenTelemetry.html#register_application_tracer/1

I think the correct target is https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md

chulkilee:
Markdown cleanup

- Consistent heading (`#` over `=`)
- Remove whitespaces between code fence and info string as that's more commonly used
- Remove trailing spaces
- Add blank lines around lists

Updated contents

- Add hex badge for two hex packages

Fix links

- Fix links to the old API repo (#171 / #172)
- Use links with tag, instead of using `main` branch (#174)


tsloughter:
This looks good, why is it just a draft? Even if there is more you want to do can you mark this as ready and I'll merge it?

tsloughter (approved):

apps/opentelemetry_exporter/README.md
```diff
@@ -1,62 +1,86 @@
...
-  [{processors, 
+  [{processors,
     [{otel_batch_processor,
-        #{exporter => {opentelemetry_exporter, #{protocol => grpc,
-                                                 endpoints => [{http, "localhost", 9090, []}]}}}}]}]}
+        #{exporter => {opentelemetry_exporter, #{endpoints => [{http, "localhost", 9090, []}]}}}}]}]}
 ```
 
-An Elixir release uses `releases.exs`:
+Elixir with `releases.exs`:
```
tsloughter, line 65:
```suggestion
Elixir with `runtime.exs`:
```

README.md
```diff
@@ -1,132 +1,56 @@
...
+To instrument your project:
 
-> w3c_trace_context_interop:run().
-```
+- Add and use [OpenTelemetry API](apps/opentelemetry_api/) to add instrumentation manually
+- Add and configure instrumentation packages that use [OpenTelemetry API](apps/opentelemetry_api/)
+- Add and configure [OpenTelemetry SDK](apps/opentelemetry/) and exporters
 
-Then, clone the [W3C Trace Context repo](https://github.com/w3c/trace-context) and run the tests:
+To instrument your package or create instrumentation package for existing package:
```
tsloughter, line 39:
```suggestion
To instrument your package or create an instrumentation package for an existing package:
```

apps/opentelemetry/README.md
```diff
@@ -0,0 +1,98 @@
...
+[![License - Apache-2.0](https://img.shields.io/hexpm/l/opentelemetry)](LICENSE)
+
+![Build Status - Erlang](https://github.com/open-telemetry/opentelemetry-erlang/workflows/Erlang/badge.svg)
+![Build Status - Elixir](https://github.com/open-telemetry/opentelemetry-erlang/workflows/Elixir/badge.svg)
+
+---
+
+## Features
+
+Provides the reference implementation for OpenTelemetry API.
```
tsloughter, line 18:
```suggestion
Provides the default implementation for OpenTelemetry API.
```

I think "default" is better since it conveys that you can and should use it. To me "reference implementation" means something that you shouldn't use for real world work (like Weston is a reference implementation of Wayland).

apps/opentelemetry_api/README.md
```diff
@@ -1,140 +1,179 @@
...
   ]
 end
 ```
 
-### Registering and Using Tracers Directly
+## Configuring OpenTelemetry SDK
+
+Since OpenTelemetry API provides only no-op implementation, your project has to include the implementation of it to do the actual jobs.
+
+OpenTelemetry SDK (`opentelemetry` package) provides the reference implementation in Erlang. Check out the [documentation](../opentelemetry/) on how to add and configure the package.
```
tsloughter, line 46:
```suggestion
OpenTelemetry SDK (`opentelemetry` package) provides the default implementation in Erlang. Check out the [documentation](../opentelemetry/) on how to add and configure the package.
```

tsloughter (approved):

apps/opentelemetry/README.md
chulkilee: I disagree, because "reference implementation" does not mean it shouldn't be used for real world work literally. It is just "full implementation of the spec". I'm not familiar with Wayland, but that's "a case".

"default" sounds strange, since it says "this is a good first-option" - which is actually not. I don't know what OpenTelemetry refers those those implementations - I saw "reference implementation" in several places.

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk_exporters/zipkin.md

> Consider using V1SpanConverter.java as a reference implementation for converting v1 model to OpenTelemetry.

https://github.com/open-telemetry/opentelemetry-swift

> The OpenTelemetrySdk library is the reference implementation of the API.

https://blog.newrelic.com/product-news/what-is-opentelemetry/

> SDKs (reference implementations of the APIs) created specifically for languages like Java, Python, Go, Erlang, and more

apps/opentelemetry/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+TODO: what's implemented and what's not
```
chulkilee, line 20: @tsloughter could you give me a list of this? Or we may just use this line :)

We may add a link to [Compliance of Implementations with Specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/spec-compliance-matrix.md) but the doc would show the compliance of the latest version, not the specific Erlang package version..

apps/opentelemetry_api/lib/open_telemetry/tracer.ex
```diff
@@ -57,7 +57,7 @@ defmodule OpenTelemetry.Tracer do
   end
 
   @doc """
-  Takes a `t:OpenTelemetry.ctx/0` and the `t:OpenTelemetry.span_ctx/0` and the Tracer sets
+  Takes a `t:OpenTelemetry.ctx/0 - TODO` and the `t:OpenTelemetry.span_ctx/0` and the Tracer sets
```
chulkilee, line 60: This type does not exist in Elixir `OpenTelemetry` module. From the code, it seems like erlang `otel_ctx:t()` - but not sure Elixir `OpenTelemetry` has it. We may refer erlang type directly.

apps/opentelemetry_api/mix.exs
```diff
@@ -53,7 +53,7 @@ defmodule OpenTelemetry.MixProject do
     end)
     |> Enum.concat([
       {:cmark, "~> 0.7", only: :dev, runtime: false},
-      {:ex_doc, "0.21.0", only: :dev, runtime: false},
+      {:ex_doc, "~> 0.21.0", only: :dev, runtime: false},
```
chulkilee, line 56: I found that syntax highlight works are broken even with this change

- ex_doc with earmark: elixir working / erlang not working
- ex_doc with cmark: elixir not working / erlang working

I'll leave it to another PR for now. As ex_doc dropped cmark supports in ex_doc 0.22.0, we may just use earmark instead of cmark...

apps/opentelemetry/README.md
tsloughter: Thanks for finding the other uses of this for me to correct :). I brought this up with the Java SDK docs because they used "reference implementation" and they agreed and changed it. 

Not sure about the zipkin one but Swift should definitely change and I'll bring that up. The New Relic blog probably won't be able to get changed.

I'm open to something besides "default", but "reference implementation" has way too much baggage in my mind and will for many others.

apps/opentelemetry/README.md
tsloughter: It should have a link to the matrix, yes, and the list of what isn't implemented can be found in the matrix as well. I'd say mainly it should focus on the high level, like simply saying 'metrics" and "logging", and saying to "see the spec matrix" for detailed breakdown of specifics like what environment variables are supported already.

apps/opentelemetry_api/lib/open_telemetry/tracer.ex
tsloughter: Maybe this was before the `Ctx` module was added to the Elixir lib. I'm guessing it should reference it.

apps/opentelemetry_api/mix.exs
tsloughter: Yea, I ran into this as well. It was a major pain and found we are just stuck on an old version of `ex_doc` for now.

apps/opentelemetry/README.md
tsloughter: Could just get rid of "default" and "reference" and just use "implementation".

apps/opentelemetry_api/mix.exs
chulkilee: Created #204 204

apps/opentelemetry_api/lib/open_telemetry/tracer.ex
chulkilee: Added `OpenTelemetry.Ctx.t` as `:otel_ctx.t()`, and changed this line accordingly. Let me know if it's incorrect.

apps/opentelemetry/README.md
chulkilee: Having a link to the latest matrix is useful only when using the latest version of this package - which may not be useful when a dev is checking this README on old version.

For now, removing this line. Let's readd it later.

I agree that it should be high-level items, such as `Metrics`, `Logging`, etc. **per OpenTelemetry guideline**. It would be nice if OpenTelemetry has a list of stable components at specific version (e.g. 1.0.1). Right now it seems like I need to browse all components :(

For example: https://github.com/open-telemetry/opentelemetry-java

* OpenTelemetry API:
  * stable apis including Tracer, Span, SpanContext, and Baggage
  * semantic conventions Generated code for the OpenTelemetry semantic conventions.
  * context api The OpenTelmetry Context implementation.
  * metrics api alpha code for the metrics API.
* sdk defines the implementation of the OpenTelemetry API.

yatender-oktalk:
@chulkilee Hey if possible can you resolve the conflicting files, so that @tsloughter can merge it, as you have improved documentation a lot and if we don't merge all your hard work will go in vein.

hauleth (approved):

tsloughter:
@chulkilee ping

tsloughter:
Thanks again @chulkilee , I think most of these changes were incorporated at this point (at least the type reference stuff and links to specs for sure), and the repo has changed so much since this was opened I'm going to close it at this point.

PR #177, opened 2020-11-29
use elixir formatter

chulkilee:
- Add GitHub Actions job to check the format
- Run `mix format` to existing elixir code


chulkilee:
I tested the github actions check with unformatted file and confirmed it correctly mark the check failed.

I don't know why the test coverage for untouched file is decreased 🤷

tsloughter (approved):

Issue #182, opened 2020-12-03
Passing a parent context to a new span


ggpasqualino:
Hi there, I have noticed that in 0.5.0 neither `start_span` nor `with_span` functions in the module `otel_tracer`accept the `parent` option anymore.
The `start_span` function is accepting now a context parameter that seems to be used for that now, but the `with_span` doesn't have a corresponding one. And in the Elixir API, none of them accept this now.

I can take a look at adding the new functions. What do you think?

tsloughter:
@ggpasqualino if none of the Elixir API functions accept a context then this is a bug. 

Argh, and you are right about `with_span`.

Sorry about that, these are quick fixes I can get done today and release 0.5.1

ggpasqualino:
@tsloughter awesome, thank you!

tsloughter:
Oh.. I now see the bigger issue. The Elixir Tracer module had been only including macros that act on the current context with the assumption that anything that the user required passing in either a context variable or a tracer they would use `otel_tracer`.

I think that idea doesn't hold now that parent can only be passed via a Context. Still should be a simple enough change, I think.

pallix:
I want to reuse the `trace_id` of another service, to correlate traces from different services. I manage to write this proof of concept:

```

require Record
@fields Record.extract(:span_ctx, from_lib: "opentelemetry_api/include/opentelemetry.hrl")
Record.defrecordp(:span_ctx, @fields)

def hello do
    {trace_id, ""} = Integer.parse("0000000000123457", 16)
    span_id = :opentelemetry.generate_span_id()
    parent_span_ctx = span_ctx(trace_id: trace_id, span_id: span_id, tracestate: [])
    ctx = %{{:otel_tracer, :span_ctx} => parent_span_ctx}

    Tracer.with_span ctx, "hello4.1" do
      Process.sleep(200)
    end
end
```

It was hard to figure it out. Is there a simple way to achieve that?

tsloughter:
@pallix yes, you should reuse the context or span context, not simply the trace id. Is there a reason you are only reusing the trace id?

pallix:
> yes, you should reuse the context or span context, not simply the trace id. Is there a reason you are only reusing the trace id?

Not really, I'm unfamiliar with the API and trying to understand it. If I want to correlate two spans from two different services, do I need to specify both `trace_id` and `span_id` or is specifying `span_id` enough?

tsloughter:
[...]

PR #183, opened 2020-12-03
expand Elixir trace API

tsloughter:
start_span, no matter the arity, always returns only a span_ctx now

the Tracer module includes macros that accept Context variables
instead of reading the context from the pdict, allowing explicit
parent spans to be used again.

### DIFF
--- a/.github/workflows/erlang.yml
+++ b/.github/workflows/erlang.yml
@@ -30,9 +30,9 @@ jobs:
       with:
         path: |
           _build
-        key: ${{ runner.os }}-build-${{ matrix.otp_version }}-${{ hashFiles(format('rebar.lock')) }}
+        key: ${{ runner.os }}-build-${{ matrix.otp_version }}-${{ hashFiles(format('rebar.lock')) }}-1
         restore-keys: |
-          ${{ runner.os }}-build-${{ matrix.otp_version }}-
+          ${{ runner.os }}-build-${{ matrix.otp_version }}-1-
     - name: Compile
       run: rebar3 compile
     - name: EUnit tests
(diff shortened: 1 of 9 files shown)


tsloughter:
Resolves https://github.com/open-telemetry/opentelemetry-erlang/issues/182

apps/opentelemetry/test/opentelemetry_SUITE.erl
```diff
@@ -106,7 +106,7 @@ with_span(Config) ->
     otel_tracer:set_current_span(SpanCtx1),
 
     Result = some_result,
-    ?assertMatch(Result, otel_tracer:with_span(Tracer, <<"with-span-2">>,
+    ?assertMatch(Result, otel_tracer:with_span(Tracer, <<"with-span-2">>, #{},
```
ferd, line 109: This doesn't match the type specs where the Ctx is expected first under the new API and this should not pass in Dialyzer if used outside of a test

ferd (changes requested):

.github/workflows/erlang.yml
```diff
@@ -79,9 +79,9 @@ jobs:
       with:
         path: |
           _build
-        key: ${{ runner.os }}-build-${{ matrix.otp_version }}-${{ hashFiles('rebar.lock') }}
+        key: ${{ runner.os }}-build-${{ matrix.otp_version }}-${{ hashFiles('rebar.lock') }}-1
         restore-keys: |
-          ${{ runner.os }}-dialyzer-${{ matrix.otp_version }}-
+          ${{ runner.os }}-dialyzer-${{ matrix.otp_version }}-1-
```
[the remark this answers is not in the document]
tsloughter: It clears the cache by making a new key/prefix.

apps/opentelemetry/test/opentelemetry_SUITE.erl
tsloughter: I think you are mistaking the API which has `with_span/4` and the SDK which has only `with_span/5`.

In the API the context is read in and passed to the SDK, so the SDK only has to have a call that accepts the context as the first argument.

ferd: I'm looking at the direct call in https://github.com/open-telemetry/opentelemetry-erlang/blob/1372b6569d3a27199a05718fcf57b33177282c7d/apps/opentelemetry_api/src/otel_tracer.erl#L67-L74 and I guess that's the start_opts/4 rather than the context being passed as an empty map yeah. My bad.

ferd (approved):

PR #184, opened 2020-12-05
Make span context (re)construction more ergonomic

sadesyllas:
Hello and congratulations for your approach in this project.

#### TL;DR

This commit enables one to choose how to construct a new span, by defining both the context and the active span, using a trace id and a span id, in the same call (ie, in the `start_opts` of `start_span` and `with_span`).

#### Description

While working with this library's API, I found it hard to reconstruct the span context between processes while working in a fully asynchronous manner (eg, RabbitMQ and Broadway with lots of processes here and there and wishing to only share a trace id and a span id between them).

As such, this commit implements the ability to set the current span using a trace id and a span id, through a new implementation of `otel_tracer:set_current_span/2`.

### DIFF
--- a/apps/opentelemetry_api/lib/open_telemetry/tracer.ex
+++ b/apps/opentelemetry_api/lib/open_telemetry/tracer.ex
@@ -17,16 +17,7 @@ defmodule OpenTelemetry.Tracer do
       end
   """
 
-  @type start_opts() :: %{
-          optional(:parent) => OpenTelemetry.span() | OpenTelemetry.span_ctx(),
-          optional(:attributes) => OpenTelemetry.attributes(),
-          # TODO sampler should is an opaque type defined in the implementation
-          optional(:sampler) => term(),
-          optional(:links) => OpenTelemetry.links(),
-          optional(:is_recording) => boolean(),
-          optional(:start_time) => :opentelemetry.timestamp(),
-          optional(:kind) => OpenTelemetry.span_kind()
-        }
+  @type start_opts() :: :otel_span.start_opts()
 
   @doc """
   Starts a new span and does not make it the current active span of the current process.
@@ -36,8 +27,8 @@ defmodule OpenTelemetry.Tracer do
   `parent` option given then the Tracer checks for an extracted SpanContext to use as the parent. If
   there is also no extracted context then the created Span is a root Span.
   """
-  defmacro start_span(name, opts \\ quote(do: %{})) do
-    quote bind_quoted: [name: name, start_opts: opts] do
+  defmacro start_span(name, start_opts \\ quote(do: %{})) do
+    quote bind_quoted: [name: name, start_opts: start_opts] do
       :otel_tracer.start_span(:opentelemetry.get_tracer(__MODULE__), name, start_opts)
     end
   end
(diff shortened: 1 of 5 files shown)


tsloughter:
> (eg, RabbitMQ and Broadway with lots of processes here and there and wishing to only share a trace id and a span id between them).

Why "only a trace id and span id"?

sadesyllas:
I turned this PR into a Draft PR since new tests need to be written, which I have not come around to doing yet.

Once the tests are complete, I will remove the Draft status from this PR.

sadesyllas:
> > (eg, RabbitMQ and Broadway with lots of processes here and there and wishing to only share a trace id and a span id between them).
> 
> Why "only a trace id and span id"?

@tsloughter, I was thinking about a (large scale) system where a request enters the system using service A, then it travels through, eg, RabbitMQ to another part of the system and then it comes back.

I suppose that we have links to cover this case, if I am not mistaken, so another trace could start somewhere else perhaps as a linked span, but I was wondering if one could simply *resume* the trace under the same trace id (eg, a common trace id logged in graylog which could span several asynchronous parts of the system).

In the same vein, if an action is then split into several parallel requests (eg, service A asks services B and C to act on a request in parallel), then span A could end up having two children, namely spans B and C, in the end. 

Apart from the `trace_id/span_id` functionality, I also added the option to use a `context` in `start_opts`, so as to make it easier to pass on `otel_ctx.t()`s while creating spans.

I felt that the combination of the two enhances the ergonomics of the API.
 
What do you think about the above use cases?

Do you believe that this commit helps in that regard or is there another way that I may have missed?

tsloughter:
Quick response, passing a context in the Elixir api is covered by https://github.com/open-telemetry/opentelemetry-erlang/pull/183 -- it is the first argument, not part of `start_opts`. Will reply to the rest in a bit.

sadesyllas:
> Quick response, passing a context in the Elixir api is covered by #183 -- it is the first argument, not part of `start_opts`. Will reply to the rest in a bit.

I removed the `context` related part of the PR (ie, setting it in `start_opts`), since it has been covered in #183.

tsloughter:
For going through rabbitmq you want to look at propagation and https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/messaging.md -- we'll want a rabbitmq instrumentation library that any user can plugin to their system using rabbit. The doc also covers when to use a link when tracing code communicating through a messaging system.

sadesyllas:
Thank your for the link, I'll check it out, although from a quick read it seems that the correct way to propagate a trace is with links.

And you are right about the need for an instrumentation library.

bryannaegele:
I believe the example given for links is a number of requests that then got rolled up into a batch, with the batch being a separate trace but containing links to each if the originating requests that make up the batch.

For the use case you originally illustrated, it's suggested you simply propagate the context to your messages and then have your consumer extract it. There's no official way of doing this right now. At work we're just serializing/deserializing the context from tuple -> list, list -> tuple with headers.

I don't think there is anything official for this stuff yet. I suppose we could adapt what we're doing with an adapter around the w3c propagation module.

sadesyllas:
Thanks for the insight @bryannaegele.

So the proposal here is that in the context of the same application, we can pass around the context as an opaque structure, if needed.

In the distributed scenario, the serialized context carries both the trace id and the span id, so it could be passed to an entirely different stack as well (ie, other clients to the same backend, etc).

And now that I come to think about it, the trace id, being a 128-bit integer, would need to be passed in a different way if the receiving system has problems dealing with such integer sizes, so some form of *context-aware* serialization must take place.

I think that after all the input I gathered in this PR, I should close it, unless you see any merit in keeping the `{trace_id, span_id}` parent functionality.

sadesyllas:
I just tried out what @bryannaegele suggested, ie, merely to (de)serialize the context myself and it worked great in a PoC with two services.

As such, I am closing this PR.

Thank you all for your input! :vulcan_salute:

PR #188, opened 2020-12-09
add rationale document with versioning/releasing details

tsloughter:
This document will contain additional answers to questions that commonly come up about the design of this project. An example of what should probably be added would be use of the process dictionary and ETS.

I wrote this up quickly this morning and haven't reviewed, will try to get to that later this morning :). 

The initial entry in this document is about versioning and naming and fulfills the task https://github.com/open-telemetry/opentelemetry-specification/issues/1267

Note that a separate experimental Application is not absolutely required, but I do think it makes the chances of users relying on unstable APIs in 1.0 much lower and that makes our and their lives easier.

I really didn't want to have to also split the SDK into 2 Applications but I couldn't figure out how to rationalize that since it still has some public APIs and we want any 1.x API to work with the latest SDK release.

I'm completely open to alternatives. Please read the design doc to understand the goals https://github.com/open-telemetry/oteps/pull/143

### DIFF
--- a/docs/rationale.md
+++ b/docs/rationale.md
@@ -28,10 +28,12 @@ the pre-1.0 world as functionality was floating from SDK to API at times.
 ### Experimental API (`opentelemetry_api_experimental`)
 
 The experimental package is where any API that is not stable when 1.0 is
-released must live. At this time (prior to 1.0) that means Metrics and Logging.
+released [MUST] live. At this time (prior to 1.0) that means Metrics and Logging.
 
 This package will always be 0.x because it is never stable and modules will be
 removed when they are moved to the stable API package.
+
+[MUST]: https://tools.ietf.org/html/rfc2119#section-1
 
 ### API (`opentelemetry_api`)
 
--- a/docs/rationale.md
+++ b/docs/rationale.md
@@ -1,0 +1,129 @@
+Rationale
+---
+
+When creating a library, often times designs and decisions are made that get lost over time. This document tries to collect information on design decisions to answer common questions that may come up when you explore.
+
+# Versioning and Releasing
+
+The OTP Applications and OpenTelemetry Spec itself use semver v2.
+
+In this document references to OTP concepts is done by capitalizing the word
+(for example Application and Release) while the generic term (like release) is
+lowercase).
+
+## Module Prefix
+
+The module prefix for all modules in any of these core Applications is
+`otel`. This means modules can move between Applications without their name
+changing and because of Erlang's flat namespace there is no code to change for a
+user when a stable API graduates from experimental to stable, if the user was
+using the latest version of the experimental API.
+
+This also allows flexibility for modules that might be in the SDK but are found
+to be better placed in the API, or vice versa. This has happened a few times in
+the pre-1.0 world as functionality was floating from SDK to API at times.
+
+## OTP Applications
+
+### Experimental API (`opentelemetry_api_experimental`)
+
+The experimental package is where any API that is not stable when 1.0 is
+released must live. At this time (prior to 1.0) that means Metrics and Logging.
+
+This package will always be 0.x because it is never stable and modules will be
+removed when they are moved to the stable API package.
+
+### API (`opentelemetry_api`)
+
+The API package is the stable API package that we must provide semver defined
+backwards compatibility once a major (1.0) release is made. When an API becomes
+stable its modules are moved from `opentelemetry_api_experimental` to
+`opentelemetry_api` and a new minor release of both is published.
+
+At the time of 1.0 the following APIs will live in the `opentelemetry_api`
+package:
+
+* Tracing
+* Baggage
+* Context
+
+### Experimental SDK (`opentelemetry_sdk_experimental`)
+
+The experimental SDK contains the implementations for the APIs in the
+experimental API of the same version.
+
+Any setup for signals contained in the experimental SDK must be done on startup
+of the experimental SDK -- for example, setting the default Meter would be done
+in `start/2` of `opentelemetry_sdk_experimental`.
+
+This Application is versioned in lockstep with the experimental API and will
+never go above 0.x.
+
+### SDK (`opentelemetry`)
+
+Functionality is implemented in this Application and the API is dynamically
+configured to use a particular SDK -- at this time there is only 1 SDK
+implementation, the default implementation.
+
+A goal is that the latest SDK can always be used with any version of the API, so
+that a user can always pull the latest implementation into their final Release
+to run with any API that was used in instrumented Applications within the
+Release.
+
+### OTLP Exporter (`opentelemetry_exporter`)
+
+Exporter implementations are tied to the SDK's public API.
+
+## Releases
+
+### Experimental API
+
+As noted in the previous section `opentelemetry_api_experimental` is versioned
+separately from the rest and will always remain 0.x.
+
+### API
+
+Additions to the API are released with minor version bumps.
+
+### Experimental SDK
+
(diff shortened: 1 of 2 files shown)


tsloughter:
Side note: ^ there is something really wrong with codecov

docs/rationale.md
```diff
@@ -0,0 +1,129 @@
...
+## Examples
+
+Purely for illustration purposes, not intended to represent actual releases:
+
+- v1.0.0 release:
+   - `opentelemetry_api` 1.0.0
+     - Contains APIs for tracing, baggage, propagators
+   - `opentelemetry_api_experimental` 0.2.0
+   - `opentelemetry_sdk` 1.0.0
+   - `opentelemetry_sdk_experimental` 1.0.0
```
lzchen, line 124: Is this supposed to also be `0.2.0` for ``opentelemetry_sdk_experimental`?

tsloughter: Yup, good catch!

docs/rationale.md
```diff
@@ -0,0 +1,129 @@
...
+`otel`. This means modules can move between Applications without their name
+changing and because of Erlang's flat namespace there is no code to change for a
+user when a stable API graduates from experimental to stable, if the user was
+using the latest version of the experimental API.
+
+This also allows flexibility for modules that might be in the SDK but are found
+to be better placed in the API, or vice versa. This has happened a few times in
+the pre-1.0 world as functionality was floating from SDK to API at times.
+
+## OTP Applications
```
lzchen, line 26: What does "OTP" stand for?

tsloughter: It technically stands for "Open Telecom Platform", but it is really only ever referred to as OTP these days. It is a collection of libraries and principles for creating Erlang libs that work together in a common way for ease of combining and introspection. Application specifically is documented here https://erlang.org/doc/design_principles/applications.html

I try to make it clear when Application is being used in the OTP sense vs application in the generic sense. In this doc at least "application" is never used.

hauleth (approved):

docs/rationale.md
ferd: [Open Telecom Platform](http://erlang.org/doc/system_architecture_intro/sys_arch_intro.html); It's the general development framework in Erlang and Elixir and is part of every system shipped in production. If you've ever heard of Erlang supervisors, they're part of that framework.

We tend not to re-define it in the community because that's part of the base context for almost anyone learning these languages.

docs/rationale.md
```diff
@@ -0,0 +1,129 @@
+Rationale
+---
+
+When creating a library, often times designs and decisions are made that get lost over time. This document tries to collect information on design decisions to answer common questions that may come up when you explore.
```
garthk, line 4: Choose one: wrapping lines, or not. :)

docs/rationale.md
```diff
@@ -0,0 +1,129 @@
...
+
+# Versioning and Releasing
+
+The OTP Applications and OpenTelemetry Spec itself use semver v2.
+
+In this document references to OTP concepts is done by capitalizing the word
+(for example Application and Release) while the generic term (like release) is
+lowercase).
```
garthk, line 12:
```suggestion
OTP Applications and the [OpenTelemetry Spec] itself use [semver v2].

In this document references to [OTP] concepts is done by capitalizing the word
(for example Application and Release) while the generic term (like release) is
lowercase).

[OTP]: http://erlang.org/doc/system_architecture_intro/sys_arch_intro.html
[OpenTelemetry Spec]: https://github.com/open-telemetry/opentelemetry-specification
[semver v2]: https://semver.org/spec/v2.0.0.html 
```

docs/rationale.md
```diff
@@ -0,0 +1,129 @@
...
+
+## OTP Applications
+
+### Experimental API (`opentelemetry_api_experimental`)
+
+The experimental package is where any API that is not stable when 1.0 is
+released must live. At this time (prior to 1.0) that means Metrics and Logging.
+
+This package will always be 0.x because it is never stable and modules will be
+removed when they are moved to the stable API package.
```
garthk, line 34: It's a bit over the top, but I'm a fan of RFC 2119 clarifying what we mean when we say words like “must”:
```suggestion
The experimental package is where any API that is not stable when 1.0 is
released [MUST] live. At this time (prior to 1.0) that means Metrics and Logging.

This package will always be 0.x because it is never stable and modules will be
removed when they are moved to the stable API package.

[MUST]: https://tools.ietf.org/html/rfc2119#section-1
```
I will not feel at all upset if you don't commit this suggestion.

garthk (commented):
Looking good so far!

ferd (approved):

docs/rationale.md
tsloughter: Fixed :)

chulkilee:
Shouldn't we just add a link to OTEL spec when https://github.com/open-telemetry/oteps/pull/143 is merged and added to the spec?

It would be nice to have only this package specific guides in this repo, and using the OTEL project guide as a reference, instead of duplicating contents.

tsloughter:
@chulkilee we'll want a link to the spec added when this is merged, instead of the OTEP: https://github.com/open-telemetry/opentelemetry-specification/pull/1291

docs/rationale.md
```diff
@@ -0,0 +1,137 @@
+Rationale
+---
+
+When creating a library, often times designs and decisions are made that get
+lost over time. This document tries to collect information on design decisions
+to answer common questions that may come up when you explore.
+
+# Versioning and Releasing
```
GregMefford, line 8: It looks like [the spec proposal _requires_ that this versioning information be in a file called `VERSIONING` in the root of the project](https://github.com/open-telemetry/opentelemetry-specification/pull/1291/files?file-filters%5B%5D=.md#diff-e7de8bed9eb1605a44c385930415011d900acccecc8b48398a75ce9fa707cdd3R36).

I don't have any issue _also_ having this `rationale.md` file that could cover various other topics, but maybe the stuff specifically about versioning should be moved to that file with a link from here to there.

tsloughter: Yup, good catch.

docs/rationale.md
```diff
@@ -0,0 +1,137 @@
...
+   - `opentelemetry_api` 1.0.0
+     - Contains APIs for tracing, baggage, propagators
+   - `opentelemetry_api_experimental` 0.2.0
+   - `opentelemetry_sdk` 1.0.0
+   - `opentelemetry_sdk_experimental` 0.2.0
+- v1.15.0 release (with metrics)
+   - `opentelemetry_api` 1.15.0
+     - Contains APIs for tracing, baggage, propagators, metrics
+   - `opentelemetry_sdk` 1.15.0
+
```
GregMefford, line 137:
```suggestion
- v1.0.0 release:
   - `opentelemetry_api` 1.0.0
     - Contains APIs for tracing, baggage, propagators
   - `opentelemetry_api_experimental` 0.2.0
     - Contains APIs for metrics
   - `opentelemetry_sdk` 1.0.0
   - `opentelemetry_sdk_experimental` 0.2.0
- v1.15.0 release (with metrics)
   - `opentelemetry_api` 1.15.0
     - Contains APIs for tracing, baggage, propagators, metrics
   - `opentelemetry_api_experimental` 0.42.0
     - No longer contains APIs for metrics
   - `opentelemetry_sdk` 1.15.0
   - `opentelemetry_sdk_experimental` 0.42.0
   
```

GregMefford (commented):
This seems good to me - I just made a bunch of small suggestions / tweaks.
The only major thing I think we need to sort out is moving things to a separate `VERSIONING` file at the root of the repo.

Issue #189, opened 2020-12-11
Resource Detectors


tsloughter:
The SDK needs an interface for defining resource detectors and the logic for running the defined detectors at startup https://github.com/open-telemetry/opentelemetry-specification/blob/2208a5250bac228f681010edac3df893af793096/specification/resource/sdk.md#detecting-resource-information-from-the-environment

Then specific detectors can be created as separate Applications to be included by the user.

tsloughter:
This was completed in #190

PR #190, opened 2020-12-13
add resource detection behaviour and configuration

tsloughter:


tsloughter:
Resolves #189

apps/opentelemetry/src/otel_resource.erl
```diff
@@ -55,6 +56,17 @@ attributes({otel_resource, Resource}) ->
 merge({otel_resource, CurrentResource}, {otel_resource, OtherResource}) ->
     {otel_resource, lists:ukeymerge(1, CurrentResource, OtherResource)}.
 
+
+-spec run_detectors(list()) -> t().
+run_detectors(Detectors) ->
+    lists:foldl(fun({Module, Config}, Resource) ->
```
garthk, line 62: Any chance you can also take a bare `Module`, defaulting `Config` to an empty list? Elixir flavoured, I know…

apps/opentelemetry/src/otel_resource_app_env.erl
```diff
@@ -23,10 +23,12 @@
...
 -module(otel_resource_app_env).
 
--export([get_resource/0,
+-behaviour(otel_resource_detector).
+
+-export([get_resource/1,
          parse/1]).
 
-get_resource() ->
+get_resource(_Config) ->
```
garthk, line 31: Then, you could match that empty list here.

Solid: folding the environment resources into the system. 👍

apps/opentelemetry/src/otel_resource_env_var.erl
```diff
@@ -18,14 +18,16 @@
...
+
+-export([get_resource/1,
          parse/1]).
 
 -define(OS_ENV, "OTEL_RESOURCE_ATTRIBUTES").
 -define(LABEL_LIST_SPLITTER, ",").
 -define(LABEL_KEY_VALUE_SPLITTER, "=").
 
-get_resource() ->
+get_resource(_Config) ->
```
garthk, line 30: … and here.

apps/opentelemetry/src/opentelemetry_sup.erl
```diff
@@ -33,9 +33,12 @@ init([Opts]) ->
                  intensity => 1,
                  period => 5},
 
+    Detectors = proplists:get_value(detectors, Opts, []),
+    Resource = otel_resource:run_detectors(Detectors),
+
     %% configuration server
     TracerServer = #{id => otel_tracer_server,
-                     start => {otel_tracer_provider, start_link, [otel_tracer_server, Opts]},
+                     start => {otel_tracer_provider, start_link, [otel_tracer_server, Resource, Opts]},
```
garthk, line 41: … so if any `get_resource/1` crashes, it'll take down `opentelemetry_sup` with it?

apps/opentelemetry/src/otel_resource.erl
tsloughter: Hehe, look 2 lines down, that is what it does :)

apps/opentelemetry/src/opentelemetry_sup.erl
tsloughter: Yea, at the moment. They also need a way to timeout (some of these could be hitting remote endpoints to collect attributes) and so will run as a separate process so that if they crash or just take too long the system can move on without them.

I need to verify it is ok that we simply ignore resources that timeout... as opposed to having to let it eventually update the resource when it finishes, even if it is later than the timeout.

tsloughter: agh, the spec is not clear on what it should do. going to have to ping people.

I may make it its own genserver if I understand how its used right... but spec is unclear on what should happen if fetching resources fails.

tsloughter:
Working on improving the how resource detection runs before merging this.

My first idea was a state machine that spawns each resource collection function with a timeout to receive its response.

The tracer provider will be started after the statem and call its `get_resource` function with some timeout (5 seconds?), the state machine will ignore `get_resource` messages until it has left the `collecting` state. If it times out then the Provider gets an empty resource.

ferd:
What are these resources supposed to stand for?

tsloughter:
@ferd stuff like the kubernetes namespace, pod id or ec2 instance name (so in that case fetched over http to that ec2 local endpoint that each instance exposes).

tsloughter:
I moved the detectors to run in processes so they can crash and timeout (just realized I don't have anything kill the ones that timeout, so I need to add that).

I'm not positive on what to do for resource detectors that fail or if the `get_resource` call times out -- right now they are ignored and return an empty resource, respectively.

Also the `gen_statem` is kinda funky. I started out as it would collect all results and then merge after all have been received or a timeout has triggered. It had to merge after receiving because they must merge in a certain order defined by the user, so a list of references had to be kept to then pull from the map of results in order to merge... 

I find the current code cleaner and simpler but technically the detector timeout each detector gets is multiplied by the number of detectors that go before it.

If detectors weren't defined to have to be fast I'd think that has to be changed but in this case I think the simpler collection of results is ok -- but I'm not married to it.

ferd:
If you don't really know what to do with the timeout, it might make sense to wait forever, but at least emit logs or any sort of telemetry data about which handler you're waiting on. You can't make a decision on behalf of the handler, so don't. The alternative option is to set a final timeout we a defined behaviour ("the resource is just ignored or goes without state"), but that timeout value ought to be configurable and visible to someone debugging stuff.

tsloughter:
@ferd yea, right now it is it is that we just log and ignore. I think that is fine for the time being. I'll make these configurable and then merge.

tsloughter:
Err, no, both timeouts already are configurable, so I guess I'll just merge.

PR #194, opened 2021-01-22
add configuration module to merge OS env with app env

tsloughter:
This adds support for OS variables to define propagators, samplers and log level.

Note that log level doesn't do anything yet. I *think* it should be used to set `logger:set_application_level`, but it likely needs to also expose itself through `otel_configuration` so that other opentelemetry libs, like third party exporters can utilize it as well.

### DIFF
--- a/apps/opentelemetry/src/otel_configuration.erl
+++ b/apps/opentelemetry/src/otel_configuration.erl
@@ -1,0 +1,176 @@
+%%%------------------------------------------------------------------------
+%% Copyright 2021, OpenTelemetry Authors
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
+%%
+%% http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing, software
+%% distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


apps/opentelemetry/src/otel_configuration.erl
```diff
@@ -0,0 +1,167 @@
...
+        processors(Opts))).
+
+general(Opts) ->
+    merge_list_with_environment(general_sdk, Opts).
+
+processors(AppEnvOpts) ->
+    Processors = proplists:get_value(processors, AppEnvOpts, [{otel_batch_processor, #{}}]),
+
+    Config = lists:map(fun({Name, Opts}) ->
+                               {Name, merge_with_environment(Name, Opts)}
```
ferd, line 39: Wouldn't `Opts` need to get the mapping applied first to get the name of the OS var? As it's passed in right now (AppEnvOpts's values), none of the fields look like they would contain the `{OSVar, Key, Default, Transform}` structure expected of `merge_with_environment/2`.

It might help to make a stricter type than `list()` for these.

tsloughter: `merge_with_environment` is what then calls `config_mappings(Name)` and gets the list 4-tuple.

I think I can just have the merge functions take the list of 4-tuples, instead of the name, and make the type spec stricter ,good point, will update.

tsloughter: Updated.

ferd (approved):

Issue #202, opened 2021-02-12
Setting an additional header on endpoint definition


kdisneur:
I configured this library to send traces through GRPC locally to a Jaeger-Opentelemetry container and it worked well.

Now, I tried to switch from my local Jaeger to NewRelic, the difference being they ask for an `Api-Key` in a header to authenticate the request.

I would like to avoid deploying an OpenTelemetry Collector application as I was planning to just run my application on Heroku.

I was not able to find a way to do it with the current implementation. Can you give me some pointer or at least confirm it's not supported right now?

Thanks for your help 🙇

tsloughter:
Good question... I'd be curious what other languages do for this. Maybe it is just that the exporter accepts a list of additional headers.

What exporter are you using? I didn't think there was a Jaeger exporter?

kdisneur:
I use the opentelemetry_exporter defined in this repository. As it’s only for my local development, I just send all the traces to the [experimental version of jaeger that supports opentelemetry](https://www.jaegertracing.io/docs/1.21/opentelemetry/).

i planned to use the same opentelemetry_exporter for new relic, it should only require an additional header in the HTTP request in order to work 🤞. 

i tried to read the source a little bit and discovered I could set the key channel_opts in the configuration (alongside endpoints and protocols) to define my own unary interceptor.
I don’t know a lot about gRPC but I think this the common place where people put their authentication logic?

i was able to intercept the requests but I’m now a bit stuck as I have no idea I how to update the request before calling the next handler

kdisneur:
> Maybe it is just that the exporter accepts a list of additional headers.

If I understand properly the spec, you are right, they expect the exporters to support headers: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#configuration-options

tsloughter:
Oh cool, ok, I didn't realize Jaeger or New Relic supported the OpenTelemetry exporter.

Auth is probably just headers. Tho I might have to add authenticated channels to grpcbox, shouldn't be much work. But just adding the ability to append some header would be the simplest.

tsloughter:
Ah, yup. Will get that added.

tsloughter:
This was added in #206 and I'll try to have a 0.6.1 or 1.0-rc1 release out tonight or tomorrow.

kdisneur:
Thank you very much @tsloughter 🙇 🙇

PR #206, opened 2021-02-16
exporter: support for endpoint and header configuration

tsloughter:
This is quick and a little dirty but it adds the ability to add headers to export requests, needed for #202 

I'll have another PR with support for the remaining exporter config options (certificate, compression and timeout).

### DIFF
--- a/apps/opentelemetry/src/otel_configuration.erl
+++ b/apps/opentelemetry/src/otel_configuration.erl
@@ -19,7 +19,8 @@
 %%%-------------------------------------------------------------------------
 -module(otel_configuration).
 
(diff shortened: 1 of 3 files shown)


bryannaegele:
Are we going to try to run everything through the new configuration module or will the exporter handle its own?

tsloughter:
@bryannaegele a bit of both. When I first started this patch I was going with the exporter being configured through the SDK and thus `otel_configuration` but then changed to the configuration being part of the `opentelemetry_exporter` application, so `otel_configuration` is only used as a convenience function.

bryannaegele (approved):

PR #211, opened 2021-02-18
fix unknown type spec references

tsloughter:


apps/opentelemetry/src/otel_metric_aggregator_mmsc.erl
```diff
@@ -33,16 +33,16 @@
...
+-spec update(ets:tab(), otel_meter:name() | {otel_meter:name(), otel_meter:labels()}, otel_meter:number_kind(), number()) -> boolean().
 update(Tab, Key, _, Number) ->
     update_mmsc(Tab, Key, Number).
 
--spec checkpoint(ets:tab(), otel_meter:key()) -> boolean().
+-spec checkpoint(ets:tab(), {otel_meter:name(), otel_meter:labels()}) -> boolean().
 checkpoint(Tab, NameLabelSet) ->
     MS = ets:fun2ms(fun(A=#active_instrument{key=Key,
                                              aggregator=Aggregator,
                                              current=Current}) when Key =:= NameLabelSet ,
```
bryannaegele, line 44:
```suggestion
                                             current=Current}) when Key =:= NameLabelSet,
```

bryannaegele (approved):

ferd (approved):

apps/opentelemetry/src/otel_metric_aggregator_mmsc.erl
tsloughter: Yea, I am not always consistent with style in these cases. The space was intentional since `,` is more like an operation here, but I'm fine using a style of no space.

bryannaegele: I though it was a typo since you were fixing spacing on the changed line below :)

PR #215, opened 2021-03-03
1.0.0-rc.1 - split metrics out to experimental applications

tsloughter:
This is more work then I realized.. The elixir api stuff I'm unsure about module names for. Like should everything be `OpenTelemetryExperimental.*` or just `OpenTelemetry` since that is technically legal...

This will bring us to an actual 1.0.0 RC release (0.6.x is the same api but still has metrics still in it).

Opening as a draft because I'm sure there are changes still needed in some of the Elixir modules and I just don't have the time/patience to deal with it right now, but want to get reviews anyway.


ferd (approved):
I figure the risk of keeping the same module names but a different app is that during a transition period, the experimental stuff may be brought in through transitive dependencies of some kind and then release-building will complain of module names clashing.

This however still seems nicer to fix (by excluding apps) than having to rewrite all modules for using a different name (i.e. `otelx_meter`) since it would allow drop-in replacements where possible while only changing the app name (the effect might be unpredictable with wild API changes but not much that could be done with that either way).

That being said, while that works reasonably well with Erlang, we should just go with what is the least surprise for Elixir folks, but I'm not sure I'm good to give advice there.

.github/workflows/erlang.yml
```diff
@@ -23,17 +23,18 @@ jobs:
     - uses: actions/checkout@v2
     - name: Run Collector
       run: docker-compose up -d
-    - uses: gleam-lang/setup-erlang@v1.0.0
+    - uses: erlef/setup-elixir@v1
```
GregMefford, line 26: Is this intended to also use the Elixir image? Obviously it will also have Erlang available, but just wanted to make sure this was intended.

GregMefford (approved):
In terms of prepping the main packages for a 1.0.0 RC, this looks fine to me.

As for the experimental parts, my gut feeling is that any use of experimental stuff would be an "at your own risk" situation where the user should expect to have to make changes to that in the future. That said, I think it would be nice to keep all the names looking like the normal `OpenTelemetry` modules instead of `OpenTelemetryExperimental`, so that the transition to a real release would potentially be easier on both us as maintainers and on users/integration authors who had been using the experimental versions. One concern I can think of are that if we have `Application` config settings, we may want to have them under the `_experimental` namespace so that Elixir appropriately shows warnings when you have configurations for an app that you aren't including (better DevEx when they miss adding the dependency and don't understand why it's not working).

bryannaegele:
I know we discussed shipping metrics and logging as separate packages, so I wonder if we need the experimental at all or just have a metrics package which falls under the OpenTelemetry namespace. Then as we get the metrics to 1.0, that would just roll into the main package.

apps/opentelemetry_api_experimental/mix.exs
```diff
@@ -0,0 +1,95 @@
+defmodule OpenTelemetry.MixProject do
```
GregMefford, line 1: I think this should say `Experimental`

.github/workflows/erlang.yml
tsloughter: Yea, there isn't a `erlef/setup-erlang` yet I don't think.

apps/opentelemetry_api_experimental/mix.exs
tsloughter: Yup, fixed.

tsloughter:
@bryannaegele I think having a single experimental application/package is simpler, instead of having one for metrics and logs and then having to deprecate them in hex. It doesn't make moving them to the main applications any easier as far as I can tell.

tsloughter:
@GregMefford definitely agree that configuration for metrics and logs (signals that are still experimental) should be under the experimental application.

bryannaegele:
Yeah, we talked ourselves to the same place :) I did take a look a the namespacing for Elixir though and having it under `OpenTelemetry.Metrics` would be fine and matches how Java is organizing theirs.

PR #221, opened 2021-03-11
fix tracer configuration merging

tsloughter:
Ran into this while working on some examples for docs.

This patch also makes the default exporter be undefined. I'm not sure what the best is here, when its defaulting to OTLP it crashes bad (confusing). So maybe it should instead be stdout or OTLP but handle the crash better to log a message to let the user it tried  using OTLP but something wasn't configured or available.

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   format:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
     steps:
       - uses: actions/checkout@v2
       - uses: erlef/setup-elixir@v1
@@ -35,7 +35,7 @@ jobs:
       matrix:
         otp_version: ['23.1', '22.3.4.2', '21.3.8.16']
         elixir: ['1.11.1']
-        os: [ubuntu-latest]
+        os: [ubuntu-18.04]
     env:
       OTP_VERSION: ${{ matrix.otp_version }}
       ELIXIR_VERSION: ${{ matrix.elixir }}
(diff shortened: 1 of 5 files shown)


ferd (approved):

ferd:
Approved the changeset on intent. All tests failing is necessary to fix first.

That being said, the idea of setting OTLP and getting crash logs explaining what is missing in the config would likely be best for someone wanting it to work in most cases, assuming their intent is ultimately to have that exporter going. For the purposes of docs, would having a "demo" stdout exporter make sense to configure rather than defaulting to nothing?

tsloughter:
Ugh, its the libwxgtk actions change I saw someone mention on twitter.

tsloughter:
https://twitter.com/my_olive_tree/status/1369526310448607234

@puzza007 do you know the fix?

tsloughter:
Woops, pinged Erlanger Paul Oliver instead of Erlanger @paulo-ferraz-oliveira :)

tsloughter:
Aggh, erlef/setup-elixir is also broken but because of libcrypto.

tsloughter:
So I've updated it to have some useful warning logs if the exporter fails to initialize.

I'm still unsure if we should default to OTLP exporter or something like stdout. Defaulting to the OTLP exporter, which has default host and port does make for simpler configuration in demos/guides when exporting to a local collector, but maybe that is actually worse since most people in prod will not have the collector on localhost.

paulo-ferraz-oliveira:
Yeah, for the libwx stuff I just "downgraded" to ubuntu-18.04. I'm not sure if it's solved in `gleam-lang/setup-erlang`, at the moment. But I believe I tried the latest version and other issues arose. Ah, yeah, it was the fetching of versions from Erlang Solutions that aren't there.

puzza007:
> Woops, pinged Erlanger Paul Oliver instead of Erlanger @paulo-ferraz-oliveira :)

PR #232, opened 2021-03-24
Fix bug preventing configuring span processors

garthk:
If I specify a configuration key under `processors` other than `otel_batch_processor`, it crashes with a function clause error. I trust that's unintentional. :)


tsloughter:
Yup. Thanks!

tsloughter (approved):

PR #242, opened 2021-04-23
add auto registration of all loaded applications as tracers

tsloughter:
I really thought this was already added, but apparently not! Now all applications will have tracers created for them automatically unless disabled in configuration. I don't think it should be an issue to default to just creating all of them, even if ones like `kernel` will never actually be used.


hauleth (approved):

bryannaegele:
There's no issue if they're called twice, correct? All of the libraries currently register themselves.

tsloughter:
@bryannaegele correct

PR #243, opened 2021-04-23
opentelemetry.io: instrumentation docs page

tsloughter:


website_docs/instrumentation.md
```diff
@@ -0,0 +1,375 @@
...
+OpenTelemetry.register_application_tracer(:myapp)
+{{< /tab >}}
+
+{{< /tabs >}}
+
+Giving names to each `Tracer`, and in the case of Erlang/Elixir having that name
+be the name of the Application, allows for the ability to blacklist traces from
+a particular Application. This can be useful if, for example, a dependency turns
+out to be generating too many or in some way problematic spans and it is desired
+to disable their generation.
```
ferd, line 47: Is the matching of app-to-tracer done on a module basis or on a group_leader/process hierarchy basis? I.e. if I make an app an included app, or that I call code from a library app within a stateful app, which is it that shows up?

website_docs/instrumentation.md
```diff
@@ -0,0 +1,375 @@
...
+{{< tab >}}
+tracer = OpenTelemetry.get_tracer(:my_app)
+span_ctx = OpenTelemetry.Tracer.start_span(tracer, "hello-world", %{})
+...
+OpenTelemetry.Tracer.end_span(span_ctx)
+{{< /tab >}}
+
+{{< /tabs >}}
+
+In most cases you will not need to manually register or lookup a
```
ferd, line 76:
```suggestion
In most cases you will not need to manually register or look up a
```

website_docs/instrumentation.md
```diff
@@ -0,0 +1,375 @@
...
+text_map_propagators:
+  [&:otel_baggage.get_text_map_propagators/0,
+   &:otel_tracer_default.w3c_propagators/0],
+...
+{{< /tab >}}
+
+{{< /tabs >}}
+
+If you instead need to use the [B3 protocol](), originally from the [Zipkin
+project](), then replace `otel_tracer_default:w3c_propagators/0` and
```
ferd, line 358: missing links

ferd (approved):
This is going to be very useful for users; there are two missing links to fix, so I'm approving this conditional to their inclusion.

website_docs/instrumentation.md
tsloughter: It is a module basis. A persistent term is used to map module to tracer so that the macro can lookup the tracer with just `?MODULE`.

I wasn't sure how detailed I should get about how things actually work in this document.

ferd: Yeah that's probably fine and in this case it does exactly what's on the label, I was checking for accuracy.

website_docs/instrumentation.md
```diff
@@ -0,0 +1,377 @@
...
+                        
+                        %% do work here
+                        
+                        ?end_span(SpanCtx)
+                    end),
+{{< /tab >}}
+
+{{< tab >}}
+span_ctx = OpenTelemetry.Tracer.start_span(<<"child">>)
+Ctx = OpenTelemetry.Ctx.get_current()
```
hauleth, line 176:
```suggestion
ctx = OpenTelemetry.Ctx.get_current()
```

website_docs/instrumentation.md
```diff
@@ -0,0 +1,377 @@
...
+                        ?with_span(<<"other-process">>, #{links => [Link]},
+                                   fun() -> ok end)
+                    end),
+{{< /tab >}}
+
+{{< tab >}}
+task = Task.async(fn ->
+                    # a new process has a new context so the span created
+                    # by the following `with_span` will have no parent
+                    link = OpenTelemetry.link(parent)
```
hauleth, line 218: Where is `parent` coming from? There probably should be line like:

```elixir
parent = OpenTelemetry.current_span_ctx()
```

Before the `Task.async/1` call.

tsloughter: Derp, yup, fixing. Clearly I need to put all these examples into the getting started repos I have and actually run them :)

PR #246, opened 2021-05-12
Better compliance with status spec

bryannaegele:
Statuses are not allowed to be freeform and descriptions are only acceptable for errors. For the other statuses, descriptions are to be ignored.

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#set-status

### DIFF
--- a/apps/opentelemetry_api/include/opentelemetry.hrl
+++ b/apps/opentelemetry_api/include/opentelemetry.hrl
@@ -30,6 +30,10 @@
 -define(SPAN_KIND_PRODUCER, producer).
 -define(SPAN_KIND_CONSUMER, consumer).
 
+-define(OTEL_STATUS_UNSET, unset).
+-define(OTEL_STATUS_OK, ok).
+-define(OTEL_STATUS_ERROR, error).
+
 -record(span_ctx, {
           %% 128 bit int trace id
           trace_id          :: opentelemetry:trace_id(),
@@ -70,11 +74,7 @@
          }).
 
 -record(status, {
-          code    :: atom() | integer(),
+          code = ?OTEL_STATUS_UNSET :: opentelemetry:status_code(),
           %% developer-facing error message
-          message :: unicode:unicode_binary()
+          message = <<"">>          :: unicode:unicode_binary()
          }).
-
--define(OTEL_STATUS_UNSET, unset).
--define(OTEL_STATUS_OK, ok).
--define(OTEL_STATUS_ERROR, error).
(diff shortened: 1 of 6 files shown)


bryannaegele:
One callout is the spec calls for the API to have the function be called `set_status` taking the arguments we're currently using to build the status record and then passing to this function. I'm reluctant to make the change at this point but wanted to point it out if anyone cares.

apps/opentelemetry_api/include/opentelemetry.hrl
```diff
@@ -70,11 +74,7 @@
          }).
 
 -record(status, {
-          code    :: atom() | integer(),
+          code = ?OTEL_STATUS_UNSET :: opentelemetry:status_code(),
           %% developer-facing error message
-          message :: unicode:unicode_binary()
+          message = <<"">>          :: unicode:unicode_binary()
```
hauleth, line 79: Why not `unicode:chardata/0`? It would make the API a little bit simpler for the Erlang users.

hauleth (approved):

apps/opentelemetry_api/include/opentelemetry.hrl
bryannaegele: If I recall, we ran into some interop issues in some places around unicode charlists versus binary, so it settled out to using the binary everywhere as the default. I don't think it was this specific place (I believe it was around headers) but I think that was the origin of just doing the unicode binary.

apps/opentelemetry_api/test/opentelemetry_api_SUITE.erl
```diff
@@ -120,7 +120,7 @@ update_span_data(_Config) ->
     Events = opentelemetry:events([{opentelemetry:timestamp(),
                                     <<"timed-event-name">>, []}]),
     Status = opentelemetry:status(?OTEL_STATUS_OK, <<"This is Ok">>),
-    ?assertMatch(#status{code = ?OTEL_STATUS_OK, message = <<"This is Ok">>}, Status),
+    ?assertMatch(#status{code = ?OTEL_STATUS_OK, message = <<"">>}, Status),
```
tsloughter, line 123: What happened to the message set in the call above?

bryannaegele: I just updated the tests, but the spec says only an error status can have a description. For Ok and Unset, any description provided by the user is to be ignored.

> Description MUST be IGNORED for StatusCode Ok & Unset values.

tsloughter: aaaah, ok

apps/opentelemetry_api/include/opentelemetry.hrl
tsloughter: My only other question before approving is related, why not allow `undefined`?

bryannaegele: I think we'd need to check the exporter would work ok with that since it has to export a binary for that either way, but it wasn't allowed to begin with. There was always a binary value guard on it. I was just aiming to tighten down the allowed values for the status and the description with this.

PR #253, opened 2021-05-30
Remove unused header in otel_span_ets

wingyplus:
The warning detected by erlang_ls that `otel_tracer.hrl` is unused in
otel_span_ets module. This change remove this include.

PR #254, opened 2021-05-31
Upgrade Elixir and OTP in GitHub Workflow

wingyplus:
This change do:

* Add Elixir 1.12.1 and upgrade 1.11 to latest release.
* Add OTP 24.0.1 and upgrade major OTP release to latest release.
* Use rebar3 v1.15.2 for OTP 21. And v1.16.1 for OTP 22+.

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -41,6 +41,7 @@ jobs:
           - otp_version: '21.3.8.24'
             elixir_version: '1.11.4'
             rebar3_version: '3.15.2'
+            os: ubuntu-18.04
         exclude:
           - otp_version: '21.3.8.24'
             elixir: '1.12.1'
@@ -73,6 +74,7 @@ jobs:
           - otp_version: '21.3.8.24'
             elixir_version: '1.11.4'
             rebar3_version: '3.15.2'
+            os: ubuntu-18.04
         exclude:
           - otp_version: '21.3.8.24'
             elixir: '1.12.1'
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -33,16 +33,16 @@ jobs:
     name: Test on Elixir ${{ matrix.elixir_version }} (OTP ${{ matrix.otp_version }}) and ${{ matrix.os }}
     strategy:
       matrix:
-        otp_version: ['24.0.2', '23.3.4.1', '22.3.4.19']
+        otp_version: ['24.0.2', '23.3.4.2', '22.3.4.20']
         elixir: ['1.12.1', '1.11.4']
         rebar3_version: ['3.16.1']
         os: [ubuntu-18.04]
         include:
-          - otp_version: '21.3.8.23'
+          - otp_version: '21.3.8.24'
             elixir_version: '1.11.4'
             rebar3_version: '3.15.2'
         exclude:
-          - otp_version: '21.3.8.23'
+          - otp_version: '21.3.8.24'
             elixir: '1.12.1'
     env:
       OTP_VERSION: ${{ matrix.otp_version }}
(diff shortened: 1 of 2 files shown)


hauleth (approved):

hauleth:
Why remove OTP 21?

.github/workflows/elixir.yml
```diff
@@ -33,8 +33,8 @@ jobs:
     name: Test on Elixir ${{ matrix.elixir_version }} (OTP ${{ matrix.otp_version }}) and ${{ matrix.os }}
     strategy:
       matrix:
-        otp_version: ['23.1', '22.3.4.2', '21.3.8.16']
-        elixir: ['1.11.1']
+        otp_version: ['24.0.1', '23.3.4.1', '22.3.4.19']
```
hauleth, line 36: As we are supporting Elixir 1.11 we shouldn't remove OTP 21 from matrix, just ignore such set of options.

hauleth (changes requested):

wingyplus:
> Why remove OTP 21?

The discussion is in #253. And Elixir 1.12 requires to run OTP 21 or greater.  That’s why I decide to drop it. Or do you have any suggestions?

hauleth:
We can exclude only Elixir 1.12 and OTP 21 and still run them - https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-excluding-configurations-from-a-matrix

wingyplus:
> We can exclude only Elixir 1.12 and OTP 21 and still run them - https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-excluding-configurations-from-a-matrix

@hauleth Thanks for your suggestion. Do you mind if I upgrade to 21.3.8.23?

hauleth:
Not at all.

wingyplus:
The OTP 21 is added back to the CI on both Elixir and OTP workflow. Please kindly review.

hauleth (approved):

wingyplus:
I’ve found similar issue with #253. Not sure why it happens.

```
escript: exception error: undefined function rebar3:main/1
8
  in function  escript:run/2 (escript.erl, line 758)
9
  in call from escript:start/1 (escript.erl, line 277)
10
  in call from init:start_em/1 
11
  in call from init:do_boot/3 
12
Error: Process completed with exit code 127.
```

tsloughter:
I thought it was because of rebar3 only being built on recent otp's now (otp-24 dropped deprecated stuff so couldn't use older builds with it). But I Thought 21 was new enough...

@ferd ?

ferd:
We used the opportunity of clamping down on the rebar3 vs OTP compatibility window to just line up directly with what the OTP team supports all at once rather than doing it progressively, with the rationale that OTP24's JIT would provide enough of an incentive.

Versions 3.16.x are 22..24 so far.

wingyplus:
@ferd i see in you mentioned in the blog (https://ferd.ca/you-ve-got-to-upgrade-rebar3.html) that 3.15.2 support OTP 19-23. Can we use that one?

I just see that setup-beam allow us to set rebar3 version (https://github.com/erlef/setup-beam#matrix-example-rebar3). So if we want to keep otp21, we might set the rebar3 to 3.15.2 to the pipeline and use the latest version for otp 22+.

And I notice that erlef change to setup-beam. Do you need me to change it instead of setup-elixir?

ferd:
There's no reason it wouldn't work to use 3.15.2 for OTP-21 specifically. We cut the security release on that one specifically to still provide support as far back as posslbe (and eventually added 13.3.x with user contribs as well for people either needing older behaviour or OTP-18). 

For a while this will be plenty, but over time the differences may accumulate and cause divergences.

tsloughter:
Aaah, 22, not 21.

wingyplus:
@tsloughter @hauleth I have a fix for OTP 21 issue by using OTP 21 with rebar3 3.15.2 instead. Please kindly review.

hauleth (approved):

ferd:
Is it normal that the 4 checks that passed don't contain Erlang?

Looking at the checks I can see the following is broken for Erlang: https://github.com/open-telemetry/opentelemetry-erlang/actions/runs/896362703
image
And for Elixir:
image

I think we can't merge this, it appears the CI is broken but still showing as 4 successful checks.

ferd (changes requested):
The CI steps appear to still be broken.

wingyplus:
Umm, if i understand it correctly. The run-on empty because I forget adding os into the include.

Let me investigate.

wingyplus:
I've added `os` into the include matrix. This should fix the problem can we try to trigger the pipeline?

ferd:
Done, it seems to be running them all now.

wingyplus:
@ferd Thanks for your help. It's because I accidentally change Elixir to 1.12 in Erlang dialyzer which does not require since change to setup-beam. And I found the `matrix.elixir` variable is inconsistent, which cause the wrong test name. 

All of this should fixed in latest changes.

ferd (approved):
Everything seems to work successfully now, let's get this merged!

wingyplus:
@hauleth @ferd @tsloughter Thanks for your help. :)

hauleth:
💚 💙 💜 💛 ❤️

Issue #247, opened 2021-05-14
Add schema url to Tracer


tsloughter:
Schema URLs have been added to the spec and must be supported as a property on Tracers and included in OTLP exports.

tsloughter:
Closed by #331

PR #256, opened 2021-06-28
adding schema_url to the resources

fr0stbyte:
addressing issue #247

### DIFF
--- a/apps/opentelemetry/src/otel_resource.erl
+++ b/apps/opentelemetry/src/otel_resource.erl
@@ -19,12 +19,15 @@
 -module(otel_resource).
 
 -export([create/1,
+         create/2,
          merge/2,
-         attributes/1]).
+         attributes/1,
+         schema_url/1]).
 
 -type key() :: io_lib:latin1_string().
 -type value() :: io_lib:latin1_string() | integer() | float() | boolean().
--type resource() :: {otel_resource, [{key(), value()}]}.
+-type schema() :: uri_string:uri_string() | undefined.
+-type resource() :: {otel_resource, schema(), [{key(), value()}]}.
 -opaque t() :: resource().
 
 -export_type([t/0]).
@@ -34,9 +37,16 @@
 %% verifies each key and value and drops any that don't pass verification
 -spec create(#{key() => value()} | [{key(), value()}]) -> t().
 create(Map) when is_map(Map) ->
-    create(maps:to_list(Map));
+    create(Map, undefined);
 create(List) when is_list(List) ->
-    {otel_resource, lists:filtermap(fun({K, V}) ->
+    create(List, undefined).
+
+%% verifies each key and value and drops any that don't pass verification
+-spec create(#{key() => value()} | [{key(), value()}], schema() ) -> t().
+create(Map, SchemaUrl) when is_map(Map) ->
+    create(maps:to_list(Map), SchemaUrl);
+create(List, SchemaUrl) when is_list(List) ->
+    {otel_resource, SchemaUrl, lists:filtermap(fun({K, V}) ->
                                     %% TODO: log an info or debug message when dropping?
                                     case check_key(K) andalso check_value(V) of
                                         {true, Value} ->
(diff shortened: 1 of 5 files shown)


apps/opentelemetry/src/otel_resource.erl
```diff
@@ -47,15 +56,30 @@ create(List) when is_list(List) ->
...
+%% check two schemas
+-spec check_schema(schema(), schema()) -> schema().
+check_schema(A, A) ->
+    A;
+check_schema(undefined, A) ->
+    A;
+check_schema(A, undefined) ->
+    A;
+check_schema(_, _) ->
+    throw({error, 'schemas do not match'}).
```
fr0stbyte, line 82: not sure if this is the right approach here

fr0stbyte:
this doesn't yet add the schema url to the OTLP exports

apps/opentelemetry/src/otel_resource.erl
ferd: We only use `throw()` for local flow control, `error()` for something unfixable without a config or code change, and a return value (`{ok, A} | {error, schema_mismatch}`) for something we expect the caller to be able to handle.

The "reason" is something people are expected to use in a pattern match (either in a `case` in the returned format or in a `catch` pattern within a `try...catch`), so we tend to advise the use of normal atoms, such as `schema_mismatch` as opposed to `'schemas do not match'`.

apps/opentelemetry/src/otel_resource.erl
```diff
@@ -47,15 +56,30 @@ create(List) when is_list(List) ->
...
 -spec attributes(t()) -> [{key(), value()}].
-attributes({otel_resource, Resource}) ->
+attributes({otel_resource, _, Resource}) ->
     Resource.
 
+-spec schema_url(t()) -> schema().
+schema_url({otel_resource, SchemaUrl, _}) ->
+    SchemaUrl.
+
 %% In case of collision the current, first argument, resource takes precedence.
```
ferd, line 66: Shouldn't this comment indicate that in case of conflict, the first argument resource should take precedence, rather than trigger an exception?

apps/opentelemetry/test/otel_resource_SUITE.erl
```diff
@@ -113,6 +113,33 @@ combining(_Config) ->
...
+    Resource2 = otel_resource:create(otel_resource_env_var:parse("service.name=cttest,service.version=1.1.1"),
+                                        "https://opentelemetry.io/schemas/undefined"),
+
+    Merged = otel_resource:merge(Resource1, Resource2),
+    Expected = {otel_resource, undefined,
+                                [{<<"service.name">>, <<"other-name">>},
+                                {<<"service.version">>, <<"1.1.1">>}]},
+    ?assertEqual(Expected, Merged),
+    ok.
+
```
ferd, line 142: There is a big change in the type that looks for a schema URL, which we see in the return types, but one of the functions in play and all of the conflict management implemented isn't being checked. 

I would have expected, in the first test, to have one schema win since this should be a string vs `undefined` case, and in the second test, that the returned resource wouldn't be `undefined` given clashing values are passed to `create/2`.

I'm expecting CI to fail once the syntax errors are corrected.

apps/opentelemetry/src/otel_resource.erl
```diff
@@ -47,15 +56,30 @@ create(List) when is_list(List) ->
                                end, lists:ukeysort(1, List))}.
 
 -spec attributes(t()) -> [{key(), value()}].
-attributes({otel_resource, Resource}) ->
+attributes({otel_resource, _, Resource}) ->
     Resource.
 
+-spec schema_url(t()) -> schema().
+schema_url({otel_resource, SchemaUrl, _}) ->
```
ferd, line 63: this would need exporting

apps/opentelemetry/src/otel_resource.erl
fr0stbyte: Afaict, it's not the same behavior as the other OTEL sdks. I checked go and java and they both return error / throw.

fr0stbyte: I was wrong about that. fixing

apps/opentelemetry/src/otel_resource.erl
```diff
@@ -47,15 +57,31 @@ create(List) when is_list(List) ->
...
+
 %% In case of collision the current, first argument, resource takes precedence.
+%% In case of schema conflict, it will be set to undefined
 -spec merge(t(), t()) -> t().
-merge({otel_resource, CurrentResource}, {otel_resource, OtherResource}) ->
-    {otel_resource, lists:ukeymerge(1, CurrentResource, OtherResource)}.
+merge({otel_resource, CurrentSchema, CurrentAttributes}, {otel_resource, OtherSchema, OtherAttributes}) ->
+    %% check schema url
+    Schema = check_schema(CurrentSchema, OtherSchema),
+    {otel_resource, Schema, lists:ukeymerge(1, CurrentAttributes, OtherAttributes)}.
```
tsloughter, line 73: Should we be merging the attributes and keeping them even if the Schema is turned into `undefined`? It seems the spec says it is implementation specific what to do here?

I guess we should check what other implementations are doing.

apps/opentelemetry/src/otel_resource.erl
```diff
@@ -47,15 +57,31 @@ create(List) when is_list(List) ->
...
 
-%%
+%% check two schemas. Current behavior is undefined if schemas do not match
+-spec check_schema(schema(), schema()) -> schema().
+check_schema(A, A) ->
```
tsloughter, line 77: Since `uri_string` is `iodata` this check isn't sufficient. To ensure accuracy we at least need to use `string:compare` (I guess with ignore case set to true?) and probably first call `uri_string:normalize` on the url.

tsloughter:
Thank for this but there were a lot of changes that would make updating this particular work painful. So I opened a new PR that adds schema_url to resources, tracer and supports them in OTLP exporting: https://github.com/open-telemetry/opentelemetry-erlang/pull/331

I'll be closing this PR, thanks again.

PR #257, opened 2021-06-29
Add guard to register_tracer

derekkraan:
Without this, the exporter fails later if you send the second argument
as a float, for example.

### DIFF
--- a/apps/opentelemetry_api/src/opentelemetry.erl
+++ b/apps/opentelemetry_api/src/opentelemetry.erl
@@ -133,7 +133,7 @@ set_tracer(Name, Tracer) ->
 
 -spec register_tracer(atom(), string() | binary()) -> boolean().
 register_tracer(Name, Vsn) when is_atom(Name) and is_binary(Vsn) ->
-    otel_tracer_provider:register_tracer(Name, Vsn).
+    otel_tracer_provider:register_tracer(Name, Vsn);
 register_tracer(Name, Vsn) when is_atom(Name) and is_list(Vsn) ->
     otel_tracer_provider:register_tracer(Name, list_to_binary(Vsn)).
 
--- a/apps/opentelemetry_api/src/opentelemetry.erl
+++ b/apps/opentelemetry_api/src/opentelemetry.erl
@@ -132,7 +132,7 @@ set_tracer(Name, Tracer) ->
     verify_and_set_term(Tracer, Name, otel_tracer).
 
 -spec register_tracer(atom(), string()) -> boolean().
-register_tracer(Name, Vsn) ->
+register_tracer(Name, Vsn) when is_atom(Name) and is_binary(Vsn) ->
     otel_tracer_provider:register_tracer(Name, Vsn).
 
 -spec register_application_tracer(atom()) -> boolean().
--- a/apps/opentelemetry_api/src/opentelemetry.erl
+++ b/apps/opentelemetry_api/src/opentelemetry.erl
@@ -131,9 +131,11 @@ set_default_tracer(Tracer) ->
 set_tracer(Name, Tracer) ->
     verify_and_set_term(Tracer, Name, otel_tracer).
 
--spec register_tracer(atom(), string()) -> boolean().
+-spec register_tracer(atom(), string() | binary()) -> boolean().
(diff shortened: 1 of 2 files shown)


apps/opentelemetry_api/src/opentelemetry.erl
```diff
@@ -132,7 +132,7 @@ set_tracer(Name, Tracer) ->
     verify_and_set_term(Tracer, Name, otel_tracer).
 
 -spec register_tracer(atom(), string()) -> boolean().
-register_tracer(Name, Vsn) ->
+register_tracer(Name, Vsn) when is_atom(Name) and is_binary(Vsn) ->
```
ferd, line 135: hm, this clashes with the `string()` type spec on `Vsn` which is not a supertype of `binary()` as the guard asserts. The type is also repeated when declaring the behaviour: https://github.com/open-telemetry/opentelemetry-erlang/blob/49f0dd2294e16af27700b6e9c528cbcb76c98e0f/apps/opentelemetry_api/src/otel_tracer_provider.erl#L35 
in the server implementing it: https://github.com/open-telemetry/opentelemetry-erlang/blob/49f0dd2294e16af27700b6e9c528cbcb76c98e0f/apps/opentelemetry_api/src/otel_tracer_provider.erl#L53

All documentation however uses the binary format, including tests: https://github.com/open-telemetry/opentelemetry-erlang/blob/c3ef9aabf4ffc335a535e5299d6d71ec58c9ab95/apps/opentelemetry/test/opentelemetry_SUITE.erl#L89

I think we at least need to normalize the types to all be binaries, or at least to provide both types and convert to binary internally.

bryannaegele: I agree with @ferd. For `vsn` we should probably accept both and convert to binary, then fix up any internal specs to be binary. That way any libs doing auto-registry don't have to do any conversion in either language.

tsloughter: Yea, definitely accept both.

derekkraan: Thanks for the feedback. I've added a commit. I'm not at all well versed in Erlang, so if I've made a simple mistake then please bear with me.

bryannaegele:
@derekkraan you will need to sign the CLA. The link is posted above by the bot.

derekkraan:
@bryannaegele done

bryannaegele (approved):

apps/opentelemetry_api/src/opentelemetry.erl
```diff
@@ -131,9 +131,11 @@ set_default_tracer(Tracer) ->
 set_tracer(Name, Tracer) ->
     verify_and_set_term(Tracer, Name, otel_tracer).
 
--spec register_tracer(atom(), string()) -> boolean().
-register_tracer(Name, Vsn) ->
+-spec register_tracer(atom(), string() | binary()) -> boolean().
+register_tracer(Name, Vsn) when is_atom(Name) and is_binary(Vsn) ->
     otel_tracer_provider:register_tracer(Name, Vsn).
+register_tracer(Name, Vsn) when is_atom(Name) and is_list(Vsn) ->
+    otel_tracer_provider:register_tracer(Name, list_to_binary(Vsn))
```
bryannaegele, line 138:
```suggestion
    otel_tracer_provider:register_tracer(Name, list_to_binary(Vsn)).
```

apps/opentelemetry_api/src/opentelemetry.erl
```diff
@@ -131,9 +131,11 @@ set_default_tracer(Tracer) ->
 set_tracer(Name, Tracer) ->
     verify_and_set_term(Tracer, Name, otel_tracer).
 
--spec register_tracer(atom(), string()) -> boolean().
-register_tracer(Name, Vsn) ->
+-spec register_tracer(atom(), string() | binary()) -> boolean().
+register_tracer(Name, Vsn) when is_atom(Name) and is_binary(Vsn) ->
     otel_tracer_provider:register_tracer(Name, Vsn).
```
bryannaegele, line 136:
```suggestion
    otel_tracer_provider:register_tracer(Name, Vsn);
```

PR #258, opened 2021-07-23
improve exporter failure logs and include stacktrace

tsloughter:


bryannaegele (approved):

ferd (approved):

apps/opentelemetry/src/otel_batch_processor.erl
```diff
@@ -239,23 +239,44 @@ init_exporter({ExporterModule, Config}) when is_atom(ExporterModule) ->
         ignore ->
             undefined
     catch
-        error:badarg when ExporterModule =:= opentelemetry_exporter ->
+        error:badarg:StackTrace when ExporterModule =:= opentelemetry_exporter ->
+            %% logging in debug level since config argument in stacktrace could have secrets
+            ?LOG_DEBUG("OTLP tracer, ~p, failed to initialize with exception error:badarg. stacktrace=~p",
+                       [ExporterModule, StackTrace]),
```
hauleth, line 245: Why not use structured logging like that:

```suggestion
            ?LOG_DEBUG(#{source => tracer, reason => initialization_failure, kind => error, reason => badarg, stacktrace => StackTrace}, #{report_cb => fun ?MODULE:report_cb/1}),
```

And then add `report_cb/1` hidden public function. This should provide more context for experienced users and should also improve potential future logging system integration.

tsloughter: Yea, it should be. Couldn't remember the `report_cb` stuff off the top of my head so just opened this to get something out. Obv then it stalled so may as well get the structured logs done :).

tsloughter:
Haven't got the best practices for the new logger stuff down yet, but hopefully this is good enough to merge for now. @hauleth ?

PR #260, opened 2021-08-03
Introduce otel_sampler:new/3 and fix behavior typespec

dvic:
This PR contains the following changes

- added `otel_sampler:new/3`
  *because `otel_sampler.t()` is opaque, we need `new/3` to be able to create a sampler from the `setup/1` callback*

- introduced a type `sampler_opts` for the sampler options

- introduce a type `sampler_fun` for the sampler decision function

- fixed the otel_sampler `setup/2` callback to be `setup/1`
  *this callback accepts only the sampler options*

### DIFF
--- a/apps/opentelemetry/src/otel_sampler.erl
+++ b/apps/opentelemetry/src/otel_sampler.erl
@@ -19,7 +19,8 @@
 %%%-------------------------------------------------------------------------
 -module(otel_sampler).
 
--export([setup/1,
+-export([new/3,
+         setup/1,
          get_description/1,
          always_on/7,
          always_off/7,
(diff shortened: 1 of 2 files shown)


apps/opentelemetry/src/otel_sampler.erl
```diff
@@ -31,26 +32,35 @@
...
 -type sampling_result() :: {sampling_decision(), opentelemetry:attributes(), opentelemetry:tracestate()}.
 -type description() :: unicode:unicode_binary().
--type sampler() :: {fun((otel_ctx:t(),
-                         opentelemetry:trace_id(),
-                         opentelemetry:links(),
-                         opentelemetry:span_name(),
-                         opentelemetry:span_kind(),
-                         opentelemetry:attributes(),
-                         term()) -> sampling_result()), description(), term()}.
+-type decider() :: fun((otel_ctx:t(),
```
tsloughter, line 40: I think `sampler_fun` would be a more descriptive/clear name for this type.

dvic: 👍

tsloughter (approved):

PR #263, opened 2021-08-05
Fix root span sampling

### LINKED ISSUE #262: root spans are not recorded with trace_id_ratio_based
I'm not sure what should be done here but this doesn't seem right does it?

```erlang
trace_id_ratio_based_sampler(_Config) ->
    Tracer = opentelemetry:get_tracer(),

    Sampler1 = otel_sampler:setup(always_on),

    %% this succeeds
    SpanCtx1 = otel_tracer:start_span(Tracer, <<"span-1">>, #{sampler => Sampler1}),
    ?assertMatch(true, SpanCtx1#span_ctx.is_recording),

    Sampler2 = otel_sampler:setup({trace_id_ratio_based, 1.0}),

    %% this fails
    SpanCtx2 = otel_tracer:start_span(Tracer, <<"span-1">>, #{sampler => Sampler2}),
    ?assertMatch(true, SpanCtx2#span_ctx.is_recording),

    ok.
```

It's because `ParentTraceId = trace_id(Ctx),` is `undefined` (in `otel_span_utils`), and that's because `start_span` inside `opentelemetry/otel_tracer`:

```erlang
start_span(Tracer={Module, _}, Name, Opts) ->
    Module:start_span(otel_ctx:get_current(), Tracer, Name, Opts).
```

in which `otel_ctx:get_current()` here returns `#{}`.

dvic:
This fixes #262

### DIFF
--- a/apps/opentelemetry/test/opentelemetry_SUITE.erl
+++ b/apps/opentelemetry/test/opentelemetry_SUITE.erl
@@ -461,7 +461,7 @@ default_sampler(_Config) ->
 root_span_sampling_always_off(_Config) ->
     Tracer = opentelemetry:get_tracer(),
 
-    Sampler = otel_sampler:setup(always_off),
+    Sampler = otel_sampler:new(always_off),
 
     SpanCtx1 = otel_tracer:start_span(Tracer, <<"span-1">>, #{sampler => Sampler}),
     ?assertMatch(false, SpanCtx1#span_ctx.is_recording),
@@ -477,7 +477,7 @@ root_span_sampling_always_off(_Config) ->
 root_span_sampling_always_on(_Config) ->
     Tracer = opentelemetry:get_tracer(),
 
-    Sampler = otel_sampler:setup(always_on),
+    Sampler = otel_sampler:new(always_on),
 
     SpanCtx1 = otel_tracer:start_span(Tracer, <<"span-1">>, #{sampler => Sampler}),
     ?assertMatch(true, SpanCtx1#span_ctx.is_recording),
--- a/apps/opentelemetry/test/opentelemetry_SUITE.erl
+++ b/apps/opentelemetry/test/opentelemetry_SUITE.erl
@@ -19,9 +19,9 @@ all() ->
 
 all_testcases() ->
     [disable_auto_registration, registered_tracers, with_span, macros, child_spans,
-     update_span_data, tracer_instrumentation_library,
-     tracer_previous_ctx, stop_temporary_app, reset_after, attach_ctx, default_sampler,
-     record_but_not_sample, record_exception_works, record_exception_with_message_works].
+     update_span_data, tracer_instrumentation_library, tracer_previous_ctx, stop_temporary_app,
+     reset_after, attach_ctx, default_sampler, root_span_sampling, record_but_not_sample,
+     record_exception_works, record_exception_with_message_works].
 
 groups() ->
     [{w3c, [], [propagation]},
@@ -454,6 +454,20 @@ default_sampler(_Config) ->
     %% local not sampled but is recorded should default to sampled
     SpanCtx6 = otel_tracer:start_span(Tracer, <<"span-6">>, #{}),
     ?assertMatch(false, SpanCtx6#span_ctx.is_recording),
+
+    ok.
+
+root_span_sampling(_Config) ->
+    Tracer = opentelemetry:get_tracer(),
+
+    Sampler = otel_sampler:setup({trace_id_ratio_based, 1.0}),
+
+    SpanCtx1 = otel_tracer:start_span(Tracer, <<"span-1">>, #{sampler => Sampler}),
+    ?assertMatch(true, SpanCtx1#span_ctx.is_recording),
+
+    otel_tracer:set_current_span(SpanCtx1),
+    SpanCtx2 = otel_tracer:start_span(Tracer, <<"span-2">>, #{sampler => Sampler}),
+    ?assertMatch(true, SpanCtx2#span_ctx.is_recording),
 
     ok.
 
--- a/apps/opentelemetry/test/opentelemetry_SUITE.erl
+++ b/apps/opentelemetry/test/opentelemetry_SUITE.erl
@@ -20,8 +20,8 @@ all() ->
 all_testcases() ->
     [disable_auto_registration, registered_tracers, with_span, macros, child_spans,
      update_span_data, tracer_instrumentation_library, tracer_previous_ctx, stop_temporary_app,
-     reset_after, attach_ctx, default_sampler, root_span_sampling, record_but_not_sample,
-     record_exception_works, record_exception_with_message_works].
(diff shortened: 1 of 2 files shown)


dvic:
I'm not sure if `sample/7` should receive `Ctx` or `NewSpanCtx`?

dvic:
I think I've fixed it properly now.

tsloughter:
Yup, I think this is right. Thanks! I've got to look a little closer but will merge soon I think.

dvic:
I converted this PR back to a draft. You were right to have a little closer look ;) because there still a problem: I noticed that sometime I would have missing root spans, it seems like this PR broke the other use case: when you do have parent spans, then they are not exported it seems?

dvic:
Never mind, I was testing with also #261 and that seems to be actually the problem, if I use both PRs but don't use any samplers then there are no issues.

apps/opentelemetry/test/opentelemetry_SUITE.erl
```diff
@@ -457,6 +457,20 @@ default_sampler(_Config) ->
 
     ok.
 
+root_span_sampling(_Config) ->
```
tsloughter, line 460: Can you throw in a test that has a `0.0` ratio?

tsloughter (approved):

dvic:
Please merge #264 first, it fixes the failing test in this PR (the underlying issue is not introduced in this PR, hence the separate PR).

apps/opentelemetry/test/opentelemetry_SUITE.erl
dvic: There is now a `always_on` and `always_off` test.

tsloughter:
@dvic the test failure, and I think the issue that you opened the next PR about, I think is solved by fixing the setting of `is_remote` on new spans. I can get a PR together tomorrow that should fix it and thus make this PR merge-able and your PR #264 no longer needed.

dvic:
@tsloughter could it be that there is a bug (introduced by this PR/related) with the sweeper? We're running a long process where we use `always_off` as the root span and the `otel_span_table` ETS table just keeps growing (500+ MB). I haven't looked into it yet myself but I thought it might ring a bell?

tsloughter:
@dvic are you sure the sweeper is running?

Not recorded is a tricky situation. Originally a span that wasn't sampled wouldn't be inserted into the ets table, but the addition of "non-recording" complicated that. I think this was done to be safe but it may be the case that we don't need it.

I'll see if I can find anything in the spec that relates to a non-recording span being used that would mean we need to keep it in the table.

But either way they should be swept away, so that is a separate problem I think.

dvic:
> @dvic are you sure the sweeper is running?

This is by default right? (I guess with recent Elixir versions, where, how I understand it, deps are automatically started?) Anyways, it seems to be running:

image


> Not recorded is a tricky situation. Originally a span that wasn't sampled wouldn't be inserted into the ets table, but the addition of "non-recording" complicated that. I think this was done to be safe but it may be the case that we don't need it.
> 
> I'll see if I can find anything in the spec that relates to a non-recording span being used that would mean we need to keep it in the table.
> 
> But either way they should be swept away, so that is a separate problem I think.

Clear, I'll have a look to see if I can replicate with a test.

dvic:
I can't replicate with a test (tried really quickly) but non-recording spans really seem to stay put in the ETS table. How is the sweeper supposed to work? Does it clean every 10 minutes the items that are 30 minutes and older? Is this the default setting?

tsloughter:
Yea, couldn't remember if it defaults to sweeping anything. But yea, looks like it has defaults of dropping any spans over 30 minutes old.

dvic:
> Yea, couldn't remember if it defaults to sweeping anything. But yea, looks like it has defaults of dropping any spans over 30 minutes old.

Alright, well if there is no good reason for having the non-recorded ones in ETS I suggest we skip inserting them, because for this particular type of operation (a cron job "sync" operation that involves many database operations) we now have 10x increase in memory usage :)

dvic:
I've just tested a patch that does not insert non-recording spans anymore and indeed it solves our memory problems (1100+ MB vs ~200 MB usage). Let me know if you want me to open a PR for this. By the way: all tests seem to pass, so if there is some behavior associated with these non-recording spans, it's not tested yet.

tsloughter:
Thanks, yes, I think a separate PR would be great. I'm guessing it should be merged but am still not positive, will try to figure out today.

Would be great to also figure out why the sweeper isn't working in this case. I can try to look into it today as well but if you have the opportunity to run a regular erlang trace on the sweeper to see what it is doing in your case where the table is growing that'd be a huge help. Hit me up on slack if you are able to do that and want some help.

dvic:
> Thanks, yes, I think a separate PR would be great. I'm guessing it should be merged but am still not positive, will try to figure out today.

Sure, I'll submit one then.

> Would be great to also figure out why the sweeper isn't working in this case. I can try to look into it today as well but if you have the opportunity to run a regular erlang trace on the sweeper to see what it is doing in your case where the table is growing that'd be a huge help. Hit me up on slack if you are able to do that and want some help.

The problem is not that the sweeper is not working but that the default config results in a high memory usage. We run a process that executes many queries and each query generates a span. This results in so many spans in ETS and given the fact that they are only cleaned every 10 minutes (older than 30 minutes) it takes 30 minutes or so to clear all non-recording spans. Here is an example log statement that was outputted:

```
14:11:05.842 module=otel_span_sweeper [info]  sweep old spans: ttl=1800000000000 num_dropped=5109
```

tsloughter:
@dvic aaah, ok. Yea, those defaults were picked to be very conservative.

apps/opentelemetry/test/opentelemetry_SUITE.erl
```diff
@@ -457,6 +458,38 @@ default_sampler(_Config) ->
...
+    SpanCtx2 = otel_tracer:start_span(Tracer, <<"span-2">>, #{}),
+    ?assertMatch(false, SpanCtx2#span_ctx.is_recording),
+    ?assertMatch(0, SpanCtx2#span_ctx.trace_flags),
+
+    ok.
+
+root_span_sampling_always_on(_Config) ->
+    Tracer = opentelemetry:get_tracer(),
+
+    Sampler = otel_sampler:setup(always_on),
```
tsloughter, line 480:
```suggestion
    Sampler = otel_sampler:new(always_on),
```

apps/opentelemetry/test/opentelemetry_SUITE.erl
```diff
@@ -457,6 +458,38 @@ default_sampler(_Config) ->
 
     ok.
 
+root_span_sampling_always_off(_Config) ->
+    Tracer = opentelemetry:get_tracer(),
+
+    Sampler = otel_sampler:setup(always_off),
```
tsloughter, line 464:
```suggestion
    Sampler = otel_sampler:new(always_off),
```

PR #264, opened 2021-08-06
Add test for child span sampling issue

dvic:


dvic:
The reason why it's returning `SampledFlag == true` and `IsRecording == false` is bevause `is_recording` is set to false when `end_span` is called, which is also mentioned in the spec. So this rule only applies to the result of the sampling procedure?

dvic:
Anyways, I found the bug, the parent based sampler pattern matched on `is_remote=false`, but it was `undefined`.

apps/opentelemetry/src/otel_sampler.erl
```diff
@@ -135,12 +135,11 @@ parent_based_sampler(#span_ctx{trace_flags=TraceFlags,
 parent_based_sampler(#span_ctx{is_remote=true}, #{remote_parent_not_sampled := SamplerAndOpts}) ->
     SamplerAndOpts;
 %% local parent sampled
-parent_based_sampler(#span_ctx{trace_flags=TraceFlags,
-                               is_remote=false}, #{local_parent_sampled := SamplerAndOpts})
+parent_based_sampler(#span_ctx{trace_flags=TraceFlags}, #{local_parent_sampled := SamplerAndOpts})
```
tsloughter, line 138: But now this is returning the `local_parent_sampled` sampler but not verifying that the span is not remote?

Is the issue instead that `is_remote` should be being set to false instead of `undefined`?

dvic: It was not because `is_remote=true` was pattern matched above. I have now changed the fix by adding `is_remote=false` in `root_span_ctx()`, not sure if this is the only place. The typespec by the way is still `boolean() | undefined`, not sure if you want to leave it like that or search also for other places and make it `boolean()`.

dvic: There is `get_ctx` in `otel_span_ets`, which is another potential candidate for setting `is_remote=false`.

tsloughter: I've got to check the spec this morning. This is likely a result of the multiple iterations done in the spec on how remote spans were created and stored. At one point they were stored in a separate context key from the current span context key and then I think that was switched to a remote field, but it has been so long I need to refamiliarize myself.

tsloughter: Oh wait, now its coming back to me. The is remote was not needed also when using what are called "nonrecording spans" for the extracted remote spans. But the spec does still include `is_remote`. I'm not 100% if `is_remote=true` can simply be set in `non_recording_span/3` or if it needs to be done in the propagator itself... But either way, one of those two are the fix.

dvic: Ah in that case the initial fix (not setting `is_remote=true` but implicitly checking for `false` or `undefined`) makes more sense to me. We can also change it to *explicitly* check for `false` and `undefined`? Let me know what you prefer.

tsloughter: Here is my PR that I think will fix it. I went with defaulting `is_remote` to `false` https://github.com/open-telemetry/opentelemetry-erlang/pull/265

dvic: LGTM! Do you want to keep the Elixir test in this PR or make a erlang test in #265 and can this PR be closed?

tsloughter:
The test is good to keep, just need the `is_remote` part removed from span utils in this PR.

Can you also rebase this down to 1 commit instead of 4?

dvic:
> The test is good to keep, just need the `is_remote` part removed from span utils in this PR.
> 
> Can you also rebase this down to 1 commit instead of 4?

Done :)

tsloughter (approved):

PR #267, opened 2021-08-11
Do not insert non-recording spans in ETS

dvic:
As per the discussion in https://github.com/open-telemetry/opentelemetry-erlang/pull/263#issuecomment-896845262, this PR ensures that non-recording spans are not inserted anymore in ETS.

### DIFF
--- a/apps/opentelemetry/src/otel_span_ets.erl
+++ b/apps/opentelemetry/src/otel_span_ets.erl
@@ -55,10 +55,8 @@ start_span(Ctx, Name, Opts, Processors, InstrumentationLibrary) ->
             Span2 = Processors(Ctx, Span1),
             _ = storage_insert(Span2),
             SpanCtx;
-        {SpanCtx, Span=#span{}} ->
-            %% span isn't recorded so don't run processors
-            %% but we do insert to ets table?
-            _ = storage_insert(Span),
+        {SpanCtx, #span{}} ->
+            %% span isn't recorded so don't run processors or insert into ets table
             SpanCtx
     end.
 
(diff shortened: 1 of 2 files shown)


tsloughter:
Thanks.

So yea, I still can't find a reason we'd need or want to store this span in ets.

Event `otel_span` functions ignore calls on spans that are not recording, so the table is never updated.

bryannaegele:
Probably wouldn't hurt to add a non-recording span case in https://github.com/open-telemetry/opentelemetry-erlang/blob/main/apps/opentelemetry/test/opentelemetry_SUITE.erl#L149

apps/opentelemetry/src/otel_span_ets.erl
```diff
@@ -55,10 +55,10 @@ start_span(Ctx, Name, Opts, Processors, InstrumentationLibrary) ->
             Span2 = Processors(Ctx, Span1),
             _ = storage_insert(Span2),
             SpanCtx;
-        {SpanCtx, Span=#span{}} ->
+        {SpanCtx, #span{}} ->
             %% span isn't recorded so don't run processors
             %% but we do insert to ets table?
-            _ = storage_insert(Span),
+            %% _ = storage_insert(Span),
```
tsloughter, line 61:
```suggestion
```

apps/opentelemetry/src/otel_span_ets.erl
```diff
@@ -55,10 +55,10 @@ start_span(Ctx, Name, Opts, Processors, InstrumentationLibrary) ->
             Span2 = Processors(Ctx, Span1),
             _ = storage_insert(Span2),
             SpanCtx;
-        {SpanCtx, Span=#span{}} ->
+        {SpanCtx, #span{}} ->
             %% span isn't recorded so don't run processors
             %% but we do insert to ets table?
```
tsloughter, line 60:
```suggestion
            %% span isn't recorded so don't run processors or insert into ets table
```

tsloughter (changes requested):

dvic:
@tsloughter I applied your suggestions and rebased into one commit.

dvic:
I've added a simple test case.

tsloughter (approved):

PR #270, opened 2021-09-01
Expose trace IDs in hex format

indrekj:
According to the [spec][1]
> The API MUST allow retrieving the TraceId and SpanId in the following forms:
>
>  *  Hex - returns the lowercase hex-encoded TraceId (result MUST be a 32-hex-character lowercase string) or SpanId (result MUST be a 16-hex-character lowercase string).
>  *  Binary - returns the binary representation of the TraceId (result MUST be a 16-byte array) or SpanId (result MUST be an 8-byte array).
>
> The API SHOULD NOT expose details about how they are internally stored.

This commit takes care of the Hex part.

[1]: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#retrieving-the-traceid-and-spanid)

### DIFF
--- a/apps/opentelemetry_api/src/otel_span.erl
+++ b/apps/opentelemetry_api/src/otel_span.erl
@@ -20,6 +20,8 @@
 
 -export([trace_id/1,
          span_id/1,
+         hex_trace_id/1,
+         hex_span_id/1,
          tracestate/1,
          is_recording/1,
          is_valid/1,
@@ -68,6 +70,14 @@ trace_id(#span_ctx{trace_id=TraceId }) ->
 -spec span_id(opentelemetry:span_ctx()) -> opentelemetry:span_id().
 span_id(#span_ctx{span_id=SpanId }) ->
     SpanId.
+
+-spec hex_trace_id(opentelemetry:span_ctx()) -> opentelemetry:hex_trace_id().
+hex_trace_id(#span_ctx{trace_id=TraceId}) ->
+    iolist_to_binary(io_lib:format("~32.16.0b", [TraceId])).
+
+-spec hex_span_id(opentelemetry:span_ctx()) -> opentelemetry:hex_span_id().
+hex_span_id(#span_ctx{span_id=SpanId}) ->
+    iolist_to_binary(io_lib:format("~16.16.0b", [SpanId])).
 
 -spec tracestate(opentelemetry:span_ctx() | undefined) -> opentelemetry:tracestate().
 tracestate(#span_ctx{tracestate=undefined}) ->
--- a/apps/opentelemetry_api/lib/open_telemetry/span.ex
+++ b/apps/opentelemetry_api/lib/open_telemetry/span.ex
@@ -40,6 +40,18 @@ defmodule OpenTelemetry.Span do
   defdelegate trace_id(span), to: :otel_span
 
   @doc """
+  Get the lowercase hex encoded span ID.
+  """
+  @spec hex_span_id(OpenTelemetry.span_ctx()) :: binary()
+  defdelegate hex_span_id(span), to: :otel_span
+
+  @doc """
+  Get the lowercase hex encoded trace ID.
+  """
+  @spec hex_trace_id(OpenTelemetry.span_ctx()) :: binary()
+  defdelegate hex_trace_id(span), to: :otel_span
+
+  @doc """
   Get the Tracestate of a Span.
   """
   @spec tracestate(OpenTelemetry.span_ctx()) :: OpenTelemetry.tracestate()
--- a/apps/opentelemetry_api/src/opentelemetry.erl
+++ b/apps/opentelemetry_api/src/opentelemetry.erl
@@ -60,6 +60,8 @@
 -export_type([tracer/0,
               trace_id/0,
               span_id/0,
+              hex_trace_id/0,
+              hex_span_id/0,
               trace_flags/0,
               timestamp/0,
               span_name/0,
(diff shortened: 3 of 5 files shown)


apps/opentelemetry_api/src/otel_span.erl
```diff
@@ -69,6 +71,14 @@ trace_id(#span_ctx{trace_id=TraceId }) ->
 span_id(#span_ctx{span_id=SpanId }) ->
     SpanId.
 
+-spec hex_trace_id(opentelemetry:span_ctx()) -> opentelemetry:hex_trace_id().
+hex_trace_id(#span_ctx{trace_id=TraceId }) ->
+    string:lowercase(integer_to_binary(TraceId, 16)).
```
tsloughter, line 76:
```suggestion
    io_lib:format("~32.16.0b", [TraceId]).
```

apps/opentelemetry_api/src/otel_span.erl
```diff
@@ -69,6 +71,14 @@ trace_id(#span_ctx{trace_id=TraceId }) ->
...
+
+-spec hex_span_id(opentelemetry:span_ctx()) -> opentelemetry:hex_span_id().
+hex_span_id(#span_ctx{span_id=SpanId }) ->
+    string:lowercase(integer_to_binary(SpanId, 16)).
```
tsloughter, line 80:
```suggestion
    io_lib:format("~16.16.0b", [SpanId]).
```

apps/opentelemetry_api/src/otel_span.erl
```diff
@@ -69,6 +71,14 @@ trace_id(#span_ctx{trace_id=TraceId }) ->
 span_id(#span_ctx{span_id=SpanId }) ->
     SpanId.
 
+-spec hex_trace_id(opentelemetry:span_ctx()) -> opentelemetry:hex_trace_id().
+hex_trace_id(#span_ctx{trace_id=TraceId }) ->
+    string:lowercase(integer_to_binary(TraceId, 16)).
+
+-spec hex_span_id(opentelemetry:span_ctx()) -> opentelemetry:hex_span_id().
+hex_span_id(#span_ctx{span_id=SpanId }) ->
```
tsloughter, line 79:
```suggestion
hex_span_id(#span_ctx{span_id=SpanId}) ->
```

apps/opentelemetry_api/src/otel_span.erl
```diff
@@ -69,6 +71,14 @@ trace_id(#span_ctx{trace_id=TraceId }) ->
 span_id(#span_ctx{span_id=SpanId }) ->
     SpanId.
 
+-spec hex_trace_id(opentelemetry:span_ctx()) -> opentelemetry:hex_trace_id().
+hex_trace_id(#span_ctx{trace_id=TraceId }) ->
```
tsloughter, line 75:
```suggestion
hex_trace_id(#span_ctx{trace_id=TraceId}) ->
```

tsloughter (changes requested):
Since the spec requires the trace id to be 32 characters and the span id to be 16 characters the `integer_to_binary` function can't be used -- or  at least can't be used without also adding padding to the beginning of the string. My suggested change is using `io_lib:format` to get a string of the correct length that is in hex.

tsloughter:
Thanks!

My initial thought was an api like `otel_span:trace_id(Span, Format)` might be nicer -- where `trace_id/1` would call `trace_id(Span, integer)`... but maybe not since we are unlikely to have to expose more than hex and binary formats.

apps/opentelemetry_api/src/otel_span.erl
indrekj: Good catch. I forgot that completely. https://github.com/open-telemetry/opentelemetry-erlang/compare/a19f1a44c1e63f969b87d34fe1e690994cb8767e..a97f5b7f3a034057a542bcde7be1a925ef169184

apps/opentelemetry_api/src/otel_span.erl
indrekj: https://github.com/open-telemetry/opentelemetry-erlang/compare/a19f1a44c1e63f969b87d34fe1e690994cb8767e..a97f5b7f3a034057a542bcde7be1a925ef169184

apps/opentelemetry_api/src/otel_span.erl
indrekj: https://github.com/open-telemetry/opentelemetry-erlang/compare/a19f1a44c1e63f969b87d34fe1e690994cb8767e..a97f5b7f3a034057a542bcde7be1a925ef169184

apps/opentelemetry_api/src/otel_span.erl
indrekj: https://github.com/open-telemetry/opentelemetry-erlang/compare/a19f1a44c1e63f969b87d34fe1e690994cb8767e..a97f5b7f3a034057a542bcde7be1a925ef169184

apps/opentelemetry_api/src/otel_span.erl
indrekj: Also added a test for that.

indrekj:
> but maybe not since we are unlikely to have to expose more than hex and binary formats

I don't have a strong opinion. I followed how it was done in [ruby](https://github.com/open-telemetry/opentelemetry-ruby/blob/cc28b8648e4085ff62e65e3db87ff45f82750cd1/api/lib/opentelemetry/trace/span_context.rb#L38-L50), but it seems that different language implementations do it differently (e.g. [python](https://github.com/open-telemetry/opentelemetry-python/blob/3fdfa6d6fbbb26e866ab8d40dc62ef1b4295ffc4/opentelemetry-api/src/opentelemetry/trace/span.py#L558-L577), [java](https://github.com/open-telemetry/opentelemetry-java/blob/d9ab761ea9a00381816a05d7feef08c61a06822d/api/all/src/main/java/io/opentelemetry/api/trace/SpanContext.java#L89-L95).

tsloughter (approved):

PR #272, opened 2021-09-03
Add b3multi option to OTEL_PROPAGATORS

indrekj:
According to [OTEL Spec][1] there are `b3` and `b3multi`. The latter is
supposed to be used when [Multi B3 format][2] is used. This is exactly
what b3_propagators() currently is doing.

[1]: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#general-sdk-configuration
[2]: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#b3-extract

### DIFF
--- a/apps/opentelemetry/src/otel_configuration.erl
+++ b/apps/opentelemetry/src/otel_configuration.erl
@@ -174,6 +174,8 @@ transform(propagators, PropagatorsString) when is_list(PropagatorsString) ->
 
 transform(propagator, "tracecontext") ->
     fun otel_tracer_default:w3c_propagators/0;
+transform(propagator, "b3multi") ->
+    fun otel_tracer_default:b3_propagators/0;
 transform(propagator, "baggage") ->
     fun otel_baggage:get_text_map_propagators/0;
 transform(propagator, Propagator) ->
--- a/apps/opentelemetry/test/otel_configuration_SUITE.erl
+++ b/apps/opentelemetry/test/otel_configuration_SUITE.erl
@@ -16,7 +16,8 @@
 all() ->
     [empty_os_environment, sampler, sampler_parent_based, sampler_parent_based_zero,
      sampler_trace_id, sampler_trace_id_default, sampler_parent_based_one,
-     log_level, propagators, otlp_exporter, jaeger_exporter, zipkin_exporter, none_exporter].
+     log_level, propagators, propagators_b3multi, otlp_exporter, jaeger_exporter,
+     zipkin_exporter, none_exporter].
 
 init_per_testcase(empty_os_environment, Config) ->
     Vars = [],
@@ -29,6 +30,12 @@ init_per_testcase(log_level, Config) ->
     [{os_vars, Vars} | Config];
 init_per_testcase(propagators, Config) ->
     Vars = [{"OTEL_PROPAGATORS", "baggage,afakeone"}],
+
+    setup_env(Vars),
+
+    [{os_vars, Vars} | Config];
+init_per_testcase(propagators_b3multi, Config) ->
+    Vars = [{"OTEL_PROPAGATORS", "b3multi"}],
 
     setup_env(Vars),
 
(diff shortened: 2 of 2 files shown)


tsloughter:
Thanks. I was afraid of this being the case :). 

I think this PR should break use of `b3` to ensure those using it must update to setting `multib3` and we can in the future include the single header `b3`.

indrekj:
> I think this PR should break use of b3 to ensure those using it must update to setting multib3 and we can in the future include the single header b3.

Currently, if you use `OTEL_PROPAGATORS="b3"` then there's a warning on application start:

> [warn] Ignoring uknown propagator b3 in OS environment variable $OTEL_PROPAGATORS

So basically it doesn't affect existing `b3` users as it didn't work before either. This PR just adds support to `OTEL_PROPAGATORS="b3multi"` without changing anything else.

Do you think that's fine or do you think something should be changed here?

tsloughter:
Ooh, my bad, I was taking into account my changes in the proapgators rewrite PR which has some extra lines about b3/multib3.

This should be fine.

indrekj:
@tsloughter Pushed https://github.com/open-telemetry/opentelemetry-erlang/compare/a5b7fccbca09f49b2f6143b69410b38929dbe5ad..b3fd84be88b052041f94ef6e93470b23fe57ddbd which should fix the test coverage check, but workflow has to be approved

indrekj:
:facepalm: missed to include the test: https://github.com/open-telemetry/opentelemetry-erlang/compare/b3fd84be88b052041f94ef6e93470b23fe57ddbd..42b17a65b0a6a7c0714be75727d9fc739a0d6535

indrekj:
It seems that this doesn't work exactly as I expected.

So the `propagators` are evaluated correctly. However, in `src/opentelemetry_app.erl` there's:
```
setup_text_map_propagators(Opts) ->
    Propagators = proplists:get_value(text_map_propagators, Opts, []),
    ...
```
Which uses `text_map_propagators`, instead of `propagators`, so setting the env variable doesn't seem to have any effect.

```
    Opts = otel_configuration:merge_with_os(
             application:get_all_env(opentelemetry)),
```
basically looks:
```
    %% [
    %%  {text_map_propagators,[fun otel_baggage:get_text_map_propagators/0,fun otel_tracer_default:w3c_propagators/0]},
    %%  {propagators,[fun otel_tracer_default:b3_propagators/0]},
    %%  {traces_exporter,{opentelemetry_exporter,#{}}}]
```

I'm not really what's the best way to fix this. Rename propagators to text_map_propagators?

tsloughter:
Oh, where is `propagators` used? It should always be `text_map_propagators` except in that `transform` function.

indrekj:
The issue is that it is not used. Basically, propagators sets it to b3 (when using env var) but then that is just ignored. So even though I use b3multi env variable, everything is still propagated using w3c trace context. Currently, it seems that OTEL_PROPAGATORS is basically a no-op. At least that's how I understand this.

indrekj:
@tsloughter I added a commit that fixes the issue I mentioned: https://github.com/open-telemetry/opentelemetry-erlang/pull/272/commits/6942f0864c661329b856b2937fe610511d9561aa . I have no idea if this is the actual approach you'd want to use but it should be backwards compatible and it does work.

tsloughter:
I think this is good.

tsloughter (approved):

indrekj:
@tsloughter I rebased against main branch but it says code cov reduced 0.02%. Would it be possible to access codecov to see what exactly it means? I'm getting Unauthorized when accessing https://app.codecov.io/gh/open-telemetry/opentelemetry-erlang/compare/272

tsloughter:
@indrekj the codecov is really screwed up and needs to be fixed, it is not accurate right now. So I would say don't worry about it, but if you want and have the time you are welcome to try fixing it ;).

It looks like I have to add you to the github team to give you access, that might have to go through someone else since its a cncf project.

indrekj:
@tsloughter do you think this could be merged in this case? Or is there anything else I need to do?

tsloughter:
Hey, yea, sorry about that, I thought I'd get the other propagators change in much earlier and thought merging it with the other PR would be easier than the other way around. I'll merge this now since that wasn't the case.

PR #273, opened 2021-09-07
Update propagators with composite

tsloughter:
This is #269 but with a composite propagator instead of the only way to use propagators being a list.

I wanted to try this in part to see if it would lead me to a better/simpler API. I don't think it did. I open to this version if people see it as clearer, and it does stick closer to the spec https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md

But I think I'd go with #269 -- curious to hear others thoughts?

### DIFF
--- a/apps/opentelemetry_api/src/opentelemetry.erl
+++ b/apps/opentelemetry_api/src/opentelemetry.erl
@@ -34,11 +34,9 @@
          register_application_tracer/1,
          get_tracer/0,
          get_tracer/1,
-         set_text_map_propagators/1,
-         set_text_map_extractors/1,
+         set_text_map_propagator/1,
          set_text_map_extractor/1,
          get_text_map_extractor/0,
-         set_text_map_injectors/1,
          set_text_map_injector/1,
          get_text_map_injector/0,
          timestamp/0,
@@ -157,30 +155,13 @@ get_tracer() ->
 get_tracer(Name) ->
     persistent_term:get({?MODULE, Name}, get_tracer()).
 
-%% setting the propagators is the same as setting the same list for
-%% injectors and extractors
-set_text_map_propagators(List) when is_list(List) ->
-    CompositeInjector = otel_propagator_text_map_composite:create(List),
-    CompositeExtractor = otel_propagator_text_map_composite:create(List),
-    set_text_map_injector(CompositeInjector),
-    set_text_map_extractor(CompositeExtractor);
-set_text_map_propagators(_) ->
-    ok.
-
-set_text_map_extractors(List) when is_list(List) ->
-    CompositeExtractor = otel_propagator_text_map_composite:create(List),
-    set_text_map_extractor(CompositeExtractor);
-set_text_map_extractors(_) ->
-    ok.
+%% setting the propagator is the same as setting the same injector and extractor
+set_text_map_propagator(Propagator) ->
+    set_text_map_injector(Propagator),
+    set_text_map_extractor(Propagator).
 
 set_text_map_extractor(Propagator) ->
     persistent_term:put({?MODULE, text_map_extractor}, Propagator).
-
-set_text_map_injectors(List) when is_list(List) ->
-    CompositeInjector = otel_propagator_text_map_composite:create(List),
-    set_text_map_extractor(CompositeInjector);
-set_text_map_injectors(_) ->
-    ok.
 
 set_text_map_injector(Propagator) ->
     persistent_term:put({?MODULE, text_map_injector}, Propagator).
--- a/apps/opentelemetry_api/src/opentelemetry.erl
+++ b/apps/opentelemetry_api/src/opentelemetry.erl
@@ -34,6 +34,7 @@
          register_application_tracer/1,
          get_tracer/0,
          get_tracer/1,
+         set_text_map_propagators/1,
          set_text_map_extractors/1,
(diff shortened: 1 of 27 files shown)


apps/opentelemetry_api/src/opentelemetry.erl
```diff
@@ -153,28 +155,36 @@ get_tracer(Name) ->
...
 %% injectors and extractors
 set_text_map_propagators(List) when is_list(List) ->
-    set_text_map_injectors(List),
-    set_text_map_extractors(List);
+    CompositeInjector = otel_propagator_text_map_composite:create(List),
+    CompositeExtractor = otel_propagator_text_map_composite:create(List),
+    set_text_map_injector(CompositeInjector),
+    set_text_map_extractor(CompositeExtractor);
 set_text_map_propagators(_) ->
     ok.
```
ferd, line 163: Would it make sense here to allow a single entry to work? I.e. people could pass in their own composite (or single) callback there, so long as they respect the sort of `{Mod::atom(), Args::list()}` format?

ferd:
as confusing as this is (as mentioned in side-channels, this feels really like OO ontology to please type checkers so it feels a bit foreign here), it sounds like the ability to be able to pass in arguments/state to the callback modules would be an interesting win there, considering the prior ones had no such support.

apps/opentelemetry_api/src/opentelemetry.erl
tsloughter: Ah yea, there was supposed to be a `set_text_map_propagator` like there is `set_text_map_injector/extractor`.

tsloughter: Tho maybe the plural forms should be removed. It is more important that simpler configuration works, which can do this without needing a public function. Would simplify the api, hopefully being less confusing.

tsloughter: Updated to only have the `set_text_map_propagator` function.

bryannaegele (commented):
This is great. Being able to inject/extract with arbitrary inputs is going to be helpful to have built into the main libs and there's still a nice balance with the default implementations.

Looks like there's a couple of tests to fix up, but otherwise 👍🏻

tsloughter:
@ferd yea, the ability to include options may be useful.

Right now there is no `init/setup/create` function for any propagator but the composite one, but options can be defined by configuring the propagator as a tuple.

I suppose a `create` could be added to the behaviour and all propagators either have the option to define it or be required to.

Which is something I don't like the in the current spec, it adds `create` to only the composite propagator and it isn't clear why. In an OOP language you'll have an `init` that can already be used to "create" a new composite propagator, and any other propagator, so why this special function only for it...

tsloughter:
@bryannaegele note the other PR also has all those things. But to be clear, you find the composite version to be better (clearer to the user)?

I think we'll probably go with this but there are still the open question of how/whether to have an init for all propagators.

Also the question of configuration. My idea is that we only support `text_map_propagators` which creates a composite and the only way to not use a composite is to use code to set the propagator.

Related to configuration. What happens if propagators, injectors and extractors are all listed in the configuration? Do they get appended together or one override the other...

tsloughter:
Ok, I think this is ready now.

Note that in configuration right now you can only set the propagators (text map propagators to be specific) and not injectors/extractors individually. So the functions in `opentelemetry` must be used if a user wants to use separate propagators for injecting and extracting.

apps/opentelemetry_api/src/otel_propagator.erl
```diff
@@ -12,55 +12,71 @@
...
 
-run_injectors(TextMap, Injectors) ->
-    lists:foldl(fun({Inject, {Key, ToText}}, TextMapAcc) ->
-                        Inject(TextMapAcc, Key, ToText);
-                   (_, TextMapAcc) ->
-                        TextMapAcc
-                end, TextMap, Injectors).
+%% @hidden
+-spec builtin_to_module(builtin() | module()) -> module().
+builtin_to_module(tracecontext) ->
```
bryannaegele, line 66: Are there two different versions of this atom? Line 69 has the same but with an underscore.

bryannaegele (commented):
> So the functions in opentelemetry must be used if a user wants to use separate propagators for injecting and extracting.

Just making sure I understand - you're referring to `otel_propagator_baggage:inject(Context, Carrier, CarrierSetFun, Options)` for instance, correct?

apps/opentelemetry_api/src/otel_propagator.erl
tsloughter: Yea. I added a comment in the module about that. Since the TraceContext module is `otel_propagator_trace_context` but the otel specs for like the environment variable use `tracecontext` I figured we should allow both when configuring.

tsloughter:
@bryannaegele no, users should not use propagator modules directly like that at all. In that comment I made I was referring just to globals, like:

```erlang
opentelemetry:set_text_map_injector({otel_propagator_b3multi, []})
```

Which makes the global text map injector be `b3multi` but doesn't touch what the extractors are.

To not use the global propagators a user should do:

```erlang
otel_propagator_text_map:inject({otel_propagator_baggage, []}, [{<<"existing-header">>, <<"I exist">>}]),
```

bryannaegele (approved):

PR #281, opened 2021-09-30
Add B3 single propagation format support

indrekj:
[B3 single format][1] looks like this: `b3={TraceId}-{SpanId}-{SamplingState}-{ParentSpanId}`,
where the last two fields are optional.

[OpenTelemetry B3 requirements][2]:

B3 has both single and multi-header encodings. It also has semantics
that do not map directly to OpenTelemetry such as a debug trace flag,
and allowing spans from both sides of request to share the same id. To
maximize compatibility between OpenTelemetry and Zipkin implementations,
the following guidelines have been established for B3 context
propagation.

When extracting B3, propagators:
* MUST attempt to extract B3 encoded using single and multi-header formats. The single-header variant takes precedence over the multi-header version. :green_circle: 
* MUST preserve a debug trace flag, if received, and propagate it with subsequent requests. Additionally, an OpenTelemetry implementation MUST set the sampled trace flag when the debug flag is set. :yellow_circle: We do not preserve the debug flag at the moment. We didn't do this before and not doing this in this PR either. We however do set sampling to 1 if that flag is present.
* MUST NOT reuse X-B3-SpanId as the id for the server-side span. :green_circle: 

When injecting B3, propagators:
* MUST default to injecting B3 using the single-header format :green_circle: 
* MUST provide configuration to change the default injection format to B3 multi-header :green_circle: 
* MUST NOT propagate X-B3-ParentSpanId as OpenTelemetry does not support reusing the same id for both sides of a request. :green_circle: 

[1]: https://github.com/openzipkin/b3-propagation#single-header
[2]: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#b3-requirements

### DIFF
--- a/apps/opentelemetry_api/src/otel_propagator_b3.erl
+++ b/apps/opentelemetry_api/src/otel_propagator_b3.erl
@@ -1,0 +1,111 @@
+%%%------------------------------------------------------------------------
+%% Copyright 2019, OpenTelemetry Authors
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
+%%
+%% http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing, software
+%% distributed under the License is distributed on an "AS IS" BASIS,
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+%%
+%% @doc An implementation of {@link otel_propagator_text_map} that injects and
+%% extracts trace context using the B3 specification from Zipkin.
+%%
+%% Since `trace_context' and `baggage' are the two default propagators the
+%% global TextMap Propagators must be configured if B3 is to be used for
+%% propagation:
+%%
+%% ```
+%% {text_map_propagators, [b3, baggage]},
+%% '''
+%%
+%% To use B3 multi-header format use:
+%%
+%% ```
+%% {text_map_propagators, [b3multi, baggage]},
+%% '''
+%%
+%% ```
+%% CompositePropagator = otel_propagator_text_map_composite:create([b3, baggage]),
+%% opentelemetry:set_text_map_propagator(CompositePropagator).
+%% '''
+%%
+%% It is also possible to set a separate list of injectors or extractors.
+%% For example, if the service should extract B3 encoded context but you
+%% only want to inject context encoded with the W3C TraceContext format
+%% (maybe you have some services only supporting B3 that are making requests
+%% to your server but you have no reason to continue propagating in both
+%% formats when communicating to other services further down the stack).
+%% In that case you would instead set configuration like:
+%%
+%%
+%% ```
+%% {text_map_extractors, [b3, trace_context, baggage]},
+%% {text_map_injectors, [trace_context, baggage]},
+%% '''
+%%
+%% Or using calls to {@link opentelemetry} at runtime:
+%%
+%% ```
+%% B3CompositePropagator = otel_propagator_text_map_composite:create([b3, trace_context, baggage]),
+%% CompositePropagator = otel_propagator_text_map_composite:create([trace_context, baggage]),
+%% opentelemetry:set_text_map_extractor(B3CompositePropagator),
+%% opentelemetry:set_text_map_injector(CompositePropagator).
+%% '''
+%% @end
+%%%-----------------------------------------------------------------------
+-module(otel_propagator_b3).
+
+-behaviour(otel_propagator_text_map).
+
+-export([fields/1,
+         inject/4,
+         extract/5]).
+
+-include("opentelemetry.hrl").
+
+-define(B3_CONTEXT_KEY, <<"b3">>).
+
+%% Returns all the keys the propagator sets with `inject'
+fields(b3single) ->
+    otel_propagator_b3single:fields(b3single);
+fields(b3multi) ->
+    otel_propagator_b3multi:fields(b3multi);
+fields(_) ->
+    [].
+
+-spec inject(Context, Carrier, CarrierSetFun, Options) -> Carrier
+              when Context :: otel_ctx:t(),
+                   Carrier :: otel_propagator:carrier(),
+                   CarrierSetFun :: otel_propagator_text_map:carrier_set(),
+                   Options :: b3multi | b3single.
+inject(Ctx, Carrier, CarrierSet, Options=b3single) ->
+    otel_propagator_b3single:inject(Ctx, Carrier, CarrierSet, Options);
+inject(Ctx, Carrier, CarrierSet, Options=b3multi) ->
+    otel_propagator_b3multi:inject(Ctx, Carrier, CarrierSet, Options);
+inject(_Ctx, Carrier, _CarrierSet, _Options) ->
+    Carrier.
+
+% Extract trace context from the supplied carrier. The b3 single header takes
+% precedence over the multi-header format.
+%
+% If extraction fails, the original context will be returned.
+-spec extract(Context, Carrier, CarrierKeysFun, CarrierGetFun, Options) -> Context
+              when Context :: otel_ctx:t(),
+                   Carrier :: otel_propagator:carrier(),
+                   CarrierKeysFun :: otel_propagator_text_map:carrier_keys(),
+                   CarrierGetFun :: otel_propagator_text_map:carrier_get(),
+                   Options :: otel_propagator_text_map:propagator_options().
+extract(Ctx, Carrier, CarrierKeysFun, CarrierGet, Options) ->
+    case otel_propagator_b3single:extract(Ctx, Carrier, CarrierKeysFun, CarrierGet, Options) of
+        Result when Result =/= undefined -> Result;
+        _ ->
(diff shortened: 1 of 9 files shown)


apps/opentelemetry_api/src/otel_propagator_b3multi.erl
```diff
@@ -91,58 +90,132 @@ inject(Ctx, Carrier, CarrierSet, _Options) ->
             Carrier
     end.
 
+% Extract trace context from the supplied carrier. The b3 single header takes
+% precedence over the multi-header format.
```
tsloughter, line 94: Hm, this is interesting. Does the otel spec also say this or only the b3 spec? I'm confused about how in otel a user could specify b3 or b3multi but its supposed to use both either way?

indrekj: That's in the otel spec: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#b3-requirements . First point in the Extract section. That's also how it's done in the ruby implementation: https://github.com/open-telemetry/opentelemetry-ruby/blob/main/propagator/b3/lib/opentelemetry/propagator/b3/text_map_extractor.rb#L48

apps/opentelemetry_api/src/otel_propagator_b3multi.erl
```diff
@@ -91,58 +90,132 @@ inject(Ctx, Carrier, CarrierSet, _Options) ->
...
+        Sampled = parse_is_sampled(CarrierGet(?B3_SAMPLED, Carrier)),
+
+        SpanCtx = otel_tracer:from_remote_span(TraceId, SpanId, Sampled),
+        otel_tracer:set_current_span(Ctx, SpanCtx)
+    catch
+        throw:invalid ->
+            undefined;
+
+        %% thrown if _to_integer fails
+        error:badarg ->
```
ferd, line 148: you might also get it if someone passes in an invalid string encoding for headers somehow. Then `string:length(...)` will fail with a badarg. That does end up being `undefined` all the same though.

apps/opentelemetry_api/src/otel_propagator_b3.erl
```diff
@@ -0,0 +1,96 @@
...
+%%
+%% Or using calls to {@link opentelemetry} at runtime:
+%%
+%% ```
+%% B3CompositePropagator = otel_propagator_text_map_composite:create([b3, trace_context, baggage]),
+%% CompositePropagator = otel_propagator_text_map_composite:create([trace_context, baggage]),
+%% opentelemetry:set_text_map_extractor(B3CompositePropagator),
+%% opentelemetry:set_text_map_injector(CompositePropagator).
+%% '''
+%% @end
```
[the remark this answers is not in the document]
indrekj: 99% copied from the existing b3multi module though

apps/opentelemetry_api/src/otel_propagator_b3multi.erl
tsloughter: Ah, the use of being able to specify one or the other is for injection, makes sense.

tsloughter:
A bit of a nitpick but I think this will be clearer if the module `otel_propagator_b3` is used to call extract of both modules `otel_propagator_b3multi` and a new module `otel_propagator_b3single`.

Injection can be based on a option to the propagator. So the configuration will then turn `b3` into `{otel_propagator_b3, b3single}` and `b3multi` into `{otel_propagator_b3, b3multi}`.

Does that make sense?

apps/opentelemetry_api/src/otel_propagator_b3.erl
```diff
@@ -0,0 +1,96 @@
...
+-spec inject(Context, Carrier, CarrierSetFun, Options) -> Carrier
+              when Context :: otel_ctx:t(),
+                   Carrier :: otel_propagator:carrier(),
+                   CarrierSetFun :: otel_propagator_text_map:carrier_set(),
+                   Options :: otel_propagator_text_map:propagator_options().
+inject(Ctx, Carrier, CarrierSet, _Options) ->
+    case otel_tracer:current_span_ctx(Ctx) of
+        #span_ctx{trace_id=TraceId,
+                  span_id=SpanId,
+                  trace_flags=TraceOptions} when TraceId =/= 0 andalso SpanId =/= 0 ->
```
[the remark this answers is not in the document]
tsloughter: Using `,` here will be consistent with the rest of the project's style, so probably worth accepting the change.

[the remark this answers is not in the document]
indrekj: That's how it was in b3multi as well, but changed both to be more consistent: https://github.com/open-telemetry/opentelemetry-erlang/compare/af27b2f7791fc1be02d6b4b11985f3aa723c0553..f780814913894dde6220cdcd4e27edefdc0900f3

apps/opentelemetry_api/src/otel_propagator_b3.erl
```diff
@@ -0,0 +1,96 @@
...
+            Options = case TraceOptions band 1 of 1 -> <<"1">>; _ -> <<"0">> end,
```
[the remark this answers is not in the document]
indrekj: Debug flag currently isn't propagated so here we deal only with `1` and `0`. Tbh I like the explicit version a bit more (I'm quite new to erlang so the code gold currently is a bit hard to grasp though I think I understand what is happening). I would keep it as is unless you feel strongly about changing it.

indrekj:
@tsloughter Yes, updated.

apps/opentelemetry_api/src/otel_propagator_b3.erl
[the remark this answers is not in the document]
tsloughter: Yea, the `case` version is good.

apps/opentelemetry_api/src/otel_propagator_b3multi.erl
tsloughter: Yea, just update the comment I guess.

apps/opentelemetry_api/src/otel_propagator_b3.erl
```diff
@@ -0,0 +1,110 @@
...
+                   Carrier :: otel_propagator:carrier(),
+                   CarrierSetFun :: otel_propagator_text_map:carrier_set(),
+                   Options :: otel_propagator_text_map:propagator_options().
+inject(Ctx, Carrier, CarrierSet, Options) ->
+    case Options of
+        b3single ->
+            otel_propagator_b3single:inject(Ctx, Carrier, CarrierSet, Options);
+        b3multi ->
+            otel_propagator_b3multi:inject(Ctx, Carrier, CarrierSet, Options)
+    end.
```
tsloughter, line 90:
```suggestion
inject(Ctx, Carrier, CarrierSet, Options=b3single) ->
    otel_propagator_b3single:inject(Ctx, Carrier, CarrierSet, Options);
inject(Ctx, Carrier, CarrierSet, Options=b3multi) ->
    otel_propagator_b3multi:inject(Ctx, Carrier, CarrierSet, Options);
inject(_Ctx, Carrier, _CarrierSet, _Options) -> 
    Carrier.
```

Since we want to never ever crash it is good to handle cases like this that technically shouldn't ever have a value outside of the 2 we allow but of course the user could also technically manually set a propagator that is `{otel_propagator_b3, gibberish}`, so best to be on the safe side and ignore it.

apps/opentelemetry_api/src/otel_propagator_b3.erl
```diff
@@ -0,0 +1,110 @@
...
+fields(b3single) ->
+    otel_propagator_b3single:fields(b3single);
+fields(b3multi) ->
+    otel_propagator_b3multi:fields(b3multi).
+
+-spec inject(Context, Carrier, CarrierSetFun, Options) -> Carrier
+              when Context :: otel_ctx:t(),
+                   Carrier :: otel_propagator:carrier(),
+                   CarrierSetFun :: otel_propagator_text_map:carrier_set(),
+                   Options :: otel_propagator_text_map:propagator_options().
```
tsloughter, line 83:
```suggestion
                   Options :: b3multi | b3single.
```

apps/opentelemetry_api/src/otel_propagator_b3multi.erl
indrekj: https://github.com/open-telemetry/opentelemetry-erlang/compare/6d963712d68353dca6ff18eb2d4d9adc7567e7dd..13b7dba5c47feaa15e78a549cdd27164631300c9

apps/opentelemetry_api/src/otel_propagator_b3.erl
indrekj: https://github.com/open-telemetry/opentelemetry-erlang/compare/c6bd10196b6af8c32e207d8d32e0d4340dfb10c0..6d963712d68353dca6ff18eb2d4d9adc7567e7dd

apps/opentelemetry_api/src/otel_propagator_b3.erl
indrekj: https://github.com/open-telemetry/opentelemetry-erlang/compare/c6bd10196b6af8c32e207d8d32e0d4340dfb10c0..6d963712d68353dca6ff18eb2d4d9adc7567e7dd

tsloughter (approved):

Issue #282, opened 2021-09-30
Ending a span should take an optional timestamp


fr0stbyte:
According to the spec : https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#end ending the span can take an optional timestamp. If none is provided, the current timestamp should be used to end the span

PR #283, opened 2021-09-30
first pass

fr0stbyte:
fixing issue #282

### DIFF
--- a/apps/opentelemetry/src/otel_span_utils.erl
+++ b/apps/opentelemetry/src/otel_span_utils.erl
@@ -19,7 +19,8 @@
 -module(otel_span_utils).
 
 -export([start_span/3,
-         end_span/1]).
+         end_span/1,
+         end_span/2]).
 
 -include_lib("opentelemetry_api/include/opentelemetry.hrl").
 -include("otel_sampler.hrl").
@@ -91,6 +92,12 @@ end_span(Span=#span{end_time=undefined,
 end_span(Span) ->
     Span.
 
+-spec end_span(opentelemetry:span(), integer() | undefined) -> opentelemetry:span().
+end_span(Span, Timestamp) when is_integer(Timestamp) ->
+    Span1 = Span#span{end_time=Timestamp},
+    end_span(Span1);
+end_span(Span, _) ->
+    end_span(Span).
 %%
 
 sample(Ctx, Sampler, TraceId, Links, SpanName, Kind, Attributes) ->
--- a/apps/opentelemetry_api/include/otel_tracer.hrl
+++ b/apps/opentelemetry_api/include/otel_tracer.hrl
@@ -27,6 +27,9 @@
 -define(end_span(SpanCtx),
         otel_span:end_span(SpanCtx)).
 
+-define(end_span(SpanCtx, Timestmap),
+        otel_span:end_span(SpanCtx, Timestamp)).
+
 -define(is_recording(),
         otel_span:is_recording(?current_span_ctx)).
 
--- a/apps/opentelemetry_api/include/otel_tracer.hrl
+++ b/apps/opentelemetry_api/include/otel_tracer.hrl
@@ -21,11 +21,8 @@
 
 %% updates the current context with the updated span context that
 %% has `is_recording' set to `false'
--define(end_span(),
-        otel_tracer:set_current_span(otel_span:end_span(?current_span_ctx))).
-
--define(end_span(SpanCtx),
-        otel_span:end_span(SpanCtx)).
+-define(end_span(Timestamp),
+        otel_tracer:set_current_span(otel_span:end_span(?current_span_ctx, Timestamp))).
 
 -define(end_span(SpanCtx, Timestmap),
         otel_span:end_span(SpanCtx, Timestamp)).
--- a/apps/opentelemetry_api/lib/open_telemetry/tracer.ex
+++ b/apps/opentelemetry_api/lib/open_telemetry/tracer.ex
@@ -126,19 +126,9 @@ defmodule OpenTelemetry.Tracer do
 
   The Span in the current Context has its `is_recording` set to `false`.
   """
-  def end_span() do
-    do_end_span(:opentelemetry:timestamp())
-  end
-
-  def end_span(timestamp) do
-    do_end_span(timestamp)
-  end
-
-
-  defp do_end_span(timestamp) do
+  def end_span(timestamp \\ :undefined) do
     non_recording_span = :otel_span.end_span(:otel_tracer.current_span_ctx(), timestamp)
-    _ = :otel_tracer.set_current_span(non_recording_span)
-    non_recording_span
+    :otel_tracer.set_current_span(non_recording_span)
   end
 
   @doc """
--- a/apps/opentelemetry_api/lib/open_telemetry/tracer.ex
+++ b/apps/opentelemetry_api/lib/open_telemetry/tracer.ex
@@ -127,7 +127,16 @@ defmodule OpenTelemetry.Tracer do
   The Span in the current Context has its `is_recording` set to `false`.
   """
   def end_span() do
-    non_recording_span = :otel_span.end_span(:otel_tracer.current_span_ctx())
+    do_end_span(:opentelemetry:timestamp())
+  end
+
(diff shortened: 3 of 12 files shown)


apps/opentelemetry/src/otel_span_utils.erl
```diff
@@ -83,12 +83,15 @@ root_span_ctx() ->
...
 %% @end
 %%--------------------------------------------------------------------
--spec end_span(opentelemetry:span()) -> opentelemetry:span().
+-spec end_span(opentelemetry:span(), integer() | undefined) -> opentelemetry:span().
 end_span(Span=#span{end_time=undefined,
-                    trace_flags=TraceFlags}) when ?IS_SAMPLED(TraceFlags) ->
-    EndTime = opentelemetry:timestamp(),
+                    trace_flags=TraceFlags}, Timestamp) when ?IS_SAMPLED(TraceFlags) ->
+    EndTime = case Timestamp of
+            x when is_integer(x) -> Timestamp,
```
tsloughter, line 90:
```suggestion
            X when is_integer(X) -> Timestamp,
```

tsloughter:
The `end_span/1` macro and functions need to stay. It looks like this currently would require users to pass a timestamp or `undefined` to `end_span`?

I know `nil` is a keyword in Elixir for `:nil` but I think since this isn't being used for anything but passing to the Erlang side (as in, there is no `nil` returned to a user, nor do they pass it themselves, they just don't pass a second argument) I think it'd be better to do `\\ :undefined`. Such a small thing but a surprisingly annoying quirk when doing interop.

apps/opentelemetry_api/include/otel_tracer.hrl
```diff
@@ -27,6 +27,9 @@
 -define(end_span(SpanCtx),
         otel_span:end_span(SpanCtx)).
 
+-define(end_span(SpanCtx, Timestmap),
```
tsloughter, line 30: You should add this also for the `end_span()` macro with no arguments above.

You'll want to rebase on `main` as well since these macros changed a bit in my latest PR that was just merged.

tsloughter:
Looking good. But another thing that is missing is adding `end_span(timestamp)` to `tracer.ex`. That is the Elixir function that will end the currently active span from the process dictionary.

apps/opentelemetry_api/include/otel_tracer.hrl
fr0stbyte: For the first point, I am not exactly sure what you mean.
If I change the macro with no arguments to
`-define(end_span(Timestamp),
       otel_tracer:set_current_span(otel_span:end_span(?current_span_ctx, Timestamp)).
`
there will be conflict w/ the other macro /1 

I could change it to `...otel_span:end_span(?current_span_ctx, opentelemetry:timestamp())` but wouldn't that be the same with `otel_span:end_span(ctx)` ?

tsloughter: Actually, that `?end_span(SpanCtx)` shouldn't be there. There are no other `SpanCtx` taking span macros calling `otel_span` functions in that header. So better to replace it with the timestamp version.

tsloughter: Note a lot of tests sadly use that macro and will need a change of `?end_span(SpanCtx)` -> `otel_span:end_span(SpanCtx)`.

fr0stbyte: updated

apps/opentelemetry_api/lib/open_telemetry/tracer.ex
```diff
@@ -127,7 +127,16 @@ defmodule OpenTelemetry.Tracer do
   The Span in the current Context has its `is_recording` set to `false`.
   """
   def end_span() do
-    non_recording_span = :otel_span.end_span(:otel_tracer.current_span_ctx())
+    do_end_span(:opentelemetry:timestamp())
```
tsloughter, line 130:
```suggestion
    do_end_span(:opentelemetry.timestamp())
```

But preferably only in the span utils module is the timestamp created unless passed to the user. I think the couple of lines that have to be duplicated by not sharing `do_end_span` between the two functions but instead each one calling `set_current_span` is fine.

Actually, `set_current_span` returns the span now, so you don't need to have the extra line returning `non_recording_span` anymore and can get rid of `_ = `

apps/opentelemetry_api/src/otel_span.erl
```diff
@@ -161,8 +162,21 @@ update_name(_, _) ->
...
-    _ = Module:end_span(SpanCtx),
+end_span(SpanCtx) when ?is_recording(SpanCtx) ->
+    end_span(SpanCtx, opentelemetry:timestamp()),
     SpanCtx#span_ctx{is_recording=false};
 end_span(SpanCtx) ->
     SpanCtx.
+
+-spec end_span(SpanCtx, Timestamp) -> SpanCtx when
+    SpanCtx :: opentelemetry:span_ctx(),
+    Timestamp :: integer() | undefined.
```
tsloughter, line 173:
```suggestion
    Timestamp :: integer().
```

I may be wrong here, and may have even suggested the `undefined` option, but I think it shouldn't be there when I look at the code now.

apps/opentelemetry_api/src/otel_span.erl
```diff
@@ -161,8 +162,21 @@ update_name(_, _) ->
...
+end_span(SpanCtx=#span_ctx{span_sdk={Module, _}}, Timestamp) when ?is_recording(SpanCtx) ->
+    EndTime = case Timestamp of
+        X when is_integer(X) -> X;
+        _ -> opentelemetry:timestamp()
+    end,
+    _ = Module:end_span(SpanCtx, EndTime),
+    SpanCtx#span_ctx{is_recording=false};
+end_span(SpanCtx, _) ->
+    SpanCtx.
```
tsloughter, line 182:
```suggestion
end_span(SpanCtx=#span_ctx{span_sdk={Module, _}}, Timestamp) when ?is_recording(SpanCtx)
                                                                  , is_integer(Timestamp) ->
    _ = Module:end_span(SpanCtx, EndTime),
    SpanCtx#span_ctx{is_recording=false};;
end_span(SpanCtx=#span_ctx{span_sdk={Module, _}}, Timestamp) when ?is_recording(SpanCtx) ->
    _ = Module:end_span(SpanCtx),
    SpanCtx#span_ctx{is_recording=false};
end_span(SpanCtx, _) ->
    SpanCtx.
```

tsloughter: Oh, I see, `end_span/1` was removed from `otel_span_ets`. I would add back `end_span/1` there, and have it pass `undefined` as the Timestamp to `end_span/3` which will take the span context, processors and end timestamp. Basically trying to make it so the only place a timestamp is got with `opentelemetry:timestamp()` is in that span utils module, and only place that `undefined` is an acceptable argument for timestamp is that final `end_span/3` function in `otel_span_ets`.

I noticed `otel_span_ets:end_span/2` is exported, but I don't know that it is used anywhere but internally to that module?

fr0stbyte: Imo, both end_span/2 and end_span/3 should not be exported

tsloughter: Yup, agreed.

apps/opentelemetry_api/src/otel_span.erl
```diff
@@ -161,8 +162,21 @@ update_name(_, _) ->
...
 end_span(SpanCtx) ->
     SpanCtx.
+
+-spec end_span(SpanCtx, Timestamp) -> SpanCtx when
+    SpanCtx :: opentelemetry:span_ctx(),
+    Timestamp :: integer() | undefined.
+end_span(SpanCtx=#span_ctx{span_sdk={Module, _}}, Timestamp) when ?is_recording(SpanCtx)
+                                                                  , is_integer(Timestamp) ->
+    _ = Module:end_span(SpanCtx, EndTime),
+    SpanCtx#span_ctx{is_recording=false};;
```
tsloughter, line 177:
```suggestion
    SpanCtx#span_ctx{is_recording=false};
```

PR #284, opened 2021-10-01
remove unneeded current span functions from otel_tracer

tsloughter:
`?end_span` is updated to work update the current context through
calling `otel_tracer:set_current_span`. In order to still return
the new `SpanCtx` this function was also changed to return the
`SpanCtx` that was passed to it.

### DIFF
--- a/apps/opentelemetry_api/include/otel_tracer.hrl
+++ b/apps/opentelemetry_api/include/otel_tracer.hrl
@@ -19,10 +19,10 @@
 -define(set_current_span(SpanCtx),
         otel_tracer:set_current_span(SpanCtx)).
 
-%% use tracer call to ensure the current span is overwritten with
-%% an ended (`is_recording' set to `false') is put in the context
+%% updates the current context with the updated span context that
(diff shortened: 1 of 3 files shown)


tsloughter:
I don't like that `tracer.ex` still has these functions because I don't know a better place to put functions that work on the current span. The Erlang API keeps all of those in the header macros so that there is a separation between the API working on the current process dictionary and the one working against explicit arguments.

Though I don't like that the Erlang header is `otel_tracer.hrl`, these aren't actually Tracer operations. They aren't really just Span operations either... Maybe there should be `otel.hrl`? Maybe it is better to just accept the naming not being perfect in order to not break more stuff?

ferd:
Eh, unless there's a clear path forwards and we know where we're going, any change is just another guess that makes an actual fix's friction a bit worse?

tsloughter:
Yea, you are right. The Elixir naming (having Span operations that work on the current span only available in `Tracer`) is weird enough we might want to see if we can figure out a clear path, but if not just leave it.

ferd (approved):

PR #287, opened 2021-10-06
separate application tracer registration from user named tracers

### LINKED ISSUE #279: Use single persistent term to map modules to tracers
Because inserting to persistent terms is proportional to the number of persistent terms we should be using a map as the lookup of the tracer stored in a persistent term, as @bryannaegele points out on slack.

So something like:

```erlang
-spec get_tracer(atom()) -> tracer().
get_tracer(Name) ->
    maps:get(Name, persistent_term:get({?MODULE, tracers}, #{}), get_tracer()).
```

And updating the setting of the mapping for module->tracer with code that builds up the whole map and then sets the persistent term.

tsloughter:
Fixes #279 

Because `persistent_term:put` slows down as the number of terms
grows a project with thousands of modules can cause startup to
timeout as it sets a term for every module in it. This patch
replaces that term for each module with a single term containing
a map of `module => tracer`.

Users must be able to register named tracers at runtime and we
want to not trigger a global GC by updating the tracer map
used for module to tracer mapping. So any user registered tracer
will have its own `persistent_term`.

The macro for looking up a tracer used in macros like `?with_span`
now looks up the tracer by looking up the module in the
"application tracers" (the persistent term containing the map).
While a user looking up a named tracer will do so with the function
`opentelemetry:get_tracer/1` which looks directly for a term with
that name.

This means a user could register a tracer with the same name as a
module. Whether this is good or bad or needs a check to guard
against is not covered by this PR.


apps/opentelemetry_api/src/opentelemetry.erl
```diff
@@ -29,16 +29,18 @@
 -module(opentelemetry).
 
 -export([set_default_tracer/1,
-         set_tracer/2,
          register_tracer/2,
-         register_application_tracer/1,
```
indrekj, line 34: I think this is still integrated in the Elixir module: https://github.com/open-telemetry/opentelemetry-erlang/blob/main/apps/opentelemetry_api/lib/open_telemetry.ex#L130

PR #266, opened 2021-08-09
User defined ssl options in grpc endpoints

pedro-gutierrez:
Allow to set user defined ssl options for https grpc endpoints. Eg calling `:otel_batch_processor.set_exporter(:opentelemetry_exporter, config)`  where `config` is:

```
 %{
  endpoints: [
    {:https, 'api.honeycomb.io', 443,
     [
       verify: :verify_peer,
       cacertfile::certifi.cacertfile(),
       customize_hostname_check: [
         match_fun: :public_key.pkix_verify_hostname_match_fun(:https)
       ]
     ]}
  ],
  headers: [
    {"x-honeycomb-team", "..."},
    {"x-honeycomb-dataset", "..."}
  ],
  protocol: :grpc
}
``` 

allows us to avoid ssl warnings in OTP 24 such as the one described in https://github.com/open-telemetry/opentelemetry-erlang/issues/259.

Fixes https://github.com/open-telemetry/opentelemetry-erlang/issues/259.

PR #292, opened 2021-10-13
improve exporter module documentation and readme

tsloughter:


apps/opentelemetry_exporter/src/opentelemetry_exporter.erl
```diff
@@ -69,15 +71,34 @@
 -define(DEFAULT_PORT, 4317).
 -define(DEFAULT_TRACES_PATH, "v1/traces").
 
--record(state, {protocol :: grpc | http_protobuf | http_json,
+-type headers() :: [{unicode:chardata(), unicode:chardata()}].
+-type scheme() :: http | https.
```
bryannaegele, line 75: Do these get parsed to atoms? Line 70 has scheme as a string.

apps/opentelemetry_exporter/README.md
```diff
@@ -26,18 +30,56 @@ The default protocol is `http_protobuf`, to override this and use grpc add `prot
...
+
+- `otlp_endpoint`: The URL to send traces and metrics to, for traces the path `v1/traces` is appended to the path in the URL.
+- `otlp_traces_endpoint`: URL to send only traces to. This takes precedence for exporting traces and the path of the URL is kept as is, no suffix is appended.
+- `otlp_headers`: List of additional headers (`[{unicode:chardata(), unicode:chardata()}]`) to add to export requests.
+- `otlp_traces_headers`: Additional headers (`[{unicode:chardata(), unicode:chardata()}]`) to add to only trace export requests.
+    
+    
+``` erlang
+{opentelemetry_exporter,
+  [{otlp_endpoint, "https://api.honeycomb.io:443"},
```
ferd, line 57: it might be good to add a note that SSL options aren't provided with this config. There's a new way to configure them and we could point out to https://erlef.github.io/security-wg/secure_coding_and_deployment_hardening/ssl for ways to safely configure it.

apps/opentelemetry_exporter/src/opentelemetry_exporter.erl
```diff
@@ -69,15 +71,34 @@
 -define(DEFAULT_PORT, 4317).
 -define(DEFAULT_TRACES_PATH, "v1/traces").
 
--record(state, {protocol :: grpc | http_protobuf | http_json,
+-type headers() :: [{unicode:chardata(), unicode:chardata()}].
+-type scheme() :: http | https | "http" | "https" | <<"http">> | <<"https">>.
```
ferd, line 75: there's a syntax error here, and I believe this is because the typespec doesn't allow specifying literal strings or binaries. You can only have `http | https | string() | binary()` and then rely on docs to restrict the valid strings or binaries you accept.

apps/opentelemetry_exporter/README.md
ferd: I thought https://github.com/open-telemetry/opentelemetry-erlang/pull/266 accomplished that already

tsloughter: Only for grpc. Looks like httpc never got any update.

PR #296, opened 2021-10-20
remove sampler option to start_span and with_span

tsloughter:
This was removed from the spec a loooooooooooooong time ago. Luckily we didn't add all those sampler options to libraries so hopefully this isn't an issue for many users? I have an OTEP almost ready for handling the use case of a specific sampler for a tracer.


test/otel_tests.exs
```diff
@@ -34,14 +37,16 @@ defmodule OtelTests do
...
     :otel_batch_processor.set_exporter(:otel_exporter_pid, self())
     OpenTelemetry.register_tracer(:test_tracer, "0.1.0")
+    t = {module, t_config} = :opentelemetry.get_tracer()
 
     sampler = :otel_sampler.new(:always_off)
+    t_always_off = {module, tracer(t_config, sampler: sampler)}
 
-    Tracer.with_span "span-1", %{sampler: sampler} do
-      Tracer.with_span "span-2" do
+    :otel_tracer.with_span(t_always_off, "span-1", %{}, fn _ ->
```
bryannaegele, line 45: I think we should continue to try to only use the elixir macro implementations in here and set the sampler in the setup, otherwise we're just retesting the erlang internals. I'm also afraid a user would look at this test as an example of setting up sampling in Elixir and determine this pattern is how you accomplish that.

ferd: What would be the canonical elixir way to do that anyway? If it is _the_ only way to do it, then I'd say it makes sense to have it in tests anyway.

tsloughter: @bryannaegele oh, this is not how a user should ever be doing anything. The tracer record is not public, they can not do this anyway. It is only done in the Erlang and Elixir tests in order to test different sampler behaviour.

PR #303, opened 2021-11-01
use get_application_tracer in Elixir macros

tsloughter:
The Erlang macros already properly use get_application_tracer.
The reason to use get_application_tracer is the macros lookup
the tracer for the particular application by module name.


tsloughter:
We should maybe just get rid of the separation between tracers registered automatically for each application and those registered by the user, and then there would be only one "get tracer" function available anyway

indrekj:
> We should maybe just get rid of the separation between tracers registered automatically for each application and those registered by the user, and then there would be only one "get tracer" function available anyway

That would simplify things. Currently the tracers & application tracers part is quite confusing, at least to me.

indrekj (approved):
looks right to me, but I'll try it out in real code tomorrow

tsloughter:
Agree it will probably confuse people.

Likely will merge this to get erlang and elixir to at least match asap and then have a PR for just getting rid of the separation.

ferd (approved):

PR #304, opened 2021-11-03
add simple span processor

tsloughter:
The Simple Span Processor does a blocking call when ending a span
that returns after the span is exported.

### DIFF
--- a/apps/opentelemetry/src/otel_simple_processor.erl
+++ b/apps/opentelemetry/src/otel_simple_processor.erl
@@ -1,5 +1,5 @@
 %%%------------------------------------------------------------------------
-%% Copyright 2019, OpenTelemetry Authors
+%% Copyright 2021, OpenTelemetry Authors
 %% Licensed under the Apache License, Version 2.0 (the "License");
 %% you may not use this file except in compliance with the License.
 %% You may obtain a copy of the License at
@@ -12,20 +12,15 @@
 %% See the License for the specific language governing permissions and
 %% limitations under the License.
 %%
-%% @doc The Batch Span Processor implements the `otel_span_processor'
-%% behaviour. It stores finished Spans in a ETS table buffer and exports
-%% them on an interval or when the table reaches a maximum size.
+%% @doc This Span Processor synchronously exports each ended Span.
 %%
-%% Timeouts:
-%%   exporting_timeout_ms: How long to let the exports run before killing.
-%%   check_table_size_ms: Timeout to check the size of the export table.
-%%   scheduled_delay_ms: How often to trigger running the exporters.
+%% Use this processor if ending a Span should block until it has been
+%% exported. This is useful for cases like a serverless environment where
+%% the application will possibly be suspended after handling a request.
 %%
-%% The size limit of the current table where finished spans are stored can
-%% be configured with the `max_queue_size' option.
 %% @end
 %%%-----------------------------------------------------------------------
--module(otel_batch_processor).
+-module(otel_simple_processor).
 
 -behaviour(gen_statem).
 -behaviour(otel_span_processor).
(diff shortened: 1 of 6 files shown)


tsloughter:
Opened as a draft since it still needs some docs and I don't like that both processors are started in the supervisor no matter what (but that may be acceptable for now).

apps/opentelemetry/src/otel_simple_processor.erl
```diff
@@ -0,0 +1,223 @@
...
+    handle_event_(idle, EventType, Event, Data).
+
+exporting({call, _From}, {export, _}, _) ->
+    {keep_state_and_data, [postpone]};
+exporting(internal, {export, From, Span}, Data=#data{exporting_timeout_ms=ExportingTimeout}) ->
+    {OldTableName, RunnerPid} = export_span(Span, Data),
+    {keep_state, Data#data{runner_pid=RunnerPid,
+                           current_from=From,
+                           handed_off_table=OldTableName},
+     [{state_timeout, ExportingTimeout, exporting_timeout}]};
```
ferd, line 113: This implies that the events have individual timeouts for their send time, which ignores the queuing time. Is this accurate? I'm unfamiliar with the spec here.

But let's assume this is correct, then the desired behaviour, if I have process A and processes B sending a synchronous span at once, and both hit a connection timeout of 2 seconds, would mean that A would wait ~2 seconds, and B would wait ~4 seconds before returning.

apps/opentelemetry/src/otel_simple_processor.erl
```diff
@@ -0,0 +1,223 @@
...
+
+%%
+
+complete_exporting(Data=#data{current_from=From,
+                              handed_off_table=ExportingTable})
+  when ExportingTable =/= undefined ->
+    {next_state, idle, Data#data{current_from=undefined,
+                                 runner_pid=undefined,
+                                 handed_off_table=undefined},
+     [{reply, From, ok}]}.
```
ferd, line 149: I guess here we can already have sent the timeout result. The gen_statem module however supports a way to use a proxy process to hide the dirty messages:

https://github.com/erlang/otp/blob/b3d4affcf9880255f6edc2e67095015e6ed2aca2/lib/stdlib/src/gen_statem.erl#L591-L596

Doing so would recommend using the option `{clean_timeout, infinity}` in `gen_statem:call/3`, which would set up a proxy process to ignore the duplicate messages. You might want to either use that, or make a special non-replying `complete_exporting/1` function acknowledge timeout races and prevents sending two replies to a single call.

ferd (changes requested):

apps/opentelemetry/src/otel_simple_processor.erl
tsloughter: It is unrelated to the queue time, correct. This is a configuration on the processor to kill the exporter if it takes too long. I punted on a client side timeout that can be passed for now -- I think its possible to do it decently now with `send_request` or whatever it is that ensures old replies aren't a problem.

apps/opentelemetry/src/otel_simple_processor.erl
tsloughter: Oh, thanks. I might use that, but this made me realize, don't I just need to set `current_from` to `undefined` in the timeout reply (which it currently isn't doing and is a bug)?

Then `complete_exporting` can not send a reply if the `current_from` is `undefined`.

ferd: Actually, not right now, unless you have a way to either cancel the state timeout, or to change the state. As this is implemented right now, we set the timer, complete, stay in the same state. 

So we're likely to hit an invalid leftover state timeout if there is a decent time between requests, such that sending one span, succeeding, responding okay, then waiting N seconds will still trigger a timeout message. The timeout is never cancel, the state never changes, so the state timeout should fire at some point. I'm not 100% what's the behaviour if you set a state timeout, then set a new one, whether that cancels the old state timeout or just adds more.

tsloughter: Wait no, it repeats the state and is supposed to retry, so it can't set `current_from` to `undefined`... grr

tsloughter: Yup, realized that after replying. 

Setting the state timeout does cancel and create a new one though.

tsloughter: Bleh, also, why am I sending an error if it repeats. Need to either not repeat or not return an error.

tsloughter: Updated to transition to idle.

`repeat` was used in batch processor to do another export of newly ended spans that have been waiting while the current exporting was blocking them from being exported.

I realize now that might be the wrong way to go about since there should be an export timeout message in the queue since we postpone it if in the exporting state, so there should be no need to repeat the state. But I'll want to verify my thinking again on that and send a separate PR.

ferd (approved):

PR #309, opened 2021-11-10
fix: Removes macro that references missing function

isaacsanders:


tsloughter (approved):

tsloughter:
Woops, thanks! :) Yea, links is only able to be set at start of a span but that wasn't always the case.

tsloughter:
@isaacsanders can you sign the CLA? I'll merge once you do.

PR #310, opened 2021-11-12
make trace/span id generation customizable

tsloughter:
A behaviour, otel_id_generator, is added which has callbacks for
generating trace and span ids, along with the default versions
of those functions.

The implementation to use is kept in the tracer record.

### DIFF
--- a/apps/opentelemetry/src/otel_span_ets.erl
+++ b/apps/opentelemetry/src/otel_span_ets.erl
@@ -25,7 +25,7 @@
          handle_call/3,
          handle_cast/2]).
 
--export([start_span/6,
+-export([start_span/7,
          end_span/1,
          end_span/2,
          get_ctx/1,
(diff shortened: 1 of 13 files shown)


tsloughter:
Crap, just realized this is wrong in the noop tracer.

tsloughter:
Fixed, based on my PR to the spec, will have to see how it goes https://github.com/open-telemetry/opentelemetry-specification/pull/2121

apps/opentelemetry/src/otel_span_ets.erl
```diff
@@ -46,10 +46,11 @@ start_link(Opts) ->
...
 %% @doc Start a span and insert into the active span ets table.
--spec start_span(otel_ctx:t(), opentelemetry:span_name(), otel_sampler:t(), otel_span:start_opts(),
-                 fun(), otel_tracer_server:instrumentation_library()) -> opentelemetry:span_ctx().
-start_span(Ctx, Name, Sampler, Opts, Processors, InstrumentationLibrary) ->
-    case otel_span_utils:start_span(Ctx, Name, Sampler, Opts) of
+-spec start_span(otel_ctx:t(), opentelemetry:span_name(), otel_sampler:t(), otel_id_generator:t(),
+                 otel_span:start_opts(), fun(), otel_tracer_server:instrumentation_library())
+                -> opentelemetry:span_ctx().
+start_span(Ctx, Name, Sampler, IdGeneratorModule, Opts, Processors, InstrumentationLibrary) ->
+    case otel_span_utils:start_span(Ctx, Name, Sampler, IdGeneratorModule, Opts) of
```
ferd, line 53: starting to think this whole injection mechanism coming from other langs means we'd eventually be better served by a default map that we merge with user options to override individual entries.

If everything in the universe can be overridden, the number of call params will never end growing.

ferd (approved):

apps/opentelemetry/src/otel_span_ets.erl
tsloughter: Eh, it is a little different. It isn't an override the user is passing, it is simply that more has been added to the Tracer and before there was no need to pass the actual Tracer down because there were just a couple pieces needed from it.

It may be that we should be passing the actual Tracer all the way down and not pulling out fields. I may make a PR to do that. I think pretty much everything in the Tracer is now an argument.

PR #314, opened 2021-11-19
exporter: update trace protos to v0.11.0

tsloughter:


tsloughter:
CI is failing because the committed module generated by gpb was generated on OTP-24 and won't compile on earlier versions. I opened an issue on gpb about a possible fix but guess I'll also just build OTP-23 locally and re-generate the module.

ferd (approved):

PR #316, opened 2021-11-24
tracers: store startup created tracers the same way as user created

tsloughter:
To simplify the understanding of tracer registration and fetching
this removes the separation between tracers registered on startup
for each application and those registered by a user or other code
at runtime.

Now there is a map to lookup the application name for a module
which is used to lookup the tracer, instead of before where the
map lookup actually got the tracer (meaning the tracer record was
duplicated and stored for every module!).

This means a user registering a tracer can override the tracer
registered at startup for an application name if they use the same
name, but I think this is sort of expected behavior so not an issue.

The change shouldn't effect much, if any, existing user code since
it almost always that tracers are fetched behind the scenes by
the macros.


tsloughter:
Last time this changes, I swear.

Originally it wasn't done this way because it requires a second lookup (first find the application name, then use that as the key for the persistent term where the tracer is stored). But I'd argue that those are fast enough and the user is able to store the returned tracer in a variable and use it if there are times they want to remove any lookup overhead in usage.

This extra lookup saves from having to store the tracer record duplicated for every single module and it removes the separation between tracers created by a user and those created at SDK startup. This important for when Tracers become configurable and it would be really confusing to deal with two sets of tracers when going to configure how one works.

tsloughter:
Note that this shouldn't effect any code since they should all be using the macros that hide it. And I think they all do now.

apps/opentelemetry_api/src/opentelemetry.erl
```diff
@@ -151,27 +152,23 @@ register_tracer(Name, Vsn) when is_atom(Name) and is_binary(Vsn) ->
...
-        TracerTuple = otel_tracer_provider:get_tracer(InstrumentationLibrary),
-        lists:foldl(fun(M, Acc) ->
-                            Acc#{M => TracerTuple}
-                  end, #{}, Modules)
-    catch exit:{noproc, _} ->
-            %% ignore because noproc here means no provider
-            %% likely since the SDK has been included and started
-            #{}
-    end.
+    persistent_term:put(?MODULE_TO_APPLICATION_KEY, TracerMap).
```
ferd, line 161: I guess an oddity of this approach (which was already present and doesn't need fixing here) is that two of these calls can run at once and could theoretically clobber each others, but that's such low probability I wouldn't care about it yet, since I figure registering for all apps is done once at startup and then it's individual calls.

ferd (approved):

apps/opentelemetry_api/src/opentelemetry.erl
tsloughter: Right, it is run by the Application module in the SDK only. Was thinking I'd want to make sure they were also hidden from docs and such. I had started with it going through the TracerProvider, like registering a tracer does, which would serialize through a `gen_server`, but for the above reasons I figure it isn't needed.

PR #325, opened 2021-11-29
move zipkin exporter to main repo

tsloughter:


apps/opentelemetry_zipkin/README.md
```diff
@@ -0,0 +1,28 @@
+# OpenTelemetry Zipkin Exporter
+
+![](https://github.com/opentelemetry-beam/opentelemetry_zipkin/workflows/OpenTelemetry%20Zipkin%20Tests/badge.svg)
```
ferd, line 3: I figure this is where we're moving it from; the badge is unlikely to remain functional.

apps/opentelemetry_zipkin/src/opentelemetry_zipkin.app.src
```diff
@@ -0,0 +1,16 @@
...
+   [kernel,
+    stdlib,
+    inets,
+    opentelemetry_api
+   ]},
+  {env,[]},
+  {modules, []},
+
+  {licenses, ["Apache-2.0"]},
+  {links, [{"GitHub", "https://github.com/opentelemetry-beam/opentelemetry_zipkin"}]}
```
ferd, line 15: This will no longer be true

apps/opentelemetry_zipkin/test/opentelemetry_zipkin_SUITE.erl
```diff
@@ -0,0 +1,67 @@
...
+                      events = [#event{system_time_nano=erlang:system_time(nanosecond),
+                                       name = <<"event-1">>,
+                                       attributes = [{<<"attr-1">>, <<"value-1">>}]},
+                                #event{system_time_nano=erlang:system_time(nanosecond),
+                                       name = <<"event-2">>,
+                                       attributes = [{<<"attr-3">>, <<"value-3">>}]}],
+                      attributes = [{<<"attr-2">>, <<"value-2">>}]},
+    true = ets:insert(Tid, ChildSpan),
+
+    ?assertMatch(ok, opentelemetry_zipkin:export(Tid, Resource, State)),
```
ferd, line 66: this line's failing in the tests, returning `error` only. This seems to be due to the http call itself failing (or returning an unknown status), but we can't see which of the two potential log lines (error response or client error) got triggered.

ferd (changes requested):

apps/opentelemetry_zipkin/test/opentelemetry_zipkin_SUITE.erl
tsloughter: hm, damn, ok. It is passing locally and both local and github actions use the same docker compose file. But it must be something to do with the zipkin services it is trying to export to.

tsloughter:
@ferd this should be good to go now.

ferd (approved):

Issue #324, opened 2021-11-29
Dropped items tracking and exporting


bryannaegele:
Looks like we have to keep counts of dropped attributes, events, and links.

> Counts for attributes, events and links dropped due to collection limits MUST be available for exporters to report as described in the exporters specification.

https://github.com/open-telemetry/opentelemetry-specification/blob/e2c2472985b17e37a25a7dc5aa0aa071e6683c98/specification/trace/sdk.md#additional-span-interfaces

*[Dropped Attributes Count](https://github.com/open-telemetry/opentelemetry-specification/blob/e2c2472985b17e37a25a7dc5aa0aa071e6683c98/specification/trace/sdk_exporters/non-otlp.md#dropped-attributes-count)
* [Dropped Events Count](https://github.com/open-telemetry/opentelemetry-specification/blob/e2c2472985b17e37a25a7dc5aa0aa071e6683c98/specification/trace/sdk_exporters/non-otlp.md#dropped-events-count)
* [Dropped Links Count](https://github.com/open-telemetry/opentelemetry-specification/blob/e2c2472985b17e37a25a7dc5aa0aa071e6683c98/specification/trace/sdk_exporters/non-otlp.md#dropped-links-count)

Issue #323, opened 2021-11-29
Limit attribute value types


bryannaegele:
Limit accepted values.

https://github.com/open-telemetry/opentelemetry-specification/blob/e2c2472985b17e37a25a7dc5aa0aa071e6683c98/specification/common/common.md#attributes

Issue #321, opened 2021-11-28
Link count limit - attributes


bryannaegele:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#span-limits

Limit the number of attributes on links. This value should be configurable.

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md#attribute-limits-configuration

Issue #319, opened 2021-11-28
Span count limits - events


bryannaegele:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#span-limits

Limit the number of events on spans. This value should be configurable.

PR #329, opened 2021-12-16
Support limits on attributes, links and events

### LINKED ISSUE #317: Span count limit - links
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#span-limits

Limit the number of links on spans. This value should be configurable.

tsloughter:
Closes #317 #318 #319 #320 #321 #323 #324

Sadly a lot here. Opening as a draft because there is still a bug in attributes update I need to fix (it is marked with a `TODO` right now). And I want to do a review myself, I'm opening this once tests and dialyzer passed just to get it open tonight.

I also want to revisit the API in `opentelemetry` module that creates stuff like `events` and `links` and whether those should be removed. Right now the code in `otel_links` and `otel_events` has to deal with the passed in links and events being multiple different formats.

Just realized I didn't run the Elixir tests and there are failures there because they compare stuff like the resulting attributes from an export so I'll have to fix that up in the morning.


tsloughter:
Fixed the issue with adding attributes where the key already exists but the size limit has been reached. Not really efficient... but it works for now.

tsloughter:
Open question that may lead to another PR is if we should get rid of stuff like `opentelemetry:link`,  `opentelemetry:links`, ``opentelemetry:event`,  `opentelemetry:events` and their Elixir equivalents `OpenTelemetry.links`, etc...

Since they don't know about the limits on attributes those records returned by the functions aren't "finished", when adding them to the span the attributes have to be parsed again.

I'd lean towards removing them but they are likely already in use by some users.. a little.

tsloughter:
Think I got a partial middle ground. I changed the link[s] and event[s] functions to return maps representing each instead of records and moved the records to the SDK where they are better fit.

It should probably still be simplified but annoyingly don't know what ppl are already relying on and hard to get feedback.

apps/opentelemetry/src/otel_events.erl
```diff
@@ -0,0 +1,103 @@
...
+    List.
+
+dropped(#events{dropped=Dropped}) ->
+    Dropped.
+
+%%
+
+add_events(NewEvents, Events=#events{count_limit=CountLimit,
+                                     list=List}) when length(List) < CountLimit ->
+    Limit = CountLimit - length(List),
```
ferd, line 59: this calculates the length of the list twice per event addition. At low counts that's probably cheap, but this may start becoming a significant CPU hog for people with custom limits set higher. It would probably be more efficient to count the limit as an attribute integer (like with `dropped` and `count_limit`) and then pattern match on the values.

apps/opentelemetry/src/otel_links.erl
```diff
@@ -0,0 +1,103 @@
...
+create_links([], _, _, _AttributeValueLengthLimit, Dropped, Links) ->
+    {Links, Dropped};
+create_links(List, 0, _, _, Dropped, Links) ->
+    {Links, Dropped + length(List)};
+create_links([L | Rest], Limit, AttributePerLinkLimit, AttributeValueLengthLimit, Dropped, List) ->
+    case new_link(L, AttributePerLinkLimit, AttributeValueLengthLimit) of
+        drop ->
+            create_links(Rest, Limit, AttributePerLinkLimit, AttributeValueLengthLimit, Dropped + 1, List);
+        Link ->
+            create_links(Rest, Limit - 1, AttributePerLinkLimit, AttributeValueLengthLimit, Dropped, [Link | List])
```
ferd, line 66: hah, the limit here is more nicely held as a single counter, good.

apps/opentelemetry/test/opentelemetry_SUITE.erl
```diff
@@ -614,22 +636,86 @@ record_exception_with_message_works(Config) ->
     SpanCtx = ?start_span(<<"span-1">>),
     try throw(my_error) of
         _ ->
-        ok
+            ok
```
ferd, line 639: you can actually use `try throw(my_error) catch ...` and skip the `of Pat -> Clause` section since this is guaranteed to fail.

apps/opentelemetry/test/opentelemetry_SUITE.erl
```diff
@@ -614,22 +636,86 @@ record_exception_with_message_works(Config) ->
...
+
+    %% dropped because of tuple as value
+    ?set_attribute(<<"attr-2">>, {not_allowed, in, attributes}),
+
+    ?set_attribute(<<"attr-3">>, <<"attr-value-3">>),
+
+    %% dropped because count limit was set to 2
+    ?set_attribute(<<"attr-4">>, <<"attr-value-4">>),
+    %% won't be dropped because attr-1 already exists so it overrides the value
+    ?set_attribute(<<"attr-1">>, <<"attr-value-5">>),
```
ferd, line 689: oh interesting semantics there where the oldest value wins as long as the keys are new. Cool gotcha on the limits' behaviour.

apps/opentelemetry/src/otel_resource.erl
```diff
@@ -36,24 +36,31 @@
...
-merge({otel_resource, CurrentResource}, {otel_resource, OtherResource}) ->
-    {otel_resource, lists:ukeymerge(1, CurrentResource, OtherResource)}.
+merge({otel_resource, CurrentAttributes}, {otel_resource, UpdatingAttributes}) ->
+    CurrentMap = otel_attributes:map(CurrentAttributes),
+    UpdatingMap = otel_attributes:map(UpdatingAttributes),
+    %% TODO: fix, the dropped field for the new attributes will be wrong
+    %% we can't just use `otel_attributes:set' since that will drop instead
+    %% of override a key/value pair when the limit is reached even if the
+    %% new attribute matches a key in the current attributes
+    {otel_resource, otel_attributes:new(maps:merge(CurrentMap, UpdatingMap), 128, 255)}.
```
ferd, line 63: A gotcha here is that your merging won't necessarily respect the insertion order, just the iteration order. So on this operation the validation may become really tricky in terms of what's left over, and change from time to time even if the original maps compare equal. The returned values of `maps:from_list` and iterators isn't guaranteed as far as I know.

ferd (approved):
I'm approving considering this is to get us to a release candidate, but I have some concerns about performance of some operations, and highlighted a spot where the merging of maps may be counter-intuitive in its results (and maybe a sort() could be used to be deterministic)

apps/opentelemetry/src/otel_events.erl
tsloughter: Yea, I was debating keeping it in the record since it would be a lot of work for a user to go around the add functions to update the list, so it shouldn't be an issue that it gets out of sync. But figured it was an optimization to do later.

But one thing I can do quick is there is no need to check in the guard, it can just call `add_events_` and return immediately since `Limit` will be 0.

ferd: yeah that's probably a good workaround to save one call at least, +1 on that

apps/opentelemetry/src/otel_links.erl
tsloughter: It is simpler in links since you can only add links when calling start span, not add them at any time like with events.

apps/opentelemetry/test/opentelemetry_SUITE.erl
tsloughter: Hm, yea, thanks, don't know where that came from.

apps/opentelemetry/test/opentelemetry_SUITE.erl
tsloughter: Yup, something I got wrong the first time (fixed before opening the PR) :)

> for each unique attribute key, addition of which would result in exceeding the limit, SDK MUST discard that key/value pair.

apps/opentelemetry/src/otel_resource.erl
tsloughter: This is actually wrong. This was done before I learned about the mistake I made reading the spec for attributes. This can rely on the regular attributes module because that does not drop the new attributes like I originally had it doing.

PR #331, opened 2021-12-22
Support for Schema URLs on the Resource and Tracer (Instrumentation Library)

tsloughter:
Related parts of the spec:

- https://github.com/open-telemetry/opentelemetry-specification/blob/v1.8.0/specification/resource/sdk.md
- https://github.com/open-telemetry/opentelemetry-specification/blob/v1.8.0/specification/trace/api.md#get-a-tracer

The work on Resource led to converting Resource attributes to binary strings because it was too much of a pain to keep them as list strings when everything else is binary. I think the original reasoning was just that they were only latin1 strings that were small, so a charlist would be more efficient memory wise.

There is also a commit with a fix for parsing the deny list and resource detectors configuration which I can break out to another PR if anyone thinks it needs separate discussion.

### DIFF
--- a/apps/opentelemetry/src/otel_resource_app_env.erl
+++ b/apps/opentelemetry/src/otel_resource_app_env.erl
@@ -50,22 +50,27 @@ parse(_) ->
 parse_values(Key, Values) when is_map(Values) ->
     parse_values(Key, maps:to_list(Values));
 parse_values(Key, Values) when is_list(Values) ->
-    lists:flatmap(fun({SubKey, Value=[{_,_}|_]}) ->
-                          %% list of tuples means we have more subkeys
-                          parse_values([Key, ".", to_string(SubKey)], Value);
-                     ({SubKey, Value}) when is_map(Value) ->
-                          %% map value means we have more subkeys
-                          parse_values([Key, ".", to_string(SubKey)], Value);
-                     ({SubKey, Value})->
-                          [{unicode:characters_to_list([Key, ".", to_string(SubKey)]), Value}]
-                  end, Values);
+    case io_lib:printable_latin1_list(Values) of
+        true ->
+            [{unicode:characters_to_binary(Key), unicode:characters_to_binary(Values)}];
+        false ->
+            lists:flatmap(fun({SubKey, Value=[{_,_}|_]}) ->
+                                  %% list of tuples means we have more subkeys
+                                  parse_values([Key, ".", to_string(SubKey)], Value);
+                             ({SubKey, Value}) when is_map(Value) ->
+                                  %% map value means we have more subkeys
+                                  parse_values([Key, ".", to_string(SubKey)], Value);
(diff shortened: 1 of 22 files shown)


tsloughter:
I don't like the `update_instrumentation_library` call. Not sure what to do yet. And actually, it should probably be getting the default tracer and settings its instrumentation library field... but still ugly.

Hm, I was planning to do a PR after this one to remove `register_tracer` and move to just `get_tracer` like all the other languages do, that might simplify this and make it cleaner, so I might add it on here, unless I get a +1 soon and will then just do a separate PR that cleans that up.

apps/opentelemetry/src/otel_resource_app_env.erl
```diff
@@ -50,22 +50,27 @@ parse(_) ->
...
-    lists:flatmap(fun({SubKey, Value=[{_,_}|_]}) ->
-                          %% list of tuples means we have more subkeys
-                          parse_values([Key, ".", to_string(SubKey)], Value);
-                     ({SubKey, Value}) when is_map(Value) ->
-                          %% map value means we have more subkeys
-                          parse_values([Key, ".", to_string(SubKey)], Value);
-                     ({SubKey, Value})->
-                          [{unicode:characters_to_list([Key, ".", to_string(SubKey)]), Value}]
-                  end, Values);
+    case io_lib:printable_latin1_list(Values) of
```
ferd, line 53: If using the `+pc unicode` flags, the function to call is different.

The better API is to first call `io:printable_range() -> latin1 | unicode` and then call `io_lib:printable_latin1_list(S)` or `io_lib:printable_unicode_list(S)` according to the returned value.

ferd (approved):

apps/opentelemetry/src/otel_resource_app_env.erl
tsloughter: Ah, ok. But isn't utf8 a superset of latin1?

ferd: yeah. So there's a lot more printable stuff. I just don't know if it means you want to go recursive every time there's a unicode entity (say, an emoji or japanese characters) rather than just displaying the value.

tsloughter: oh, I see the confusion. And actually, I can no longer find in the spec anything about resource attributes being only ascii, maybe it was never merged, maybe it was changed. Either way, I'm changing it to be unicode.

tsloughter: Also just realized, should I always be calling `io_lib:printable_unicode_list` where I want to check strings? Even if the user has `+pc latin1` it will work if unicode is found?

tsloughter: Updated to use `printable_unicode_list`.

ferd:
looks good.

PR #332, opened 2021-12-25
Remaining validations

### LINKED ISSUE #323: Limit attribute value types

bryannaegele:
This adds validations and limits attribute values to those in the spec rather than what the protos are capable of exporting. I've also added validation to span start opts which were previously not being validated in any way.

There's a second commit here that removes validations and tracking attribute validation drops at the SDK level. Looking through the spec, [this section](https://github.com/open-telemetry/opentelemetry-specification/blob/9216eb8ff1832ae790369e49ce326cb6bc3d1c80/specification/trace/sdk.md?plain=1#L127-L129) indicates the dropped attribute count is tracking attributes dropped due to count limits. This makes sense to me since validation and allowed types are an API-level concern which aren't trackable, whereas system limits are an SDK-level concern which is trackable.

We can try to get some verification on that one.

Closes #323


bryannaegele:
Crap. Just realized `filtermap` is an OTP 24 introduction :/

tsloughter:
@bryannaegele filtermap is from R16

apps/opentelemetry/src/otel_attributes.erl
```diff
@@ -71,35 +71,27 @@ update_attributes(List, Attributes) ->
...
         true ->
-            {true, string:slice(Value, 0, ValueLengthLimit)};
+            Attributes#attributes{map=Map#{Key => string:slice(Value, 0, ValueLengthLimit)}};
         false ->
-            {true, Value}
+            Attributes#attributes{map=Map#{Key => Value}}
     end;
-is_allowed(_, _) ->
-    false.
+%% already in the map and not a binary so update
```
tsloughter, line 82: Why adding anything that isn't a binary?

bryannaegele: The API will have already filtered out everything in the validation step before getting to the SDK, so it should only need to possibly truncate a binary value if it exceeds the length limit.

tsloughter: Ah, gotcha.

One thing I forgot to do and I don't think I see here is also limiting the length of an list (array). Think you can sneak that in here? If no I'll open a separate PR after this gets merged.

bryannaegele:
> @bryannaegele filtermap is from R16

Yeah, was referring to the `maps:filtermap` which was more recent. I didn't need to be using that anyhow since there was no transform.

One questionable thing I did was returning `undefined` on invalid names. I didn't want to raise and the only other alternative would be to have some placeholder, but in a `with_span` then that `undefined` would get passed to `end_span` which presently doesn't handle that; it always expects a span ctx. We can start handling `undefined` on `end_span` or raise on the name. Thus far, we've generally stayed quiet on errors but maybe this calls for it.

apps/opentelemetry/src/otel_attributes.erl
bryannaegele: Yeah, I thought there was something for that in the spec but I couldn't find it. Is it a separate value from the binary limit?

tsloughter: Na, same limit. I'll go look for it in the spec, I'm pretty sure its there somewhere.

tsloughter: Oh, well first I guess, is each string in an array has to be truncated:

> if it is an array of strings, then apply the above rule to each of the values separately,

bryannaegele: Yeah but are we going to enforce that? I thought we decided to not validate or mess with list contents due to the perf hit. I guess since it's a resources protection issue, maybe we have to do it anyhow. At that point, I may as well go ahead and add the array homogeneity checks, too.

Up to you. If it's a real drag on performance we could add an opt-out option later, I suppose. I'd be a little surprised if folks had many lists or that they'd be lengthy enough to be truly impactful.

bryannaegele: Iterating a list of binaries and checking their length is probably the worst possible perf iteration we could do, so makes sense to just add everything if we're already making that possible.

tsloughter: It is in the spec as a requirement if we support limits, so yea, we should have it.

We can document to a user that if they want to avoid these iterations they can set the length limit to infinity. We could check homogeneity only when it isn't infinity for length. Possibly a bit odd behavior to some but simple to document. And the issue of wanting to drop before the exporter (where it'll be dropped if it isn't homogeneous) is only an issue if they add too many attributes that they hit the limit and then want latter attributes to still be included since those should have been dropped... so I don't think its a big deal.

tsloughter:
Oh, missed the part about raising. Na, should never raise. Should create a zero span where all id's are 0.

tsloughter:
And `is_valid` `is_recording` are false:

```
-define(NOOP_SPAN_CTX, #span_ctx{trace_id=0,
                                 span_id=0,
                                 trace_flags=0,
                                 tracestate=[],
                                 is_valid=false,
                                 is_recording=false,
                                 span_sdk=undefined}).
```

apps/opentelemetry/src/otel_attributes.erl
```diff
@@ -71,35 +71,38 @@ update_attributes(List, Attributes) ->
...
+    Attributes#attributes{map=Map#{Key => maybe_truncate_binary(Value, ValueLengthLimit)}};
+%% value is a list of binaries, so potentially truncate
+update_attribute(Key, [Value1 | _Rest] = Value, Attributes=#attributes{count_limit=CountLimit,
+                                                    value_length_limit=ValueLengthLimit,
+                                                    map=Map})
+  when is_binary(Value1) , (map_size(Map) < CountLimit orelse is_map_key(Key, Map)) ->
+    Attributes#attributes{map=Map#{Key => [maybe_truncate_binary(V, ValueLengthLimit) || V <- Value]}};
+%% already in the map and not a binary so update
+update_attribute(Key, Value, Attributes=#attributes{map=Map}) when is_map_key(Key, Map) ->
+    Attributes#attributes{map=Map#{Key => Value}};
```
tsloughter, line 85:
```suggestion
    Attributes#attributes{map=Map#{Key := Value}};
```

This might be more efficient since we know the key is there.

apps/opentelemetry/test/opentelemetry_SUITE.erl
```diff
@@ -711,21 +730,26 @@ too_many_attributes(Config) ->
 
     ?set_attribute(<<"attr-1">>, <<"attr-value-1">>),
 
-    %% dropped because of tuple as value
-    ?set_attribute(<<"attr-2">>, {not_allowed, in, attributes}),
+    %% dropped because of non-homogenous
+    ?set_attribute(<<"attr-2-dropped">>, {non_homogenous, <<"attributes">>}),
```
tsloughter, line 734: Can you add a test to show that a homogenous tuple is accepted?

apps/opentelemetry_api/src/otel_span.erl
```diff
@@ -62,6 +86,67 @@ is_valid(#span_ctx{trace_id=TraceId,
 is_valid(_) ->
     false.
 
+-spec is_valid_attribute(opentelemetry:attribute_key(), opentelemetry:attribute_value()) -> boolean().
+is_valid_attribute(Key, Value) when is_tuple(Value) , ?is_allowed_key(Key) ->
+    is_valid_attribute(Key, tuple_to_list(Value));
+is_valid_attribute(Key, [Value1 | _Rest] = Values) when is_binary(Value1) , ?is_allowed_key(Key) ->
```
tsloughter, line 92:
```suggestion
%% lists as attribute values must be primitive types and homogeneous
is_valid_attribute(Key, [Value1 | _Rest] = Values) when is_binary(Value1) , ?is_allowed_key(Key) ->
```

Just thought a comment here would be helpful for the future.

apps/opentelemetry_api/src/otel_tracer.erl
```diff
@@ -43,27 +43,47 @@
 -callback start_span(otel_ctx:t(),
                      opentelemetry:tracer(),
                      opentelemetry:span_name(),
-                     otel_span:start_opts()) -> opentelemetry:span_ctx().
+                     otel_span:start_opts()) -> opentelemetry:span_ctx() | undefined.
 -callback with_span(otel_ctx:t(), opentelemetry:tracer(),
-                    opentelemetry:span_name(), otel_span:start_opts(), traced_fun(T)) -> T.
+                    opentelemetry:span_name(), otel_span:start_opts(), traced_fun(T)) -> T | any().
```
tsloughter, line 48: Why adding `any()` here?

apps/opentelemetry_api/src/otel_tracer.erl
```diff
@@ -43,27 +43,47 @@
...
+                -> opentelemetry:span_ctx() | undefined.
+start_span(Ctx, Tracer={Module, _}, SpanName, Opts) ->
+    case otel_span:is_valid_name(SpanName) of
+        true ->
+            Module:start_span(Ctx, Tracer, SpanName, otel_span:validate_start_opts(Opts));
+        false ->
+            undefined
+    end.
+
+-spec with_span(opentelemetry:tracer(), opentelemetry:span_name(), otel_span:start_opts(), traced_fun(T)) -> T | any().
```
tsloughter, line 70: Same with the `any()` here.

apps/opentelemetry_api/src/otel_tracer.erl
```diff
@@ -43,27 +43,47 @@
...
 with_span(Tracer={Module, _}, SpanName, Opts, Fun) when is_atom(Module) ->
-    Module:with_span(otel_ctx:get_current(), Tracer, SpanName, Opts, Fun).
-
--spec with_span(otel_ctx:t(), opentelemetry:tracer(), opentelemetry:span_name(), otel_span:start_opts(), traced_fun(T)) -> T.
+    case otel_span:is_valid_name(SpanName) of
+        true ->
+            Module:with_span(otel_ctx:get_current(), Tracer, SpanName, otel_span:validate_start_opts(Opts), Fun);
+        false ->
+            Fun()
```
tsloughter, line 76: `Fun` is arity 1, need to pass a noop span here.

apps/opentelemetry_api/src/otel_tracer.erl
```diff
@@ -43,27 +43,47 @@
...
-    Module:start_span(Ctx, Tracer, Name, Opts).
-
--spec with_span(opentelemetry:tracer(), opentelemetry:span_name(), otel_span:start_opts(), traced_fun(T)) -> T.
+                -> opentelemetry:span_ctx() | undefined.
+start_span(Ctx, Tracer={Module, _}, SpanName, Opts) ->
+    case otel_span:is_valid_name(SpanName) of
+        true ->
+            Module:start_span(Ctx, Tracer, SpanName, otel_span:validate_start_opts(Opts));
+        false ->
+            undefined
```
tsloughter, line 67: As discussed on slack, I think this should return the `?NOOP_SPAN_CTX` from `otel_tracer_noop`. I need to verify in the spec, but I'm not sure what else it could return.

tsloughter (changes requested):

apps/opentelemetry/test/opentelemetry_SUITE.erl
bryannaegele: There's one in the API validations test [here](https://github.com/open-telemetry/opentelemetry-erlang/pull/332/files#diff-81ca87e336303c4f98ee125fabcd666912e558b3761db01845298907fb44cb84R84) but I can add a second one there.

tsloughter: Oh yea, ok, no need then.

apps/opentelemetry_api/src/otel_tracer.erl
bryannaegele: I think I added that from the `undefined` return. Shouldn't be needed. Same with those undefineds.

apps/opentelemetry/test/opentelemetry_SUITE.erl
bryannaegele: Nah. It was definitely needed. Missed a check on the singular `set_attribute/3` so there was a bug 😆

tsloughter: Ah ok, hehe.

Reminds me, should have a test for a bad span name when calling `with_span`. Since that would have caught the fact it was calling `Fun()` instead of `Fun(SpanCtx)`.

tsloughter (approved):

PR #333, opened 2021-12-26
bump version to 1.0.0-rc.4

tsloughter:
The API Application no longer uses a VERSION file because mix.exs
grabs the version to use from the src/opentelemetry_api.app.src
file, so it is no longer needed.

The Exporter Application sets its version manually in .app.src
so that its versions can be tracked separately from the SDK.


tsloughter:
Will wait for #332 before actually merging this and cutting releases.

bryannaegele (approved):

PR #336, opened 2021-12-28
elixir span docs: fix reference to attributes type

tsloughter:


tsloughter:
Added on a commit to fix some of the example code in the api.

bryannaegele (approved):

Issue #347, opened 2022-01-10
Exporter endpoint config not working when no port is provided.


julianocosta89:
Basically I'm just documenting here what we have discussed in this slack thread: https://cloud-native.slack.com/archives/C01N75YMZCN/p1641638605006400

I've started playing around with a sample app, just to get the gist of OpenTelemetry in Erlang and I'm able to send the spans through the otel-collector to Jaeger as documented here: https://github.com/julianocosta89/otel_getting_started.

I wanted to test it without the collector and I was using `https://webhook.site/` to validate the requests.
Basically what I did was changing the endpoint in my `sys.config` from `localhost` to my unique URL generated in the webhook page.

It happens that it doesn't work. No trace is sent.

I know that the endpoint would get `/v1/traces` appended to it, but still, I should get a request with the trace in the body.

When investigating further, I found out that if I use `https://webhook.site:443/` it works, but that's a port that shouldn't be required.
The same would happen if I use a default http port.

I believe that this is a bug, but please let me know if further details are required.

PR #348, opened 2022-01-11
use the default port of the scheme (http/https) instead of default OTLP port

tsloughter:
I think this makes more sense when the user gives an endpoint like `http://example.com`. But I'm actually not sure now that I'm opening this and need to try to find it in the spec.


tsloughter:
Ok, I think that the default endpoint including the port (see https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#configuration-options) makes it clear that without a port should just use the default scheme's port -- so 80 or 443.

julianocosta89:
Awesome @tsloughter!
I don't have the repo permissions but whenever this PR is merged it resolves #347.

ferd (approved):

PR #349, opened 2022-01-11
Fix using OTEL_METRICS_EXPORTER env variable

indrekj:
Previously, when env included `OTEL_METRICS_EXPORTER=none`, then the app
start failed with:

```
** (Mix) Could not start application opentelemetry: exited in: :opentelemetry_app.start(:normal, [])
    ** (EXIT) an exception was raised:
        ** (KeyError) key :metrics_exporter not found
            (opentelemetry 1.0.0) [redacted app]/deps/opentelemetry/src/otel_configuration.erl:194: :otel_configuration.update_config_map/5
            (stdlib 3.15) lists.erl:1267: :lists.foldl/3
            (opentelemetry 1.0.0) [redacted app]/deps/opentelemetry/src/otel_configuration.erl:98: :otel_configuration.general/2
            (stdlib 3.15) lists.erl:1267: :lists.foldl/3
            (opentelemetry 1.0.0) [redacted app]/deps/opentelemetry/src/opentelemetry_app.erl:27: :opentelemetry_app.start/2
            (kernel 8.0) application_master.erl:293: :application_master.start_it_old/4
```

This was because metrics_exporter was not defined in the configuration
map and the code tried to update the existing value:

```
    try transform(Transform, Value) of
        TransformedValue ->
            ConfigMap#{Key := TransformedValue}
```

Other solution would have been to use `maps:put` but I think it makes
more sense to treat it similarly to `traces_exporter`.


apps/opentelemetry/src/otel_configuration.erl
```diff
@@ -36,6 +36,7 @@
                resource_detector_timeout => integer(),
                text_map_propagators := [atom()],
                traces_exporter := {atom(), term()},
```
tsloughter, line 38:
```suggestion
               traces_exporter := {atom(), term()} | undefined,
```

Not related to your change but it did make me realize this should also have an `undefined` part of the type since it also supports `none` :)

tsloughter (approved):

tsloughter:
Thanks! But you aren't actually using the metrics, right? :)

indrekj:
> Thanks! But you aren't actually using the metrics, right? :)

No, we're not actually using it. We just have a global environment config that is shared between all services where we have explicitly marked the metrics exporter to be none.

Issue #350, opened 2022-01-21
Exporter expects `tls_certificate_check` to be started even when explicit `ssl_options` are configured


danmarcab:
In our elixir app we have the exporter configured like:

```
config :opentelemetry,
  text_map_propagators: [:trace_context],
  processors: [
    otel_batch_processor: %{
      exporter:
        {:opentelemetry_exporter,
         %{
           protocol: :grpc,
           endpoints: [
             {:http, 'localhost', 4317, [verify: :verify_none]}
           ]
         }}
    }
  ]
```

We are trying to upgrade the exporter to version 1.0 but we found out that even though we are explicitly passing `ssl_options` in our endpoint config, the exporter expects the `tls_certificate_check` app to be running. We would like to avoid it if possible.

We would expect `tls_certificate_check` to be called only when explicit `ssl_options` were not provided.

tsloughter:
If the issue is that the application is required and started then the fix will be to move it to `optional_applications` in `.app.src`

danmarcab:
> If the issue is that the application is required and started then the fix will be to move it to `optional_applications` in `.app.src`

Thank you for the reply 👍 .

In our case (elixir app using mix) `tls_certificate_check ` is not started by default, which is fine. The problem is a call to the `tls_certificate_check` is made and then the app errors saying it wasn't started.

I guess moving to optional_applications would still be good to do. Happy to change it if you think we should.

kroehre:
FWIW I was getting this error and resolved by passing `ssl_options` to the _exporter_ config:
```
config :opentelemetry,
  text_map_propagators: [:trace_context],
  processors: [
    otel_batch_processor: %{
      exporter:
        {:opentelemetry_exporter,
         %{
           protocol: :grpc,
           endpoints: ["http://localhost:4317"],
           ssl_options: [verify: :verify_none]
         }}
    }
  ]
```

dvic:
In our case we don't have this problem as long as opentelemetry_exporter is before the other opentelemetry dependencies in mix.exs. This also makes it load with mix releases.

dvic:
> In our case we don't have this problem as long as opentelemetry_exporter is before the other opentelemetry dependencies in mix.exs. This also makes it load with mix releases.

Scratch this, this seemed to be a flaky fix. Currently we need to do the following to make this work:

- add  releases configuration in `mix.exs`
```
      releases: [
        my_app: [
          applications: [tls_certificate_check: :permanent]
        ]
[...]

PR #351, opened 2022-01-21
Only call tls_certificate_check when no explicit ssl_options provided

danmarcab:
Should fix https://github.com/open-telemetry/opentelemetry-erlang/issues/350

Tries to mimic what is done in:

https://github.com/open-telemetry/opentelemetry-erlang/blob/main/apps/opentelemetry_exporter/src/opentelemetry_exporter.erl#L268-L273

### DIFF
--- a/apps/opentelemetry_exporter/src/opentelemetry_exporter.erl
+++ b/apps/opentelemetry_exporter/src/opentelemetry_exporter.erl
@@ -278,15 +278,14 @@ parse_endpoint({Scheme, Host, Port, _}, DefaultSSLOpts) ->
              port => Port,
              path => [],
              ssl_options => update_ssl_opts(HostString, DefaultSSLOpts)}};
-parse_endpoint(Endpoint=#{host := Host, port := _Port, scheme := _Scheme, path := Path}, DefaultSSLOpts) ->
-    HostString = unicode:characters_to_list(Host),
-    {true, maps:merge(#{host => HostString,
+parse_endpoint(Endpoint=#{host := Host, scheme := _Scheme, path := Path, ssl_options := SSLOptions}, _DefaultSSLOpts) when is_list(SSLOptions) ->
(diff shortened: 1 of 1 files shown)


tsloughter:
But it only calls `tls_certificate_check` if the ssl opts are `undefined`. So this shouldn't change whether it is called or not?

apps/opentelemetry_exporter/src/opentelemetry_exporter.erl
```diff
@@ -278,15 +278,14 @@ parse_endpoint({Scheme, Host, Port, _}, DefaultSSLOpts) ->
...
              path => [],
              ssl_options => update_ssl_opts(HostString, DefaultSSLOpts)}};
-parse_endpoint(Endpoint=#{host := Host, port := _Port, scheme := _Scheme, path := Path}, DefaultSSLOpts) ->
-    HostString = unicode:characters_to_list(Host),
-    {true, maps:merge(#{host => HostString,
+parse_endpoint(Endpoint=#{host := Host, scheme := _Scheme, path := Path, ssl_options := SSLOptions}, _DefaultSSLOpts) when is_list(SSLOptions) ->
+    {true, maps:merge(#{host => unicode:characters_to_list(Host),
                         path => unicode:characters_to_list(Path),
                         port => ?DEFAULT_PORT,
-                        ssl_options => update_ssl_opts(HostString, DefaultSSLOpts)}, Endpoint)};
```
danmarcab, line 286: > But it only calls `tls_certificate_check` if the ssl opts are `undefined`. So this shouldn't change whether it is called or not?

Note here `update_ssl_opts` was called with `DefaultSSLOpts`, which in our case are indeed `undefined` (and therefore `tls_certificate_check` will be called), but our `Endpoint` contained `ssl_options` as a key, which means it didn't need to be called.

PR #352, opened 2022-01-24
Update documentation on configuring the exporter via a map

maciej-szlosarczyk:
- Add info on compression
- Describe all 3 ways in which an endpoint can be configured
- Add information on default port
- Add the gotcha about using only the first endpoint
- Fix a minor typo

AKA all the things that one could only learn from reading the source code.

### DIFF
--- a/apps/opentelemetry_exporter/README.md
+++ b/apps/opentelemetry_exporter/README.md
@@ -54,19 +54,8 @@ The second element of the configuration tuple is a configuration map. It can con
 - `protocol` - one of: `http_protobuf`, `grpc` or `http_json`. Defaults to `http_protobuf`. `http_json` is not implemented yet.
 - `endpoints` - A list of endpoints to send traces to. Can take one of the forms described below. By default, exporter sends data to `http://localhost:4318`.
 - `headers` - a list of headers to send to the collector (i.e `[{<<"x-access-key">> <<"secret">>}]`). Defaults to an empty list.
-- `compression` - An atom. Setting it to `gzip` enables gzip compression.
-
-### Endpoints configuration
-
-You can pass your collector endpoints in three forms:
-
-- As a string, i.e `"https://localhost:4000"`.
-- As a map, with the following keys: `#{host => unicode:chardata(), path => unicode:chardata(), port => integer() >= 0 | undefined, scheme => unicode:chardata()}`
-- As a 4 element tuple in format `{Scheme, Host, Port, SSLOptions}`.
-
-Unless specified directly, the port is not inferred from scheme, but defaults to `4318`. If you want to use standard port 443, you need to specify it.
-
-While using `http_protobuf` protocol, currently only the first endpoint in that list is used to export traces, the rest is effectively ignored. `grpc` supports multiple endpoints.
+- `compression` - an atom. Setting it to `gzip` enables gzip compression.
+- `ssl_options` - a list of SSL options. See Erlang's [SSL docs](https://www.erlang.org/doc/man/ssl.html#TLS/DTLS%20OPTION%20DESCRIPTIONS%20-%20CLIENT) for what options are available.
 
 ### Application Environment
 
--- a/apps/opentelemetry_exporter/README.md
+++ b/apps/opentelemetry_exporter/README.md
@@ -12,14 +12,14 @@ Currently only supports the Tracer protocol using either GRPC or Protobuffers ov
 
 ### Options to Batch Processor
 
-Exporter configuration can be done as arguments to the batch processor in
-the OpenTelemetry application environment.
+Exporter configuration can be done as arguments to the batch processor in the OpenTelemetry application environment.
+
 
 For an Erlang release in `sys.config`:
 
 ``` erlang
 {opentelemetry,
-  [{processors, 
+  [{processors,
     [{otel_batch_processor,
         #{exporter => {opentelemetry_exporter, #{endpoints =>
         ["http://localhost:9090"],
@@ -31,7 +31,7 @@ The default protocol is `http_protobuf`, to override this and use grpc add
 
 ``` erlang
 {opentelemetry,
-  [{processors, 
+  [{processors,
     [{otel_batch_processor,
         #{exporter => {opentelemetry_exporter, #{protocol => grpc,
                                                  endpoints => ["http://localhost:9090"],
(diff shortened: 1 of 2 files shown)


apps/opentelemetry_exporter/README.md
```diff
@@ -48,6 +48,38 @@ config :opentelemetry, :processors,
   }
 ```
 
+  [{processors,
+    [{otel_batch_processor, #{exporter => {opentelemetry_exporter, #{}}}}]}]}
+```
+
+An Elixir release uses `releases.exs` or `runtime.ex`:
```
tsloughter, line 55: Why mention `releases.exs`?

Eek, except now I see that the original readme, that I wrote, uses `releases.exs`, haha. But `runtime.exs` is what we should be pushing strongly, right?

apps/opentelemetry_exporter/README.md
```diff
@@ -48,6 +48,38 @@ config :opentelemetry, :processors,
...
+
+### Endpoints configuration
+
+You can pass your collector endpoints in three forms:
+
+- As a string, i.e `"https://localhost:4000"`.
+- As a map, with the following keys: `#{host => unicode:chardata(), path => unicode:chardata(), port => integer() >= 0 | undefined, scheme => unicode:chardata()}`
+- As a 4 element tuple in format `{Scheme, Host, Port, SSLOptions}`.
+
+Unless specified directly, the port is not inferred from scheme, but defaults to `4318`. If you want to use standard port 443, you need to specify it.
```
tsloughter, line 79: I was actually going to be merging https://github.com/open-telemetry/opentelemetry-erlang/pull/348 which would make this not the case.

Is there any reason to keep the current behaviour or are you just documenting the current behaviour even tho you agree it is odd? :)

apps/opentelemetry_exporter/README.md
```diff
@@ -48,6 +48,38 @@ config :opentelemetry, :processors,
...
+
+Also, while the endpoints value can be a list, currently only the first endpoint in that list is used to export traces, the rest is effectively ignored.
```
tsloughter, line 81: This is not true when using grpc. The grpcbox library supports multiple endpoints and all are passed to it.

apps/opentelemetry_exporter/README.md
```diff
@@ -12,14 +12,14 @@ Currently only supports the Tracer protocol using either GRPC or Protobuffers ov
 
 ### Options to Batch Processor
 
-Exporter configuration can be done as arguments to the batch processor in
-the OpenTelemetry application environment.
+Exporter configuration can be done as arguments to the batch processor in the OpenTelemetry application environment.
```
tsloughter, line 15:
```suggestion
Exporter configuration can be done as arguments to the span processor (simple or batch) in the OpenTelemetry application environment.
```

This was originally written before there was `otel_simple_processor`. This change will require a couple other changes, like the headline of this section should drop `Batch`.

Fine if you prefer to keep this PR out of the need to update to include references to the simple processor and you can ignore this suggestion.

apps/opentelemetry_exporter/README.md
maciej-szlosarczyk: >  are you just documenting the current behaviour even tho you agree it is odd? :)

Very much this, the other day I spent quite some time to figure out why am I not getting any traces when I forgot setting the port number to 443. If #348 is merged earlier, I'll remove it.

apps/opentelemetry_exporter/README.md
maciej-szlosarczyk: Runtime.exs is the equivalent of `sys.config` - it is the file distributed when you build a release and you can edit it on the target machine. `runtime.exs` is also read while running the application straight from Mix. Both will be [supported going forward](https://elixir-lang.org/blog/2020/10/06/elixir-v1-11-0-released/):

> While config/releases.exs will continue to be supported, developers can migrate to config/runtime.exs without loss of functionality.

I guess what we want to say here is that it is configured before application start. So maybe the best wording would be: 

> In Elixir, you can configure it with `config.exs` or `runtime.exs`.

apps/opentelemetry_exporter/README.md
maciej-szlosarczyk: I missed that, updated now.

apps/opentelemetry_exporter/README.md
tsloughter: Just merged it.

apps/opentelemetry_exporter/README.md
```diff
@@ -48,6 +47,27 @@ config :opentelemetry, :processors,
...
+### The configuration map
+
+The second element of the configuration tuple is a configuration map. It can contain the following keys:
+
+- `protocol` - one of: `http_protobuf`, `grpc` or `http_json`. Defaults to `http_protobuf`. `http_json` is not implemented yet.
+- `endpoints` - A list of endpoints to send traces to. Can take one of the forms described below. By default, exporter sends data to `http://localhost:4318`.
+- `headers` - a list of headers to send to the collector (i.e `[{<<"x-access-key">> <<"secret">>}]`). Defaults to an empty list.
+- `compression` - An atom. Setting it to `gzip` enables gzip compression.
+
+### Endpoints configuration
```
tsloughter, line 59: I think we should only document, at least in the readme, it should of course be documented in the module docs, the string endpoint.

I wish the others were never supported but its too late for that now. The only thing you lose by not having the other options is the lack of ssl options per-endpoint, but I doubt that is used.

apps/opentelemetry_exporter/README.md
```diff
@@ -48,6 +47,27 @@ config :opentelemetry, :processors,
...
 ```
 
+### The configuration map
+
+The second element of the configuration tuple is a configuration map. It can contain the following keys:
+
+- `protocol` - one of: `http_protobuf`, `grpc` or `http_json`. Defaults to `http_protobuf`. `http_json` is not implemented yet.
+- `endpoints` - A list of endpoints to send traces to. Can take one of the forms described below. By default, exporter sends data to `http://localhost:4318`.
+- `headers` - a list of headers to send to the collector (i.e `[{<<"x-access-key">> <<"secret">>}]`). Defaults to an empty list.
+- `compression` - An atom. Setting it to `gzip` enables gzip compression.
```
tsloughter, line 57:
```suggestion
- `compression` - an atom. Setting it to `gzip` enables gzip compression.
- `ssl_options` - a list of SSL options. See Erlang's [SSL docs](https://www.erlang.org/doc/man/ssl.html#TLS/DTLS%20OPTION%20DESCRIPTIONS%20-%20CLIENT) for what options are available.
```

apps/opentelemetry_exporter/README.md
```diff
@@ -31,14 +30,14 @@ The default protocol is `http_protobuf`, to override this and use grpc add
 
 ``` erlang
 {opentelemetry,
-  [{processors, 
-    [{otel_batch_processor,
+  [{processors,
+    [{otel_simple_processor,
```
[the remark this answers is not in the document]
maciej-szlosarczyk: Because of this comment: https://github.com/open-telemetry/opentelemetry-erlang/pull/352#discussion_r791218479

apps/opentelemetry_exporter/README.md
```diff
@@ -31,14 +30,14 @@ The default protocol is `http_protobuf`, to override this and use grpc add
...
         #{exporter => {opentelemetry_exporter, #{protocol => grpc,
                                                  endpoints => ["http://localhost:9090"],
                                                  headers => [{"x-honeycomb-dataset", "experiments"}]}}}}]}]}
 ```
 
-An Elixir release uses `releases.exs`:
+In Elixir, you can use `config.exs` or `runtime.exs`:
```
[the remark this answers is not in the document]
maciej-szlosarczyk: https://github.com/open-telemetry/opentelemetry-erlang/pull/352#discussion_r791401829

I think it needs to be both - while `runtime.exs` replaced `releases.exs` as the suggested way to configure releases, not every project uses it now or will use it in the future. `config.exs` is the most common denominator for configuration, to quote the [documentation](https://hexdocs.pm/elixir/1.13.3/Config.html):

> This module is most commonly used to define application configuration, typically in config/config.exs:

And about runtime.exs:

>For runtime configuration, you *can* use the config/runtime.exs file. It is executed right before applications start in both Mix and releases (assembled with [mix release](https://hexdocs.pm/mix/1.13/Mix.Tasks.Release.html)).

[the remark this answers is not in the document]
tsloughter: Ok, yup, this is good :+1:

PR #353, opened 2022-02-02
fix for links on a span and fix testing of span data

tsloughter:
I hate that there are so many ways to represent links and events... but I guess it is too late to fix because it would technically break the API :(


bryannaegele:
What was breaking?

ferd (approved):

tsloughter:
@bryannaegele only thing broken was the test

PR #354, opened 2022-02-03
Fix bug related to event timestamp format.

derekkraan:
The exporter assumes that we are using `opentelemetry:timestamp/0`, which returns a timestamp in the erlang native format. This commit cleans up use of `erlang:system_time/1`.


tsloughter:
Agh, thanks! I vaguely recall meaning to look into the use of timestamps and why there were multiple forms but can't remember why... this is definitely the right fix.

tsloughter (approved):

PR #359, opened 2022-02-10
Build Erlang and Elixir docs for opentelemetry_api

erszcz:
Hey, @tsloughter, this builds the ExDoc HTML docs for Erlang and Elixir modules for `opentelemtry_api`, as discussed on the EEF Slack.


tsloughter:
Thanks!

PR #360, opened 2022-02-12
Readme updates

tsloughter:


apps/opentelemetry/README.md
```diff
@@ -0,0 +1,170 @@
...
+```
+
+The following sections detail the available SDK options and how to configure
+them through either the OS or Application environment.
+
+### Automatically Created Tracers
+
+Spans are created by Tracers and each Tracer is associated with a [name and
+version](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.8.0/specification/trace/api.md#get-a-tracer).
+The name and version can be anything but the macros provided by the API will use
```
ferd, line 51:
```suggestion
The name and version can be anything, but the macros provided by the API will always use
```
insert comma to avoid confusion between "anything but X" meaning "all the things except X" vs. the intended sense here which is "it can be anything, but <unrelated statement>."

I added the 'always' as well because I think it's the intended meaning and removes ambiguity.

apps/opentelemetry/README.md
```diff
@@ -0,0 +1,170 @@
...
+number of modules in a Release the time to create all these Tracers can become
+more than you want to wait on boot.
+
+| OS                              | Application                | Default | Type    |
+|:--------------------------------|:---------------------------|:--------|:--------|
+| OTEL_CREATE_APPLICATION_TRACERS | create_application_tracers | true    | boolean |
+
+Disabling the creation of Tracers at start-up means when the macros are used to
+start Spans the name and version of the Tracer creating those Spans will be
+undefined, but everything will continue to work as before.
```
ferd, line 64: Is there any link or docs for the tracer API/macros we can link here?

ferd (approved):

apps/opentelemetry/README.md
tsloughter: Not yet, since you can't get docs for a header of macros I'm going to be adding docs for each macro to the readme of the API.

PR #361, opened 2022-02-15
fix: readme elixir release config instructions

jhonndabi:
`releases.applications` must be a keyword list. If we try to generate a release with the configuration like the currently one in the README.md. 
`[:runtime_tools, :opentelemetry_exporter, {:opentelemetry, :temporary}, :my_instrumented_app]`
Then the following error will raise.

```
Generated basic_elixir app
** (ArgumentError) expected a keyword list as the second argument, got: [:runtime_tools, :opentelemetry_exporter, {:opentelemetry, :temporary}, :basic_elixir]
    (elixir 1.12.2) lib/keyword.ex:786: Keyword.merge/2
    (mix 1.12.2) lib/mix/release.ex:175: Mix.Release.find_release/2
    (mix 1.12.2) lib/mix/release.ex:77: Mix.Release.from_config!/3
    (mix 1.12.2) lib/mix/tasks/release.ex:1032: Mix.Tasks.Release.run/1
    (mix 1.12.2) lib/mix/task.ex:394: anonymous fn/3 in Mix.Task.run_task/3
    (mix 1.12.2) lib/mix/cli.ex:84: Mix.CLI.run_task/2
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ def project do
   [
     releases: [
       my_instrumented_release: [
-        applications: [opentelemetry_exporter: :permanent, opentelemetry: :temporary, my_instrumented_app: :permanent]
+        applications: [opentelemetry_exporter: :permanent, opentelemetry: :temporary]
       ],
 
       ...
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ def project do
   [
     releases: [
       my_instrumented_release: [
-        applications: [:runtime_tools, :opentelemetry_exporter, {:opentelemetry, :temporary}, :my_instrumented_app]
+        applications: [opentelemetry_exporter: :permanent, opentelemetry: :temporary, my_instrumented_app: :permanent]
       ],
 
       ...


README.md
```diff
@@ -76,7 +76,7 @@ def project do
   [
     releases: [
       my_instrumented_release: [
-        applications: [:runtime_tools, :opentelemetry_exporter, {:opentelemetry, :temporary}, :my_instrumented_app]
+        applications: [opentelemetry_exporter: :permanent, opentelemetry: :temporary, my_instrumented_app: :permanent]
```
bryannaegele, line 79: The app also doesn't need to be included in the list afaik. It should run automatically after these load in order.

```suggestion
        applications: [opentelemetry_exporter: :permanent, opentelemetry: :temporary]
```

tsloughter:
Weird, so mix doesn't accept anything like `[:runtime_tools, :opentelemetry_exporter]`?

The docs say that `permanent` is the default, don't know why it would require you specify the type for every app.

jhonndabi:
@tsloughter No, because isn't a valid Keyword, and they don't inject default values to make a valid Keyword.

https://github.com/elixir-lang/elixir/blob/v1.13.3/lib/mix/lib/mix/release.ex#L175
https://github.com/elixir-lang/elixir/blob/v1.13.3/lib/elixir/lib/keyword.ex#L129

bryannaegele:
It looks like there's a mismatch in functionality between applications defined at the project level vs the release level. The way you have it @tsloughter is fine for project but not releases.

https://github.com/elixir-lang/elixir/blob/v1.13.3/lib/mix/lib/mix/tasks/compile.app.ex#L293-L299
https://github.com/elixir-lang/elixir/blob/v1.13.3/lib/mix/lib/mix/tasks/compile.app.ex#L335-L337

bryannaegele (approved):

tsloughter:
@bryannaegele eh, it is a list of applications for different reasons so expected to be different types in the list. I just expected it to take a list the same as the ones taken by relx and reltool.

bryannaegele:
Yeah, the treatment by elixir/mix just doesn't seem consistent

PR #362, opened 2022-02-15
Add docs for the tracing api macros to the readme

tsloughter:


tsloughter:
There is a lot more to go but this is a start.

apps/opentelemetry_api/README.md
```diff
@@ -30,17 +30,175 @@ some_fun() ->
...
+
+A Span represents a single operation in a Trace. It has a start and end time,
+can have a single parent and one or more children. The easiest way to create
+Spans is to wrap the operation you want a Span to represent in the `with_span`
+macro. The macro handles getting a
+[Tracer](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.8.0/specification/trace/api.md#tracer)
+associated with the OTP Application the module is in, starting the Span, setting
+it as the currently active Span in the Context stored in the process dictionary
+and ending the Span when the `Fun` or body of the Elixir macro finish, even if
+an exception is thrown. After the Span is end'ed the Context in the process
```
hauleth, line 82: I would note there that the exception will be reraised.

apps/opentelemetry_api/README.md
```diff
@@ -30,17 +30,175 @@ some_fun() ->
...
+
+#### Setting Attributes
+
+[Setting
+Attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.8.0/specification/trace/api.md#set-attributes)
+can be done with a single key and value passed to `set_attribute` or through a
+map of
+[Attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.8.0/specification/common/common.md#attributes)
+all at once. Setting an attribute with a key that already exists in the Span's
+map of attributes will result in that key's value being overwritten.
```
hauleth, line 136: Is there a way to unset the attribute? Or maybe there is an value that will be ignored when serialising to wire format?

apps/opentelemetry_api/README.md
```diff
@@ -30,17 +30,175 @@ some_fun() ->
...
+
+[Set
+Status](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.8.0/specification/trace/api.md#set-status)
+will override the default Span Status of `Unset`. A Status is a code (`ok`,
+`error` or `unset`) and, only if the code is `error`, an optional message string
+that describes the error.
+
+``` erlang
+?set_status(Code, Message)
+```
```
hauleth, line 182: If `Message` is optional if `Code =/= error`, then maybe `set_status/1` would also make sense?

apps/opentelemetry_api/README.md
tsloughter: Good point, will add.

apps/opentelemetry_api/README.md
tsloughter: Hm, I guess not.

apps/opentelemetry_api/README.md
tsloughter: Ah yea, ok. Problem is there is already a `set_status/1` that accepts the `status` record. I hate overloading functions by checking the type of the argument but it may just need to be done here.

apps/opentelemetry_api/README.md
tsloughter: actually, I guess it is never caught, so not reraised, but I can mention this. I thought the Elixir version did catch it and then did a `record_exception` and a `raise`, but that isn't the case.

tsloughter:
@hauleth I've updated per your comments.

tsloughter:
Since this has `set_status` updates I've included this fix to follow the spec for how status can be updated https://github.com/open-telemetry/opentelemetry-erlang/pull/362/files#diff-748a4857daeb19ea7dace6649fa9614a9de90dcd2169a035ed39e9391745c132R140

apps/opentelemetry_api/README.md
olivermt: set it to something that cant serialize to json(?) and you essentially unset it, I have found this out myself by sending structs to set_attribute without inspect()'ing it.

jhonndabi: @hauleth I think it is possible through agent/collector [attributesprocessor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/attributesprocessor)

```
...
processors:
  attributes/example:
    actions:
      - key: db.table
        action: delete
...
```

apps/opentelemetry_api/README.md
```diff
@@ -30,17 +30,176 @@ some_fun() ->
...
+Spans is to wrap the operation you want a Span to represent in the `with_span`
+macro. The macro handles getting a
+[Tracer](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.8.0/specification/trace/api.md#tracer)
+associated with the OTP Application the module is in, starting the Span, setting
+it as the currently active Span in the Context stored in the process dictionary
+and ending the Span when the `Fun` or body of the Elixir macro finish, even if
+an exception is thrown -- however, the exception is not caught, so it does not
+change how user code should deal with raised exceptions. After the Span is
+end'ed the Context in the process dictionary is reset to its value before the
+newly started Span was set as the active Span.
```
ferd, line 85: I know this is a good thing because it properly deals with span lineage, but it may be worth stating it.

apps/opentelemetry_api/README.md
```diff
@@ -30,17 +30,176 @@ some_fun() ->
...
+end
+```
+
+`StartOpts`/`start_opts` is a map of [Span creation options](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.8.0/specification/trace/api.md#span-creation):
+
+- `kind`: 
+[SpanKind](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.8.0/specification/trace/api.md#spankind)
+defines the relationship between the Span, its parents, and its children in a
+Trace. Possible values: `internal`, `server`, `client`, `producer` and
+`consumer`.
```
ferd, line 103: does it have default values, and if so what is it? Same question for following options.

ferd (approved):

apps/opentelemetry_api/README.md
tsloughter: Ah nice @jhonndabi

apps/opentelemetry_api/README.md
tsloughter: Ah  yup, actually tried to include that originally but the wording always sounded funny I thought. Trying again with something simpler.

apps/opentelemetry_api/README.md
tsloughter: Yup, fixing.

tsloughter:
Ugh, I don't know why elixir tests at the top level are now failing.

DavidOJ315 (approved):
>

PR #363, opened 2022-02-16
improvements to zipkin exporter

tsloughter:
Zipkin tags only support strings for values so we needed to be converting attribute values that are list to a string with `io_lib:format`. 

Other change is converting an otel status to tags.


apps/opentelemetry_zipkin/src/opentelemetry_zipkin.erl
```diff
@@ -103,22 +103,36 @@ to_string(Value) when is_list(Value) ;
...
+to_binary_string(Value) when is_binary(Value) ->
+    Value;
 to_binary_string(Value) ->
-    Value.
-
-to_tags(Attributes) ->
+    %% attributes already check that the value is a binary string
+    %% and not a list string, so any other term, like a list,
+    %% can just be turned into a string of term
+    unicode:characters_to_binary(io_lib:format("~tp", [Value])).
```
ferd, line 122: The result type here is:

```
binary() |
{error, binary(), RestData} |
{incomplete, binary(), binary()}
```
it might be worth asserting that you get an actual binary() to avoid passing invalid data and crash right at the conversion instead.

ferd (changes requested):

apps/opentelemetry_zipkin/src/opentelemetry_zipkin.erl
tsloughter: Ah, yup. Funny enough I did this all over the code base. I have a branch that I've been cleaning that stuff up because Gradualizer catches it.

ferd (approved):

PR #366, opened 2022-02-24
additional doc for config grpcbox for grpc protocol

calvin-kargo:
Add doc on readme for channel config of grpcbox for use with grpc protocol

i have tried out `opentelemetry_exporter` using `grpc`, but if used as is, it will throw `:undefined_channel` error, which after some debugging turn out caused due to `grpcbox` require channel for `opentelemetry_exporter` to be specified


tsloughter:
This shouldn't be the case. Setting up a channel manually like this may work but it isn't required. Something else must be going wrong. Can you open an issue or are you in one of the Slacks where we can discuss? (erlang, elixir or cncf slack)

calvin-kargo:
here's the issue @tsloughter  https://github.com/open-telemetry/opentelemetry-erlang/issues/370

PR #380, opened 2022-03-26
start on new metrics API

tsloughter:
I'm not happy with it yet but wanted to open this for feedback.

My issue is right now an `Instrument` has to be passed to modules like `otel_counter`. This variable contains everything, including a description which could be long, having to copy that around everywhere you want to record a measurement for an instrument seems inefficient. But the Otel spec defines that everything in there is identifying, so simply the name, or even name+type combo, do not give you a reference to the unique instrument.

I think we might end up doing something where there is a separate Erlang/Elixir specific registration name -- then instrument name would be a string instead of atom -- and use that instead of the full Instrument.

Figured I'd get feedback both from here and from the metrics SIG before moving forward with anything like that.


ferd:
Yeah the spec explicitly calls all fields as identifying, but also states that two (the description and unit) are optional, and the section on [conflict detection](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument-type-conflict-detection) is only concerned with warning on a name clash, and must always return a valid instrument regardless.

I guess the critical bit is:

> It is unspecified whether or under which conditions the same or different Instrument instance will be returned as a result of duplicate instrument registration. The term identical applied to Instruments describes instances where all identifying fields are equal. The term distinct applied to Instruments describes instances where at least one field value is different.

which means that we should definitely fetch based on the entirety of all fields. 

The description can be up to 1023 chars, which at least means that using a binary can let you get a refcounted (>64 bytes/chars and up) one that won't eat up that much RAM. I'm not sure I'd be too concerned in such cases, it sort of puts an upper bound to how much data gets copied, less than some floating points or integer values may take? I'd have said that if we needed something bigger or nicer it'd make sense to register fields and carry a ref as a sort of hash, but I'm not sure it'd be that much?

tsloughter:
Ah, yea, I guess I failed to mention the other reason I don't like having to carry around an Instrument is the very fact you need the variable instead of just the name. I had thought about if memory copying was an issue (agree it probably isn't) we could create a ref of some kind to return and act as the Instrument variable.

But relying on a variable means more complicated handling. Simple example would be an HTTP server like Elli. Instead of being able to have the middleware do like:

```
otel_counter:add('elli.requests', 1, #{})
```

To use the Instrument they must be created and passed through `{callback_args, _}` to the handler, I guess in a map so they can be looked up when needed by name:

```
handle(Req, #{instruments := Instruments}) ->
  RequestsInstrument = maps:get('elli.requests', Instruments, undefined),
  otel_counter:add(RequestsInstrument, 1, #{})
```

This particular example would be a middleware and now the end user doing it, but the same would need to be done for user's custom metrics they want to use in a handler.

ferd:
A ref still requires a variable. Do you mean something more like an Alias? I wonder if it would be "legal" to add a sort of `alias(Map, Name) -> ok` call that lets you overload the API and pass in `{alias, Name}` and has the general implementation MITM the call to `unalias(Name) -> Map` and transparently pretend it was done as in the spec, but with an indirect ETS (or persistent term) lookup.

Otherwise I figure the sort of ordeal you get is that the instrument should be put in a .hrl file as a macro such that your example becomes `otel_counter:add(?INSTR_ELLI_REQUESTS, 1, #{})`, which is safe but sort of annoying.

tsloughter:
I meant that I wouldn't use a ref since it would still require a variable, so nothing much is gained, only the very minimal gain of not copying the other data.

Using an additional identifier atom we use that is not the same as the instrment's `name` -- and unique to our implementation -- is what I was thinking would be optimal if we don't just pass around the Instrument.

tsloughter:
Oh, and the macro in a header doesn't work sadly. I would have gone with that if it did. It may be possible to change the design but right now to get an Instrument you call `otel_meter:create_counter` which is able to set the module that the Instrument needs to call in the SDK when `otel_counter:add(Instrument ...)` is called.

So it is partially populated at runtime.

tsloughter:
Ok, I've realized I think I read way too much into the fact that description is an "identifying field".

I think the point is to just say when you need to log a warning about the fact that the user is creating an instrument with the same name but a different other identifying field. But that basing the lookup of instruments on only the name+meter works just fine, no need to keep around instruments with the same name but different descriptions.

tsloughter:
To be clearer, if every identifying field is the same then you just return the same instrument, if something is different, like the description, you log a warning and maybe return the original instrument (that is unspecified).

> It is unspecified whether or under which conditions the same or different Instrument instance will be returned as a result of duplicate instrument registration. The term identical applied to Instruments describes instances where all identifying fields are equal. The term distinct applied to Instruments describes instances where at least one field value is different.

> When more than one distinct Instrument is registered with the same name for identical Meters, the implementation SHOULD emit a warning to the user informing them of duplicate registration conflict(s).

https://github.com/open-telemetry/opentelemetry-specification/blob/a7a08e11e7a5988adf9dc83f11bd57c0e5110c6f/specification/metrics/api.md#instrument-type-conflict-detection

@ferd and anyone else reading this, do you agree that is what this is saying? 

Do you also agree it is a bit difficult to read? If it isn't just me then I'll work on a PR to clear up the wording.

My first thought on rewording is to lead with "same name for identical Meters" to make clear those are the "primary keys" and that "identifying fields" are not.

It could maybe use better wording than "identifying fields" but I don't know what to use instead.

ferd:
Yeah my sort of reading is that you could have many such meters existing at once, and that if some fields vary you warn. When you want to fetch one of them, you don't know which one you'll get across all the existing ones due to the overlapping names.

tsloughter:
I assume you mean "instrument" when you say "meter". But that isn't how I read it (assuming you mean instrument), I now read it as saying you'll only have 1 but whether you need to log a warning or just return the existing instrument with the same name is if all identifying fields match or not.

tsloughter:
Closing this and will be opening a new PR that changes the API slightly and adds a first take on the SDK.

The main change to the API is `create_*` no longer will make a `call` to the meter provider, it just constructs a record. So there is no overhead to creating instruments.

PR #384, opened 2022-04-26
set_status: use update_element since select_replace doesn't use index

tsloughter:


ferd (approved):
I figure there's a risk where you can run into races with this, but given the current format of the writes and the possibility of getting a `false` result if a delete has occured, this appears to be safe in context given the current operations.

ferd:
There could be a small race where the call to make a span status as ok after it was error interleaves oddly with scheduling and fails (by setting ok before error). I'm not sure how concerned you are with that.

tsloughter:
Yea, technically the possibility of a race is part of every operation in this module. But a span should only be updated by a single process. This is not currently enforced since it would require somehow tracking the "owner" (can't just be the process that started it since it could be created to be passed to a newly spawned process), but since the majority of usage will be as the current span stored in the process dictionary it isn't something too be concerned about I think.

PR #389, opened 2022-05-06
fix bug where port 443 is used even for https

tsloughter:
scheme_port/1 was expecting an atom when it was getting a string.
It would fall back to port 80 and this would result in failing to
connect to https endpoints unless they had an explicit port given
in the url.


tsloughter:
Sadly the endpoint code is all ugly and can't be cleaned up much without dropping support for the terrible idea of supporting the user passing in the already parsed url map or tuple that I allowed because I had grpcbox use it for some reason.

tsloughter:
Argh, `binary_to_existing_atom` was added in 23 :(

Not bothering with a patch for that until morning...

PR #392, opened 2022-05-12
fix lookup of exporter config to use right var name

tsloughter:


tsloughter:
Argh, this breaks the way of configuring the batch processor by passing a map of args instead of configuring with application env.

tsloughter:
Closing. Gotta make this work with the old and new way of configuring the exporter.

PR #394, opened 2022-05-12
add trace ctx to logger metadata when attached to pdict context

tsloughter:


tsloughter:
Still needs to show an example to at least make clear that you use a list of atoms to access the subkeys in logger formatter: `[otel_span_ctx, trace_id]`.

But maybe this is good enough aside from stuff like that I'll do before we merge.

ferd (approved):

bryannaegele:
@tsloughter we also need to clear it from metadata on detach

tsloughter:
@bryannaegele woop, good catch, I made sure setting to `undefined` worked but forgot about `detach` :)

tsloughter:
Updated. Also updated the changelog.

apps/opentelemetry_api/src/otel_ctx.erl
```diff
@@ -121,10 +121,13 @@ get_current() ->
 
 -spec attach(map()) -> token().
 attach(Ctx) ->
+    update_logger_process_metadata(Ctx),
     erlang:put(?CURRENT_CTX, Ctx).
 
 -spec detach(token()) -> ok.
 detach(Token) ->
+    %% at this time `Token' is a context
+    update_logger_process_metadata(Token),
```
bryannaegele, line 130: Is this not just setting the same value again rather than clearing or overriding it?

tsloughter: See the test, it checks after the with_span that the previous span is back in the metadata.

bryannaegele: Ah. Got it. So if there was no previous span, it would be `undefined`.

bryannaegele (approved):

btkostner:
Coming back from [the original discussion](https://github.com/open-telemetry/opentelemetry-erlang/discussions/391), I have deployed this to a staging Elixir project and can confirm trace metadata lines up with logs.

btkostner:
Coming back from [the original discussion](https://github.com/open-telemetry/opentelemetry-erlang/discussions/391), I have deployed this to a staging Elixir project and can confirm trace metadata lines up with logs.

apps/opentelemetry_api/src/otel_ctx.erl
tsloughter: Right.

btkostner:
@tsloughter So this is an Elixir project sending data to Datadog. For the logger backend, we've been using https://github.com/Nebo15/logger_json with the `DatadogLogger` formatter. Configuration looks like so:

```ex
# config.exs
config :logger,
  backends: [LoggerJSON],
  level: :info

config :logger_json, :backend,
  formatter: LoggerJSON.Formatters.DatadogLogger,
  metadata: :all
```

And then we use a custom plug to log Phoenix requests in a single output like so:

```ex
# endpoint.ex
plug Plug.RequestId
plug Plug.Telemetry, event_prefix: [:phoenix, :endpoint]

plug LoggerJSON.Plug, metadata_formatter: LoggerJSON.Plug.MetadataFormatters.DatadogLogger
```

The plug needs to come after `Plug.Telemetry` for the correct trace data, else it will be empty. And outputs logs like so:

```json
{
  "domain": ["elixir"],
  "duration": 5597270,
  "erl_level": "info",
  "http": {
    "url": "http://localhost/_health",
    "status_code": 200,
    "method": "GET",
    "referer": null,
    "request_id": "FvLDEr8jXx2K9E8AAAAB",
    "useragent": "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/102.0.0.0 Safari/537.36",
    "url_details": {
      "host": "localhost",
      "port": 4000,
      "path": "/_health",
      "queryString": "",
      "scheme": "http"
    }
  },
  "logger": {
    "thread_name": "#PID<0.616.0>",
    "method_name": "Elixir.LoggerJSON.Plug.call/2",
    "file_name": "lib/logger_json/plug.ex",
    "line": 44
  },
  "message": "",
  "network": {
    "client": {
      "ip": "127.0.0.1"
    }
  },
  "otel_span_ctx": {
    "span_id": "cd0375b7305d668e",
    "trace_flags": "01",
    "trace_id": "228ad09ba553e60b365a7485f107d9bb"
  },
  "phoenix": {
    "controller": "Elixir.HealthcheckAggregatorWeb.HealthController",
    "action": "index"
  },
  "request_id": "FvLDEr8jXx2K9E8AAAAB",
  "syslog": {
    "hostname": "Caprica",
    "severity": "info",
    "timestamp": "2022-05-26T14:54:58.495Z"
  }
}
```

Datadog has a weird formatting issue between otel and internal datadog ids (https://docs.datadoghq.com/tracing/connect_logs_and_traces/opentelemetry/) but I don't consider that an issue here. It's a Datadog feature / support issue.

Fighting a bit with a more native setup (standard Phoenix project, console logger config, and plugs), and I wasn't actually able to get it to appear. This is from a log function _in the controller_ and `Logger` metadata to `:all`:

```text
erl_level=info application=healthcheck_aggregator domain=elixir file=lib/healthcheck_aggregator_web/controllers/health_controller.ex function=index/2 line=7 mfa=HealthcheckAggregatorWeb.HealthController.index/2 module=HealthcheckAggregatorWeb.HealthController pid=<0.630.0> request_id=FvLDcZ_fnCVyocYAAAAU [info] testing trace data!
```

Tracking it down, seems that it gets ignored because it doesn't have a string implementation: https://github.com/elixir-lang/elixir/blob/main/lib/logger/lib/logger/formatter.ex#L227

ikvmw:
would be great to see it merged. Will it work with opentelemetry_cowboy out of the box provided I will set up logger format accordingly?

tsloughter:
> Tracking it down, seems that it gets ignored because it doesn't have a string implementation: https://github.com/elixir-lang/elixir/blob/main/lib/logger/lib/logger/formatter.ex#L227

Ah, it is because the metadata is nested. Elixir's Logger must support nested metadata and it just need to be patched to work even with `:all`...?

btkostner:
Sweet. Thank you @tsloughter! :confetti_ball:

PR #399, opened 2022-05-24
Document usage of tls_certificate_check in Elixir apps

dvic:


tsloughter:
@dvic but is it before `opentelemetry` in the `.rel` and `.script` files?

dvic:
@tsloughter Indeed, there is a difference in the `.rel` file (the order there is incorrect). 

I can confirm that whatever you put in `extra_applications`, it does not affect the ordering in the `.rel` file in the final release. The non-release .app file of course is affected, and @bryannaegele  indeed it puts the extra applications first (`{applications, [stdlib,elixir,logger,runtime_tools,opentelemetry_exporter, ...`).

I'm going to open an issue to see if this is intended behavior. It seems to me that if it works without releases locally (by putting it inside `extra_applications` a user can expect that it works as well in the mix release.

Thanks again!

tsloughter:
@dvic the only change to `mix release` would be one that works simply by chance and is not something that a user should rely on. That doesn't mean that having it match the shell startup wouldn't be good, where possible, so maybe there is something that can be done in mix -- I'd just suggest it not be documented as a feature a user can rely on :)

dvic:
> @dvic the only change to `mix release` would be one that works simply by chance and is not something that a user should rely on. That doesn't mean that having it match the shell startup wouldn't be good, where possible, so maybe there is something that can be done in mix -- I'd just suggest it not be documented as a feature a user can rely on :)

I agree! But we *could* decide that the order of the dependencies in ` defp deps` is the order in which the apps are started (of course, if you don't set `applications`, which overrides it) and that `extra_applications` is prepended to that list. This is already the behaviour for running locally, the "only" thing that would need to be changed is that this also applies for mix releases. The Elixir instructions for `opentelemetry_exporter` would then be simply that it needs to be placed before `opentelemetry` in the `deps` list.

dvic:
Clear, as https://github.com/elixir-lang/elixir/issues/11872 is closed, shall we just update this pr and reference the existing docs https://github.com/open-telemetry/opentelemetry-erlang#design? We can then update those with the additional instruction to put `opentelemetry_exporter` before `opentelemetry` in the deps so that it works with mix?

tsloughter:
@dvic I forget where we were at with this. I think what you suggest for an update is right though.

dvic:
> @dvic I forget where we were at with this. I think what you suggest for an update is right though.

Done! :)

tsloughter (approved):

PR #404, opened 2022-06-06
feat: add Elixir macro annotation for tracing

btkostner:
This adds an `@trace` annotation for tracing functions. Code was mostly copied from the [New Relic Elixir agent](https://github.com/newrelic/elixir_agent/blob/master/lib/new_relic/tracer/macro.ex) also licensed under Apache License 2.0. I assume this is better code than I can come up with that's already used in production, but I'm willing to rewrite / re architect as needed.

I wasn't sure if / how you wanted tests considering there are none for the Elixir code.

### DIFF
--- a/apps/opentelemetry_api/lib/open_telemetry/macro.ex
+++ b/apps/opentelemetry_api/lib/open_telemetry/macro.ex
@@ -1,0 +1,232 @@
+defmodule OpenTelemetry.Macro do
+  require Logger
+
+  # Function Tracer Macros
+  #   1) __on_definition__: When a function is defined & it's been marked to @trace, we
+  #        store information about the function in module attributes
+  #   2) __before_compile__: We take the list of functions that are marked to @trace and:
+  #        a) Make them overridable
+  #        b) Re-define them with our tracing code wrapped around the original logic
+
+  @moduledoc false
+
+  defguardp is_variable(name, context)
+            when is_atom(name) and (is_atom(context) or is_nil(context))
+
+  # Take no action on the definition of a function head
+  def __on_definition__(_env, _access, _name, _args, _guards, nil), do: nil
+  def __on_definition__(_env, _access, _name, _args, _guards, []), do: nil
+
+  def __on_definition__(%{module: module}, access, name, args, guards, do: body) do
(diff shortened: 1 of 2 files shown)


apps/opentelemetry_api/lib/open_telemetry/macro.ex
```diff
@@ -0,0 +1,232 @@
...
+
+  def traced_function_body(body, module, function, args, trace_info) do
+    {trace_name, trace_args} = build_trace_args(trace_info, module, function, args)
+
+    quote do
+      :otel_tracer.with_span(
+        :opentelemetry.get_application_tracer(__MODULE__),
+        unquote(trace_name),
+        Map.new(unquote(trace_args)),
+        fn _ -> unquote(body) end
```
btkostner, line 169: This is where the code is ran. I'm assuming (but want confirmation) that the `:otel_tracer.with_span` will capture any errors raised and attach them to the span.

bryannaegele:
It's possible some time down the road. Also, I'm not sure if it's going to work with 1.0. Doesn't look like they've kept it up to date, but at first glance I don't think it will break.

tsloughter:
Agh, looks like there are 2 https://hex.pm/packages/opentelemetry_decorator

Maybe we should have 3 :) and @btkostner you could send yours as a new app to https://github.com/open-telemetry/opentelemetry-erlang-contrib and we'll see if we can get a hold of `opentelemetry_decorator` on hex.

connorlay:
I'm glad I found this issue! I've been working on an Elixir DSL for adding traces:

```elixir
def MyModule do

  use MyApp.OpenTelemetry

  trace :my_span_name, foo: "some custom attribute", bar: "another custom attribute"

  def my_traced_function() do
    # ...
  end

  trace :my_other_span_name
  
  defp my_private_traced_function() do
    # ...
  end
```

Why use a macro instead of a module attribute? So we can provide Elixir documentation for the DSL itself, plus macros are more flexible than module attributes when it comes to extending them and customizing their behavior. Currently my WIP DSL supports custom attributes and automatically records exceptions raised as a span error, before reraising to the caller.

I'm happy to share the work if folks are interested 🙂

PR #405, opened 2022-06-08
replace InstrumentationLibrary with InstrumentationScope

tsloughter:
Spec change can be found in PR:
 https://github.com/open-telemetry/opentelemetry-specification/pull/2276


tsloughter:
This is mostly ready but need to find where it isn't backwards compatible and fix those spots.

I think the only place it isn't is that it would be getting rid of the `instrumentation_library` creation functions in `opentelemetry` module. I guess we just add those back and have them return scopes instead.

tsloughter:
Hmm, nevermind, it looks like we only have `get_tracer/3` instead of a `get_tracer` that accepted an InstrumentationLibrary. So no user would be using `opentelemetry:instrumentation_library`. We may not want to bother keeping the function `instrumentation_library` in the module as it could be confusing and shouldn't be being used.

apps/opentelemetry/test/opentelemetry_SUITE.erl
```diff
@@ -139,34 +139,34 @@ end_per_testcase(_, _Config) ->
     ok.
 
 disable_auto_creation(_Config) ->
-    {_, #tracer{instrumentation_library=Library}} = opentelemetry:get_tracer(
+    {_, #tracer{instrumentation_scope=Library}} = opentelemetry:get_tracer(
```
ferd, line 142: These variables (or their equivalent) were renamed in other files. Is it okay to keep the `Library` var to refer to scope in this file?

apps/opentelemetry_api/include/opentelemetry.hrl
```diff
@@ -36,7 +36,7 @@
 
 %% Holds information about the instrumentation library specified when
 %% getting a Tracer from the TracerProvider.
--record(instrumentation_library, {name       :: unicode:unicode_binary() | undefined,
+-record(instrumentation_scope, {name       :: unicode:unicode_binary() | undefined,
```
ferd, line 39: given this is in a public include file, I figure the change isn't backwards compatible; do we figure "oh people shouldn't have used this directly"? If so, I'd correct the comment above to suggest the proper constructor function to use so backwards compat is usable from elsewherer (and to mention scope instead of library as well).

apps/opentelemetry_api/src/opentelemetry.erl
```diff
@@ -86,7 +87,7 @@
 
 -type tracer()             :: {module(), term()}.
 
--type instrumentation_library() :: #instrumentation_library{}.
+-type instrumentation_scope() :: #instrumentation_scope{}.
```
ferd, line 90: is it worth maintaining the type alias as well?

`-type instrumentation_library() :: #instrumentation_scope{}.` or something?

CHANGELOG.md
```diff
@@ -7,6 +7,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+#### Added
+
+- [Instrumentation Scope replaces Instrumentation Library]()
+
+### [Exporter]
+
+#### Added
+
+- [Instrumentation Scope replaces Instrumentation Library]()
```
ferd, line 26: Probably worth adding the "if you were using the record directly, please use the function..." sort of comment there to allow the proper expected scope of backwards compatibility.

apps/opentelemetry/test/opentelemetry_SUITE.erl
tsloughter: I didn't think about it before but actually, yes. All existing usage is still using an instrumentation library name, so it works.

apps/opentelemetry_api/include/opentelemetry.hrl
tsloughter: Ugh, yea, should have made this private but it makes stuff easier in the SDK and Exporter to be public. Need it to be half-private... 

Anyway, yea, I'll fix the comment.

apps/opentelemetry_api/src/opentelemetry.erl
tsloughter: I doubt it. If any user actually uses the library record the least of what they have to deal with is the type spec.

ferd (approved):

PR #407, opened 2022-06-11
Allow custom propagator to be configured via application env

ananthakumaran:


apps/opentelemetry/src/otel_configuration.erl
```diff
@@ -455,9 +455,7 @@ transform(propagator, Value) when Value =:= "b3" ;
 %% transform(propagator, "jaeger") ->
 %%     jaeger;
 transform(propagator, Propagator) ->
```
tsloughter, line 457: Actually, need to add a guard to check that this is an atom. if it is a string then the warning and returning `undefined` should still be the result.

ananthakumaran:
Let me reopen from a new branch, I can't rebase this branch, as other repos are pointing to this commit

PR #408, opened 2022-06-11
Allow custom text propagator to be configured via application env 

### LINKED ISSUE #396: Not able to use custom propagator via config
I am not sure if this is an actual issue or if I am just not calling the correct API. I am planning to use sentry as our trace collector and interop with sentry clients written in other languages. Sentry uses a [sentry-trace](https://develop.sentry.dev/sdk/performance/#header-sentry-trace) HTTP header which is similar to `traceparent`. I figured, I could just write my own propagator and tried to enable it via config below

```
config :opentelemetry,
  text_map_propagators: [:otel_propagator_sentry],
```

But, the custom propagator just gets ignored with a warning

```
[warn] Ignoring unknown propagator otel_propagator_sentry in OS environment variable $OTEL_PROPAGATORS
```

From https://github.com/open-telemetry/opentelemetry-erlang/pull/110, it seems like custom propagator was supported?

ananthakumaran:
fixes #396

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [SDK]
+
+### Fixed
+
+- [Allow custom text propagator to be configured via application env](https://github.com/open-telemetry/opentelemetry-erlang/pull/408)
+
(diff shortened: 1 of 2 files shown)


tsloughter (approved):

tsloughter:
Sorry, one last thing, can you update the changelog file at the top of the repo to include this as a SDK fix.

PR #409, opened 2022-06-25
Add service.instance.id resource default value

fcevado:
based on #402 discussion.


ferd:
The initial thread also had mentions of maybe generating an ID every time the host name is `@localhost` since it does not provide a lot of information. Any views on this?

fcevado:
i totally forgot about that, my bad. I'll add a commit that verifies that.

fcevado:
@ferd I guess everything is covered now. :thinking:

apps/opentelemetry/src/otel_resource_detector.erl
```diff
@@ -161,7 +161,8 @@ required_attributes(Resource) ->
     ProgName = prog_name(),
     ProcessResource = otel_resource:create([{?PROCESS_EXECUTABLE_NAME, ProgName} | process_attributes()]),
     Resource1 = otel_resource:merge(ProcessResource, Resource),
-    add_service_name(Resource1, ProgName).
+    Resource2 = add_service_name(Resource1, ProgName),
+    add_service_instance(Resource2).
```
ferd, line 165: the service.instance.id is a recommended attribute -- only the instance name is required: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md#service

it may make sense to move the recommended attributes to a different function name for the sake of correctness in naming with regards to the spec.

tsloughter: Yea, can add an additional call to some, "recommended attributes" function that right now only adds service instance.

fcevado: I'm wondering if it wouldn't be good to rename this function to `default_runtime_attributes`, since those values are being assumed based on runtime properties, given that it wasn't previously assigned. otherwise, I think it would make sense to put the `service.instance.id` attribution together with the `telemetry.sdk` values, since they're other recommended values.

tsloughter: I'd be fine with renaming to `default_resource_attributes` -- I think `runtime` is too overloaded a word.

apps/opentelemetry/test/otel_resource_SUITE.erl
```diff
@@ -202,6 +202,102 @@ release_service_name(_Config) ->
         os:unsetenv("RELEASE_NAME")
     end.
 
+service_instance_id_env(_Config) ->
```
tsloughter, line 197: The test function names need to be in th elist returned by `all() ->` or they aren't run.

tsloughter:
The tests don't actually run, but I pulled this down locally and ran them and they work. So I'm going to merge this and then open a PR to make the tests run.

tsloughter:
Sorry for the long delay!

PR #410, opened 2022-07-01
add record_exception/1 to OpenTelemetry.Tracer

btkostner:
This adds `record_exception/1` to `OpenTelemetry.Tracer`. This is a shortcut for calling `OpenTelemetry.Span.record_exception/1` with the current span context.


tsloughter:
Thanks! Can you add the same in `otel_tracer.hrl`?

btkostner:
@tsloughter My erlang isn't great, but I can try. I also noticed functions like `set_attribute` and `add_event` are missing from `otel_tracer.erl`. Should I add those too?

tsloughter:
@btkostner great, thanks.

They shouldn't be added to `otel_tracer.erl` but `otel_tracer.hrl` and in there you'll see `set_attribute` and `add_event` already there. Since in Elixir you can put macros in the module we add them to the Tracer module in Elixir but in Erlang macros go in a header file, so `otel_tracer.hrl`.

btkostner:
Should be good now :+1:

tsloughter (approved):
Thanks, sorry for the delay!

PR #412, opened 2022-07-19
Stable metrics sdk

tsloughter:
Opening as a draft because I'm not 100% on the direction of this but I think it is close enough to start getting feedback.

There will be a number of tasks after this PR is merged, like supporting floats in the aggregations, filtering attributes in views and adding the async/observable metrics, in addition to many clean up tasks.

Important things to review in this PR are design decisions. One being that each measurement is a message. I do not like, and don't think we can keep, the message going to the `otel_meter_server` because that also handles other messages and should remain responsive for those. I will likely create a process, with the possibility in the future to be a pool of processes assuming aggregations remain safe to act concurrently.

The Reader process isn't properly handled yet either because it is monitored by the otel_meter_server and not linked but the ets tables it uses are owned by the server. Either the reader needs to be able to run on its own or it should be linked. I'd prefer the former.

Also getting an Instruments no longer requires a call to the server process. Initially I thought it best if it did so that it could be matched to views at that point and not on each measurement but this is likely not useful, and certainly not useful compared to not having to make a call to get an Instrument. Note that future view matching could include attributes in the criteria, so then setting up matches based only on the Instrument isn't even an option.

An issue is that views can be added at any time and after an instrument is matched once (also an issue if we were to support criteria that matches on attributes) it won't be matched again, meaning views added after the first time an instrument is used will never be matched.

Note that filtering attributes is different from using them as criteria. Filtering still needs to be added but needs to be there before this moves out of experimental.

Not sure if the matchspec in `checkpoint` is efficient but I think checkpoint should be working against all metrics at once instead of one at a time, so it likely will be removed anyway.


tsloughter:
Eh, marking as ready for review. I don't think anything major needs to be changed before merging since it is all in experimental. So I'd like to get reviews and merge this so I can be doing smaller future PRs that are easier to review.

apps/opentelemetry_api_experimental/src/otel_counter.erl
```diff
@@ -12,51 +12,25 @@
...
+add(Instrument=#instrument{module=Module,
+                           value_type=?VALUE_TYPE_INTEGER}, Number, Attributes)
+  when is_integer(Number) andalso Number >= 0 ->
+    Module:record(Instrument, Number, Attributes);
+add(Instrument=#instrument{module=Module,
+                           value_type=?VALUE_TYPE_FLOAT}, Number, Attributes)
+  when is_float(Number) andalso Number >= 0 ->
+    Module:record(Instrument, Number, Attributes);
+add(_, _, _) ->
+    %% TODO: add debug, warning or info log here?
```
ferd, line 35: I'd say we pick the lowest level if unsure, but at least have one.

apps/opentelemetry_api_experimental/src/otel_histogram.erl
```diff
@@ -0,0 +1,36 @@
...
+-spec record(otel_instrument:t(), number(), opentelemetry:attributes_map()) -> ok.
+record(Instrument=#instrument{module=Module,
+                              value_type=?VALUE_TYPE_INTEGER}, Number, Attributes)
+  when is_integer(Number) ->
+    Module:record(Instrument, Number, Attributes);
+record(Instrument=#instrument{module=Module,
+                              value_type=?VALUE_TYPE_FLOAT}, Number, Attributes)
+  when is_float(Number) ->
+    Module:record(Instrument, Number, Attributes);
+record(_, _, _) ->
```
ferd, line 35: implicit for rest of reviews: replicate log decision in all matching modules.

apps/opentelemetry_api_experimental/include/otel_metrics.hrl
```diff
@@ -0,0 +1,18 @@
+-record(instrument, {module      :: module(),
+                     meter       :: otel_meter:t(),
+                     name        :: otel_instrument:name(),
+                     description :: otel_instrument:description() | undefined,
+                     kind        :: otel_instrument:kind(),
+                     value_type  :: otel_instrument:value_type(),
+                     unit        :: otel_instrument:unit() | undefined,
+                     callback    :: otel_instrument:callback() | undefined}).
```
ferd, line 8: Is this record meant to be publicly available outside the app? It is given its own type in `otel_instrument` so I'm assuming the record ought to be private (eg `src/otel_metrics_rec.hrl` which could include this file to keep the macro definitions public)

apps/opentelemetry_api_experimental/src/otel_meter_provider.erl
```diff
@@ -12,73 +12,67 @@
...
+%% stores Meter configuration and is how Meters are accessed. An
+%% implementation must be a `gen_server' that handles the API's calls. The
+%% SDK should register a MeterProvider with the name `otel_meter_provider'
+%% which is used as the default global Provider.
 %% @end
 %%%-------------------------------------------------------------------------
 -module(otel_meter_provider).
 
--behaviour(gen_server).
+-export([get_meter/3,
```
ferd, line 24: do you want to export get_meter/4 here too?

apps/opentelemetry_api_experimental/src/otel_observable_counter.erl
```diff
@@ -12,15 +12,10 @@
...
--module(otel_metric_exporter_stdout).
-
--export([export/1]).
+-module(otel_observable_counter).
 
-export(Records) when map_size(Records) > 0 ->
-    io:format("*METRICS FOR DEBUG*~nRecords ~p~n", [Records]);
-export(_) ->
-    ok.
+-export([]).
```
ferd, line 21: is it expected that the 3 observable counters get an add() function at some point?

ferd (commented):
Started with a review of the API, I'll try to do the internals a bit later this week.

apps/opentelemetry_api_experimental/src/otel_counter.erl
tsloughter: Yea, I'll go with debug for now.

apps/opentelemetry_api_experimental/src/otel_observable_counter.erl
tsloughter: Right. In this PR I have yet to implement observables at all.

apps/opentelemetry_api_experimental/src/otel_meter_provider.erl
tsloughter: Ah, yup.

apps/opentelemetry_api_experimental/include/otel_metrics.hrl
tsloughter: Issue is it is needed in the SDK app. So I'm not sure what is best to do since I'd rather not go the route of putting it all by a module for accessing each field.

ferd: Oh well we could cheat a bit I guess. Have the content is wrapped in a sort of:

```erlang
-ifdef(OTEL_INTERNAL_MACRO).
-record(instrument,  ...)
-endif.
```

and then you could include the same file but running a `-define(OTEL_INTERNAL_MACRO, true)` prior to inclusion gives you visibility.

ferd: I don't know if that's overkill though.

apps/opentelemetry_experimental/src/otel_aggregation_sum.erl
```diff
@@ -0,0 +1,95 @@
...
+            _ = ets:update_counter(Tab, Key, {#sum_aggregation.value, Value},
+                                   init(Key, [])),
+            true
+    end.
+
+-dialyzer({nowarn_function, checkpoint/5}).
+checkpoint(Tab, Name, ?AGGREGATION_TEMPORALITY_DELTA, ?VALUE_TYPE_INTEGER, CollectionStartNano) ->
+    MS = [{#sum_aggregation{key='$1',
+                            value='$2',
+                            _='_'},
```
ferd, line 57: I would avoid a wildcard match here; if a field is ever added to the record, this will silently succeed matching but the re-insertion will reset the fields.

By forcing an explicit key match for the patterns, we'll catch changes in tests by having any match failing. OTOH it will be costlier to match though so maybe you're fine with this. If so, maybe pair the record definition with a comment about needing to update the aggregators to care for the fields?

Feel free to disregard this if we expect the spec to be stable.

apps/opentelemetry_experimental/src/otel_meter_server.erl
```diff
@@ -12,42 +12,281 @@
...
+
+start_reader(ReaderModule, ReaderConfig) ->
+    {ViewAggregationTab, MetricsTab} = create_tables(),
+
+    %% TODO: using a monitor bc the original idea was readers could continue operating without
+    %% the meter server this isn't the case right now since the ets tables are owned by the meter server
+    %% unless this is changed this needs to be switched to a link
+    %% also need to be able to recover who the readers are when the meter server restarts
+    %% discovery by looking at a supervision tree could be a good idea as it would allow for added
+    %% readers to be returned to the server after a crash
```
ferd, line 239: supervision tree would work though it can fail to restore fine if the server's failure lines up with a reader's failure; another idea is to have a parent-owned table recording which readers should exist, so in case of a restart the new server can restore state to where it should be?

apps/opentelemetry_experimental/src/otel_metric_exporter_pid.erl
```diff
@@ -13,18 +13,33 @@
...
+-include("otel_metrics.hrl").
 
--callback update(ets:tab(), otel_meter:name() | {otel_meter:name(), otel_meter:labels()}, otel_meter:number_kind(), number()) -> boolean().
+init(Pid) ->
+    {ok, Pid}.
 
--callback checkpoint(ets:tab(), otel_meter:name() | {otel_meter:name(), otel_meter:labels()}) -> boolean().
+export(Metrics, Pid) ->
+    lists:map(fun(Metric) ->
+                  Pid ! Metric
```
ferd, line 35: should this message be wrapped in a customizeable tagged tuple to make it easier to match for a consumer? Would let someone set specific exporters for each reader and know which reader is returning what even if they go to a centralized process?

apps/opentelemetry_experimental/src/otel_metric_reader.erl
```diff
@@ -0,0 +1,187 @@
...
+handle_call(_, _From, State) ->
+    {noreply, State}.
+
+handle_cast(_, State) ->
+    {noreply, State}.
+
+handle_info(collect, State=#state{exporter=undefined,
+                                  export_interval_ms=ExporterIntervalMs,
+                                  tref=TRef}) ->
+    erlang:cancel_timer(TRef, []), %% {async, true} ?
```
ferd, line 100: probably wouldn't matter. Since the interval won't change, at most you wait `2*Interval` (if you were 1ms away from sending, then cancel and go) but if the race happens where you cancel but the event has already fired, you'll wait `Interval` milliseconds. Async is unlikely to hurt, may be a bit faster, but won't be a huge block either way I think.

apps/opentelemetry_experimental/src/otel_metric_reader.erl
```diff
@@ -0,0 +1,187 @@
...
+    {ok, State}.
+
+%%
+
+collect_(ViewAggregationTab, MetricsTab) ->
+    CollectionStartTime = erlang:system_time(nanosecond),
+
+    %% Need to be able to efficiently get all from VIEW_AGGREGATIONS_TAB that apply to this reader
+
+    %% for each VIEW_AGGGREGATIONS_TAB entry look up metrics from METRICS_TAB using the name
```
ferd, line 139:
```suggestion
    %% for each VIEW_AGGREGATIONS_TAB entry look up metrics from METRICS_TAB using the name
```

apps/opentelemetry_experimental/src/otel_view.erl
```diff
@@ -0,0 +1,122 @@
...
+                                        description=ViewDescription,
+                                        aggregation_options=AggregationOptions,
+                                        selection=#selection{instrument_name=InstrumentName}})
+                               when Name =:= InstrumentName ->
+                                 {true, {View, #view_aggregation{name=case ViewName of
+                                                                          undefined ->
+                                                                              InstrumentName;
+                                                                          _ ->
+                                                                              ViewName
+                                                                      end,
```
ferd, line 83: both cases may be more readable with a call like:

```erlang
val_default(Val, Default) ->
    case Val of
        undefined -> Default;
        _ -> Val
    end.
```

apps/opentelemetry_experimental/src/otel_metrics.hrl
```diff
@@ -0,0 +1,116 @@
...
+
+-record(histogram_datapoint,
+        {
+         attributes :: opentelemetry:attributes_map(),
+         start_time_unix_nano :: integer() | undefined,
+         time_unix_nano :: integer(),
+         count :: integer(),
+         sum :: float(),
+         bucket_counts :: tuple(),
+         explicit_bounds :: list(), %% list of floats
```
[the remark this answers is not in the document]
tsloughter: Hm, yea, can't recall why I would have done it this way.

apps/opentelemetry_experimental/src/otel_aggregation_histogram_explicit.erl
```diff
@@ -0,0 +1,159 @@
...
+init(Key, Options) ->
+    Boundaries = maps:get(boundaries, Options, ?DEFAULT_BOUNDARIES),
+    RecordMinMax = maps:get(record_min_max, Options, true),
+    #explicit_histogram_aggregation{key=Key,
+                                    start_time_unix_nano=erlang:system_time(nanosecond),
+                                    boundaries=Boundaries,
+                                    bucket_counts=zero_buckets(length(Boundaries)),
+                                    record_min_max=RecordMinMax,
+                                    min=infinity, %% works because any atom is > any integer
+                                    max=neg_infinity, %% requires an explicit check
```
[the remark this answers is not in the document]
tsloughter: Right, these are the defaults until any data is encountered.

Going with `undefined` probably makes the most sense. I had hopes that there would be something I'd find that could be used for `max` that would always be less than any number, but since there isn't... wait, just realized that even though Erlang has no number size limit the Otel protocol does, so technically we could use `-2^64` here. hmm

The reason I wanted a term that I could compare with `<` is so it could be done in a matchspec -- and it means only one comparison needed, not one to check if its `undefined` and then to check if its `<`.

apps/opentelemetry_experimental/src/otel_aggregation_histogram_explicit.erl
```diff
@@ -0,0 +1,159 @@
...
+                                    sum=0
+                                   }.
+
+aggregate(Table, Key, Value) ->
+    case ets:lookup(Table, Key) of
+        [Current] ->
+            %% TODO: needs to be changed to work with concurrent updates
+            ets:insert(Table, aggregate(Value, Current));
```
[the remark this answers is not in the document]
tsloughter: Good point.

Note it is still "right" here because all measurements go through the meter server process, but plan is to not be limited to one process. Either way moving to a compare and swap function that every module could use is useful.

apps/opentelemetry_experimental/src/otel_aggregation_histogram_explicit.erl
```diff
@@ -0,0 +1,159 @@
...
+                                                      max=Max,
+                                                      sum=Sum}) ->
+    Buckets1 = bump_bucket_counts(MeasurementValue, Boundaries, Buckets),
+    Aggregation#explicit_histogram_aggregation{bucket_counts=Buckets1,
+                                               min=min(Min, MeasurementValue),
+                                               max=case Max of
+                                                       neg_infinity -> MeasurementValue;
+                                                       _ -> max(Max, MeasurementValue)
+                                                   end,
+                                               sum=Sum+MeasurementValue};
```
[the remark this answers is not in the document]
tsloughter: Oh, interesting idea. For this PR I think I'll write that in as a `TODO` to investigate it and possibly go with that.

Issue around data race with min/max actually applies to updating the buckets as well during a checkpoint. Don't want to checkpoint the values after a bucket is updated but before min/max are updated. But maybe they can all be updated together because `update_counter` takes a list.. Argh, but that doesn't work for floats, so not an option.

apps/opentelemetry_experimental/src/otel_aggregation_sum.erl
tsloughter: Yea, can't remember why I did this, I think it was just to guard against it not checkpointing if the record was ever changed. Sure it would reset the other fields but at least it would checkpoint the key and value. Not sure that makes sense anymore, so I'll probably change it in this PR.

apps/opentelemetry_experimental/src/otel_metric_exporter_pid.erl
tsloughter: Good point, I'll add that.

apps/opentelemetry_experimental/src/otel_metric_reader.erl
tsloughter: Yea, I added that because I hadn't taken the time to think if it could potentially hurt. If it can't hurt then I'll use it even if it won't really help anything.

apps/opentelemetry_experimental/src/otel_view.erl
tsloughter: Yup, thanks

apps/opentelemetry_experimental/src/otel_aggregation_sum.erl
tsloughter: Oh na, now I see, I did that since I did consider the record stable. But I think being explicit will make it less likely to cause confusion at least in the future.

apps/opentelemetry_api_experimental/include/otel_metrics.hrl
tsloughter: Yea, might be overkill. But might be worth it if we don't think of anything else.

For now though since this is in experimental I think we should move forward as is and can freely make the record private after this as long as its before it goes into the stable SDK.

apps/opentelemetry_experimental/src/otel_aggregation_histogram_explicit.erl
```diff
@@ -0,0 +1,161 @@
...
+init(Key, Options) ->
+    Boundaries = maps:get(boundaries, Options, ?DEFAULT_BOUNDARIES),
+    RecordMinMax = maps:get(record_min_max, Options, true),
+    #explicit_histogram_aggregation{key=Key,
+                                    start_time_unix_nano=erlang:system_time(nanosecond),
+                                    boundaries=Boundaries,
+                                    bucket_counts=zero_buckets(length(Boundaries)),
+                                    record_min_max=RecordMinMax,
+                                    min=infinity, %% works because any atom is > any integer
+                                    max=?MIN_DOUBLE,
```
ferd, line 44: The fact that the data model marks both the min and max value as optional does seem to mean we have to be careful not to use the MIN value unless you want it to be impossible to distinguish between a legit MIN and a placeholder (https://github.com/open-telemetry/opentelemetry-proto/blob/c5c8b28012583fda55b0cb16f73a820722171d49/opentelemetry/proto/metrics/v1/metrics.proto#L496-L501)

Is this logic referred to the collector or the exporter at this point?

apps/opentelemetry_experimental/src/otel_metrics.hrl
```diff
@@ -0,0 +1,116 @@
...
+         key :: {term(),  opentelemetry:attributes_map()},
+         start_time_unix_nano :: integer(),
+         %% instrument_temporality :: otel_aggregation:temporality(),
+         %% default: [0.0, 5.0, 10.0, 25.0, 50.0, 75.0, 100.0, 250.0, 500.0, 1000.0]
+         boundaries :: [float()],
+         record_min_max :: boolean(),
+         checkpoint :: #explicit_histogram_checkpoint{} | undefined | '_',
+         bucket_counts :: tuple(),
+         min :: number() | neg_infinity | infinity,
+         max :: number() | neg_infinity | infinity,
```
ferd, line 63: we no longer use neg_infinity here

apps/opentelemetry_experimental/src/otel_aggregation_histogram_explicit.erl
tsloughter: We don't need to distinguish because this record is only created when a measurement is taken. If minmax are to be recorded it will be set at that time.

Which reminds me, `explicit_histogram_checkpoint` should only be `number()`.

The only time this may be an issue is a delta temporality, which will reset to the default when taking the checkpoint. That isn't supported yet.

apps/opentelemetry_experimental/src/otel_metrics.hrl
tsloughter: updated. Same for the checkpoint record that doesn't include `neg_infinity` or `infinity`.

ferd (commented):
Approved on my end, though I'd like to see what @GregMefford has to say about it

ferd (approved):

tsloughter:
Going to go ahead and merge. But it would be great if @GregMefford gets some time.

If @GregMefford and/or @bryannaegele are able to make the SIG meeting tomorrow we could also do a code walk through at that time.

Future PRs will be smaller.

PR #413, opened 2022-07-21
drop spans immediately if no exporter is configured

tsloughter:
This change tries to balance not losing spans while an exporter is still being initialized and not bothering to insert spans into the export table if the user has configured the exporter to be none.

### DIFF
--- a/apps/opentelemetry/src/otel_batch_processor.erl
+++ b/apps/opentelemetry/src/otel_batch_processor.erl
@@ -114,10 +114,16 @@ init([Args]) ->
     _Tid2 = new_export_table(?TABLE_2),
     persistent_term:put(?CURRENT_TABLES_KEY, ?TABLE_1),
 
-    enable(),
+    %% only enable export table if there is going to be an exporter
+    case maps:get(exporter, Args, none) of
+        ExporterConfig when ExporterConfig =:= none ; ExporterConfig =:= undefined ->
+            disable();
+        ExporterConfig ->
+            enable()
+    end,
 
     {ok, idle, #data{exporter=undefined,
-                     exporter_config=maps:get(exporter, Args, none),
+                     exporter_config=ExporterConfig,
                      resource = Resource,
                      handed_off_table=undefined,
                      max_queue_size=case SizeLimit of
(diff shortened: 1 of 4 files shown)


tsloughter:
going to add a test that verifies those tables actually get nothing in them.

patmaddox:
Looks good over here.

[Example app](https://github.com/patmaddox/otel_mem_test)

## Before PR

https://github.com/patmaddox/otel_mem_test/commit/172da08f07179dad2d4dd549842b1564fe8b3d5f

image

## After PR

https://github.com/patmaddox/otel_mem_test/commit/246c21122e0ea0ae93d2581d4cc3164ca1ebe8f6

image

apps/opentelemetry/src/otel_batch_processor.erl
```diff
@@ -230,6 +253,23 @@ terminate(_, _, _Data) ->
...
+            %% exporter is undefined/none
+            %% disable the insertion of new spans and delete the current table
+            clear_table_and_disable(),
+            undefined
+    end.
+
+clear_table_and_disable() ->
+    disable(),
+    ets:delete(?CURRENT_TABLE),
+    new_export_table(?CURRENT_TABLE).
```
ferd, line 271: there's a potential race here since there's no locking where pid A checks, sees it's enabled, then pid B disables, deletes, recreates the table, then A writes to it. Probabilities are low.

But there's also a risk of something trying to write at the table between the delete and new one, which is likely to cause failure. Just emptying the table is likely to lower the chances of surprises without fully removing the risk of a race insertion.

tsloughter: The insert `do_insert/1` is already wrapped in a `try/catch` so I wasn't worried about those failings -- originally I was just going to do a `delete_all_objects`.

But maybe we should change to deleting objects instead of the table to get rid of the race that also exists when the table is deleted after export. That one is more unlikely since it would mean the `end_span` got the table atom name from the persistent term, then the table was swapped and passed to the exporter and deleted before it does the insert.

I just felt safer garbage collection wise with tables being dropped... at least I don't think there is any potential performance issue as long as a full `delete_all_objects` is done?

tsloughter: Now that I think about `do_insert` again I realize we likely don't need the `try/catch`! I stupidly made the `is_enabled` return `true` by default.. which meant it could fail even if the batch processor process was unable to start, so wrappedin the `try/catch` in case no table was created. But if `is_enabled` defaults to `false` and ensure that if the process crashes it gets set back to `false` (or the tables are given to an heir to stay available) then the try/catch would not be needed.

tsloughter: Well now the question of should it be acceptable to potentially lose a span if it gets written after export has completed (because the process get the table to write to before it was handled off to the exporter) but before `delete_all_objects`, or should we not pass the ETS table to the exporter but all spans taken and deleted from the table.

Ugh, there is no `select_delete` that returns the selected elements.. so there may not be a way to even do this without doing like `tab2list` followed by `ets:delete` on each element of that list.

There has got to be a better way to not leave any potential to lose a span.

tsloughter: @ferd I think I'll do the `delete_all_objects` and not create a new table each time but want to do it in another PR.

ferd: Sounds good. This does not seem like it's adding any inconsistency that wasn't there before anyway.

ferd (approved):

PR #414, opened 2022-07-23
remove resource field from tracer record

tsloughter:
The Resource is created on startup and the same for the entire
node. There is no need to carry it in the tracer record since
the processor that exports spans gets the Resource from the
tracer provider for attaching to the spans it exports.


tsloughter:
Saves a bit of memory and copying, esp if a node has a lot of resource attributes.

ferd (approved):
Are you aware of any other exporter from the community that could have relied on this current form rather than just the internal ones (which do get it from the tracer)?

tsloughter:
@ferd na. The Tracers aren't passed to the exporter anyway and the Resource is passed as an explicit argument to `export`.

PR #415, opened 2022-07-23
Add traces_exporter: :none to documentation

sax:
This came up for me recently and I was only able to discover this via (admittedly very helpful) messages in the Elixir Slack. I thought that it would be helpful for others to document explicitly.

Let me know if any changes would be preferred, or I don't mind others fixing up directly.

### DIFF
--- a/apps/opentelemetry/README.md
+++ b/apps/opentelemetry/README.md
@@ -19,7 +19,7 @@ span processor which then exports through the [OpenTelemetry
 Protocol](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.8.0/specification/protocol/otlp.md)
 over HTTP to `http://localhost:4318`, encoding the Spans with protobufs.
 
-``` erlang
+```erlang
 [
  {opentelemetry,
   [{span_processor, batch},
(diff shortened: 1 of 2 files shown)


apps/opentelemetry/README.md
```diff
@@ -41,6 +41,20 @@ config :opentelemetry_exporter,
   otlp_endpoint: "http://localhost:4318"
 ```
 
+To disable opentelemetry, the `traces_exporter` can be set to the atom `none`.
+This will drop spans as they are collected to ensure memory does not overflow.
```
ferd, line 45: I think this may not be the most helpful line. Is memory overflowing because you are generating more data than can be exported given the current system? If so, would the proper approach be to improve the system's throughput? Or is it overflowing because you don't have a working point to export to and you want to disable opentelemetry exporting until everything is in place?

What I mean here is that "stop exporting" is _one way_ to prevent memory issues, but it also does so at the cost of not exporting any trace data, which arguably may defeat the purpose of the library. A statement like:

> If your exporting pipeline is not ready, you may disable trace exporting to prevent memory overflows. This can be done by setting `traces_exporter`'s value to the atom `none`.

Would likely be more descriptive of when to disable it by contextualizing the memory overflows.

sax: Sounds good to me. This is exactly the use case that I'm in the middle of: we have added tracing throughout our application, but are unable to actually export traces until agreements are signed regarding moving data between environments.

ferd (approved):

PR #420, opened 2022-08-10
Fix OTLP Protocol allowed values

andrewhr:
As the [specification describes][1], there are 3 possible values for ENV vars that set the OTLP protocol used by OTLP exporter:
- grpc
- http/protobuf
- http/json

The SDK implementation leans on `list_to_existing_atom` function, so our expected values are in fact `http_protobuf` instead of `http/protobuf` for example.

This is confusing for ones following the generic documentation, and also may become a problem in polyglot environments: let's say, I share the same ENV setup for a Java and Erlang application.

The fix here account for the current atom values, and strings with the same snake_case format for backwards compatibility. But also accept the standard approach.

The fix includes `http/json` as possible transform, but the exporter will still fails as that isn't supported by our impl.

[1]: https://github.com/open-telemetry/opentelemetry-specification/blob/82707fd9f7b1266f1246b02ff3e00bebdee6b538/specification/protocol/exporter.md#specify-protocol

### DIFF
--- a/apps/opentelemetry/src/otel_configuration.erl
+++ b/apps/opentelemetry/src/otel_configuration.erl
@@ -358,6 +358,17 @@ transform(exporter, UnknownExporter) when is_list(UnknownExporter) ->
     ?LOG_WARNING("unknown exporter ~p. falling back to default otlp", [UnknownExporter]),
     {opentelemetry_exporter, #{}};
 
+transform(otlp_protocol, Proto) when Proto =:= "grpc"; Proto =:= grpc ->
+  grpc;
+transform(otlp_protocol, Proto) when Proto =:= "http/protobuf"; Proto =:= "http_protobuf"; Proto =:= http_protobuf ->
+  http_protobuf;
+transform(otlp_protocol, Proto) when Proto =:= "http/json"; Proto =:= "http_json"; Proto =:= http_json ->
+  ?LOG_WARNING("The http/json transport protocol is not supported by the OTLP exporter, spans will not be exported.", []),
+  undefined;
+transform(otlp_protocol, Proto) ->
+  ?LOG_WARNING("unknown ~ts transport protocol, spans will not be exported.", [Proto]),
+  undefined;
+
 transform(integer_infinity, infinity) ->
     infinity;
 transform(integer_infinity, Value) ->
--- a/apps/opentelemetry_exporter/test/opentelemetry_exporter_SUITE.erl
+++ b/apps/opentelemetry_exporter/test/opentelemetry_exporter_SUITE.erl
@@ -180,6 +180,26 @@ configuration(_Config) ->
                        ssl_options => undefined},
                      opentelemetry_exporter:merge_with_environment(#{endpoints => []})),
 
+        %% test all supported protocols
+        application:unset_env(opentelemetry_exporter, otlp_protocol),
+        os:putenv("OTEL_EXPORTER_OTLP_PROTOCOL", "grpc"),
+        ?assertMatch(#{protocol := grpc},
+                     opentelemetry_exporter:merge_with_environment(#{})),
+
+        %% the specification defines the value as "http/protobuf"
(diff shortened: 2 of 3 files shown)


tsloughter (approved):
Thanks! I could have swore I had already done this!

apps/opentelemetry/src/otel_configuration.erl
```diff
@@ -355,6 +355,13 @@ transform(exporter, UnknownExporter) when is_list(UnknownExporter) ->
     ?LOG_WARNING("unknown exporter ~p. falling back to default otlp", [UnknownExporter]),
     {opentelemetry_exporter, #{}};
 
+transform(otlp_protocol, Proto) when Proto =:= "grpc"; Proto =:= grpc ->
+  grpc;
+transform(otlp_protocol, Proto) when Proto =:= "http/protobuf"; Proto =:= "http_protobuf"; Proto =:= http_protobuf ->
+  http_protobuf;
+transform(otlp_protocol, Proto) when Proto =:= "http/json"; Proto =:= "http_json"; Proto =:= http_json ->
+  http_json;
```
tsloughter, line 363: Oops, approved too soon. We need a log here and I'd propose falling back to `http_protobuf`.

I know Ruby doesn't support grpc, so could see what they do if the protocol is set to grpc and do the same.

tsloughter (changes requested):
Changing to requesting changes.

apps/opentelemetry/src/otel_configuration.erl
andrewhr: [The Agent defalts to none and logs a warning](https://github.com/open-telemetry/opentelemetry-ruby/blob/5b0cee9678badba6e332b380d532d5f4dcd1714a/sdk/lib/opentelemetry/sdk/configurator.rb#L181-L188)

I replicated their behaviour here. I believe the rationale is something in the lines "if I expect something different from HTTP/Protobuf, it will fail anyway". Seems a sensible approach to me, what do you think Tristan?

tsloughter: The warning looks good, but we need a catch all for cases the string is not one of the values we expect. I think `undefined` should be used instead of `none`. We use `none` for setting no exporter, so prefer to just use `undefined` for a wrong protocol. 

Also, `opentelemetry_exporter:init/1` will crash in the case on line 165, it needs to return I guess `{ok, #state{protocol=undefined}}` when protocol is `undefined`.

tsloughter:
@andrewhr there is a test failure. Looks like it just needs `assertEqual` changed to `assertMatch`.

andrewhr:
@tsloughter sorry for the delay. I'm not that familiar with Erlang testing and have some issues running locally. So I just changed to `assertMatch` as you suggested 🤞

tsloughter (approved):

apps/opentelemetry_exporter/test/opentelemetry_exporter_SUITE.erl
```diff
@@ -155,6 +155,31 @@ configuration(_Config) ->
...
+                       headers => [],
+                       protocol => http_protobuf},
+                     opentelemetry_exporter:merge_with_environment(#{endpoints => []})),
+
+        %% the specification defines the value as "http/protobuf"
+        %% https://github.com/open-telemetry/opentelemetry-specification/blob/82707fd9f7b1266f1246b02ff3e00bebdee6b538/specification/protocol/exporter.md#specify-protocol
+        application:unset_env(opentelemetry_exporter, otlp_protocol),
+        os:putenv("OTEL_EXPORTER_OTLP_ENDPOINT", "http://localhost:4318"),
+        os:putenv("OTEL_EXPORTER_OTLP_PROTOCOL", "http/protobuf"),
+        ?assertMatch(#{endpoints =>
```
tsloughter, line 175: Match needs to use `:=`

Erlang distinguishes matching `:=` and creating `=>` in maps.

andrewhr:
Thanks for explaining me how `assertMatch` works in Erlang. I rewrote the tests to just check the protocol now, via match. I get to run locally by mimic of Github commands. I think we're good!

Sorry for the noise in the meantime 🙇

PR #431, opened 2022-08-17
move readers under metrics supervisor and require registered reader

tsloughter:
This patch moves readers under a readers supervisor instead of being
started from the metrics server. Readers must be declared in config
and provided with a unique atom name that is used for registration
of the process. This name is also used to create named ets tables
to store the view aggregation mappings and the metrics.

Use of named ets tables and registered processes means the pid and
tids don't need to be tracked but a future update may find it
better to remove the need for names.

### DIFF
--- a/apps/opentelemetry/src/opentelemetry_sup.erl
+++ b/apps/opentelemetry/src/opentelemetry_sup.erl
@@ -69,7 +69,7 @@ init([Opts]) ->
                 start => {otel_span_sup, start_link, [Opts]},
                 type => supervisor,
                 restart => permanent,
-                shutdown => 5000,
+                shutdown => infinity,
                 modules => [otel_span_sup]},
 
     %% `TracerServer' *must* start before the `BatchProcessor'
--- a/apps/opentelemetry_experimental/test/otel_test_utils.hrl
+++ b/apps/opentelemetry_experimental/test/otel_test_utils.hrl
@@ -2,10 +2,14 @@
 -define(UNTIL(X), (fun Until(I) when I =:= 10 ->
                            ct:fail("timeout: UNTIL(~s)", [??X]);
                        Until(I) ->
-                           case X of
+                           try X of
                                true ->
                                    ok;
                                false ->
+                                   timer:sleep(100),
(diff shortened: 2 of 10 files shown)


tsloughter:
Still not happy with it, but good enough to move forward with I think.

apps/opentelemetry/src/opentelemetry_sup.erl
```diff
@@ -69,7 +69,7 @@ init([Opts]) ->
                 start => {otel_span_sup, start_link, [Opts]},
                 type => supervisor,
                 restart => permanent,
-                shutdown => 5000,
+                shutdown => infinity,
```
ferd, line 72: Initially did a double-take about `infinity` but I realize the lower-level supervisors have a timeout, so transitively this is making sure the children supervisors have their time to shut down cleanly while timeouts still exist within the system.

apps/opentelemetry_experimental/test/otel_test_utils.hrl
```diff
@@ -2,12 +2,16 @@
...
-                           case X of
+                           try X of
                                true ->
                                    ok;
                                false ->
                                    timer:sleep(100),
                                    Until(I+1)
+                           catch
+                               _:_ ->
+                                   timer:sleep(100),
```
ferd, line 13: should we ct:pal this so there's at least a trace of failures in logs if the case is unsuccessful?

ferd (approved):

apps/opentelemetry_experimental/test/otel_test_utils.hrl
tsloughter: Ah, good point. I added it for the specific case where I expect an exception while the process is coming up, but will definitely be confusing later on when something else is throwing an exception.

apps/opentelemetry/src/opentelemetry_sup.erl
tsloughter: Only the workers at the lower level should be non-infinity, you said "lower-level" supervisors, did you find some that aren't infinity?

PR #432, opened 2022-08-17
make resource instance name tests run

tsloughter:


ferd (commented):
I'd approve adding the tests, but lots of failures happening apparently? Before seeing what the fixes require I guess I'll withhold approval?

tsloughter:
Argh, yea, I ran them locally so thought they'd pass (at least on the same OTP, 25.0.3!). Maybe it is a difference in the environment since this is using the node name for stuff.

tsloughter:
@ferd I just have it now skipping the 2 tests that rely on setting the node name with `net_kernel` which seems to not work in Github Actions.

ferd (approved):

PR #434, opened 2022-08-18
only run tls_certificate_check if ssl options are needed

tsloughter:


apps/opentelemetry_exporter/src/opentelemetry_exporter.erl
```diff
@@ -323,7 +323,16 @@ parse_endpoint(Endpoint=#{host := Host, scheme := Scheme, path := Path}, Default
...
     %% found in `Endpoint'
     {true, maps:merge(#{port => scheme_port(Scheme),
-                        ssl_options => update_ssl_opts(HostString, DefaultSSLOpts)},
+                        %% we only want to run `tls_certificate_check' if absolutely
+                        %% necessary, so wrapping the setup of ssl options here
+                        %% in a check that it won't just be overwritten by the value
+                        %% from the Endpoint anyway
+                        ssl_options => case maps:is_key(ssl_options, Endpoint) of
+                                           true ->
+                                               [];
```
ferd, line 332: shouldn't this be the map value of ssl_options in Endpoint? Isn't that dropping the config for the merge?

tsloughter: No, that is in `Endpoint` which is the second argument of the `maps:merge`. It could be made the SslOptions from the `Endpoint` but that wouldn't change anything about the result.

ferd (approved):

PR #442, opened 2022-08-26
Exporter app env SSL options

### LINKED ISSUE #417: using endpoint URL and top level ssl_options in exporter config causes error
When configuring opentelemetry_exporter 1.0.4 for client mTLS (issued certificate and private key), setting a single URL value for the endpoint and top level `ssl_options` causes an error and no data is sent to the endpoint.

example config:
```elixir
config :opentelemetry_exporter,
  otlp_protocol: :http_protobuf,
  otlp_endpoint: "https://example.com:4318",
  ssl_options: [cert: ..., key: ...]
```

example error:
```
[warning] OTLP exporter failed to initialize with exception :error::function_clause
```

expected data to be sent to the endpoint authorized by client cert and to see output:
```
[info] OTLP exporter successfully initialized
```

NOTE:

Using a map instead of a URL and moving the ssl_options there *does* work, but this should not be needed.

```elixir
config :opentelemetry_exporter,
    otlp_protocol: :http_protobuf,
    otlp_endpoint: %{
      scheme: "https",
      host: "example.com",
      port: 4318,
      ssl_options: ssl_options
    }
```

See also: https://cloud-native.slack.com/archives/C01N75YMZCN/p1658942572382929

kenichi:
* support `ssl_options` top-level application environment key
* handle URI string endpoint value parsing errors
* add example `ssl_options` config to README
* closes #417

### DIFF
--- a/apps/opentelemetry_exporter/README.md
+++ b/apps/opentelemetry_exporter/README.md
@@ -57,21 +57,15 @@ endpoints used by the exporter with the Application environment variable
 `ssl_options`:
 
 ```erlang
-%% mTLS example
-SSLOptions = [{cert, "..."}, {key, "..."}] ++ tls_certificate_check:options("example.com"),
-
 {opentelemetry_exporter,
   [{otlp_endpoint, "https://example.com:4318"},
-   {ssl_options, SSLOptions}]}
+   {ssl_options, [{cert, "..."}, {key, "..."}]}]}
 ```
 
 ```elixir
-# mTLS example
-ssl_options = [cert: "...", key: "..."] ++ :tls_certificate_check.options("example.com")
-
 config :opentelemetry_exporter,
   otlp_endpoint: "https://example.com:4318",
-  ssl_options: ssl_options
+  ssl_options: [cert: "...", key: "..."]
 ```
 
 See [secure coding with
--- a/apps/opentelemetry_exporter/README.md
+++ b/apps/opentelemetry_exporter/README.md
@@ -23,6 +23,7 @@ Available configuration keys:
 - `otlp_traces_protocol`: The transport protocol to use for exporting traces, supported values: `grpc` and `http_protobuf`. Defaults to `http_protobuf`.
 - `otlp_compression`: Compression type to use, supported values: `gzip`. Defaults to no compression.
 - `otlp_traces_compression`: Compression type to use for exporting traces, supported values: `gzip`. Defaults to no compression.
+- `ssl_options`: See [below](#ssl-options).
 
 ```erlang
 {opentelemetry_exporter,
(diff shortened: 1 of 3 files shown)


apps/opentelemetry_exporter/README.md
```diff
@@ -53,7 +54,25 @@ package also provides the [CA certificates from Mozilla](https://curl.se/docs/ca
 
 The user can override these options either as part of the endpoint or for all
 endpoints used by the exporter with the Application environment variable
-`ssl_options`
+`ssl_options`:
+
+```erlang
+%% mTLS example
+SSLOptions = [{cert, "..."}, {key, "..."}] ++ tls_certificate_check:options("example.com"),
```
tsloughter, line 61: Erlang config is not code. We'd need some option to allow adding the `tls_certificate_check:options(Domain)` being added to the ssl options for Erlang users.

You can just remove it for now and we get it added later, if you'd rather just get this merged.

apps/opentelemetry_exporter/src/opentelemetry_exporter.erl
```diff
@@ -411,23 +417,34 @@ merge_with_environment(Opts) ->
...
+               ssl_options => undefined},
 
     AppEnv = application:get_all_env(opentelemetry_exporter),
     AppOpts = otel_configuration:merge_list_with_environment(config_mapping(), AppEnv, Config),
 
-    %% append the default path `/v1/traces` only to the path of otlp_endpoint
-    Opts1 = update_opts(otlp_endpoint, endpoints, ?DEFAULT_HTTP_ENDPOINTS, AppOpts, Opts, fun endpoints_append_path/1),
-    Opts2 = update_opts(otlp_traces_endpoint, endpoints, ?DEFAULT_HTTP_ENDPOINTS, AppOpts, Opts1, fun maybe_to_list/1),
+    %% check for error in app env value parsing
+    case maps:get(otlp_endpoint, AppOpts) of
```
tsloughter, line 427: I think this is what I wanted to avoid by making a change to `transform`, but now that I see this it makes sense to ignore some options like headers, traces_headers if the endpoint is unparsable, so this may be fine, just need to think more.

apps/opentelemetry_exporter/README.md
kenichi: Ah, thanks. I simplified both blocks, so they're closer to the examples above.

tsloughter:
Argh, sorry, forgot to merge this before the latest release. 

Can you add a CHANGELOG entry? I'll get this in the next release, and try to have a more frequent cadence of releases.

tsloughter (approved):

kenichi:
Ah, long weekend - missed this until now. I'll try a changelog entry in a new PR. Thanks! 🙇🏽

PR #445, opened 2022-08-31
Use return of callback for measurements of observable instruments

tsloughter:
Alternative to #438

### DIFF
--- a/apps/opentelemetry_api_experimental/src/otel_instrument.erl
+++ b/apps/opentelemetry_api_experimental/src/otel_instrument.erl
@@ -18,7 +18,7 @@
 -module(otel_instrument).
 
 -export([new/7,
-         new/8,
+         new/9,
          is_monotonic/1]).
 
 -include("otel_metrics.hrl").
@@ -50,16 +50,17 @@ new(Module, Meter, Kind, Name, Description, Unit, ValueType) ->
                 value_type  = ValueType,
                 unit        = Unit}.
 
--spec new(module(), otel_meter:t(), kind(), name(), description(), unit(), value_type(), callback()) -> t().
-new(Module, Meter, Kind, Name, Description, Unit, ValueType, Callback) ->
-    #instrument{module      = Module,
-                meter       = Meter,
-                name        = Name,
-                description = Description,
-                kind        = Kind,
-                value_type  = ValueType,
-                unit        = Unit,
-                callback    = Callback}.
+-spec new(module(), otel_meter:t(), kind(), name(), description(), unit(), value_type(), callback(), term()) -> t().
+new(Module, Meter, Kind, Name, Description, Unit, ValueType, Callback, CallbackArgs) ->
+    #instrument{module        = Module,
+                meter         = Meter,
+                name          = Name,
+                description   = Description,
+                kind          = Kind,
+                value_type    = ValueType,
+                unit          = Unit,
+                callback      = Callback,
+                callback_args = CallbackArgs}.
 
 is_monotonic(#instrument{kind=?KIND_COUNTER}) ->
     true;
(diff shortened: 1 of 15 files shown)


tsloughter:
I'm not a fan of how it ultimately aggregates and checkpoints in the reader but those are internal and can change later.

apps/opentelemetry_experimental/src/otel_meter_server.erl
```diff
@@ -107,6 +120,15 @@ record(Provider, Instrument, Number, Attributes) ->
                                                      value=Number,
                                                      attributes=Attributes}}).
 
+-spec sync_record(atom(), otel_instrument:t(), number(), opentelemetry:attributes_map()) -> ok.
+sync_record(Provider, Instrument, Number, Attributes) ->
```
ferd, line 124: I would maybe recommend naming this `record_sync` -- my initial impression was that this would synchronize records, rather than synchronously record something. Alternatively, turning `record/4` into `async_record/4` also remains unambiguous.

apps/opentelemetry_experimental/src/otel_meter_server.erl
```diff
@@ -170,15 +204,46 @@ code_change(State) ->
...
+    ViewMatches = otel_view:match_instrument_to_views(Instrument, Views),
+    lists:map(fun(Reader=#reader{callbacks_tab=CallbackTab,
+                                 view_aggregation_tab=ViewAggregationTab}) ->
+                      Matches = per_reader_aggregations(Reader, Instrument, ViewMatches),
+                      _ = ets:insert(ViewAggregationTab, {Instrument, Matches}),
+                      case Instrument#instrument.callback of
+                          undefined ->
+                              ok;
+                          {Callback, CallbackArgs} ->
+                              ets:insert(CallbackTab, {Callback, CallbackArgs, [Instrument]});
```
ferd, line 218: This may have a weird effect? The callback acts as a key, but the key can be any function. Does it mean that adding new instruments unregisters older callbacks? If so do we need a bag table rather than a set table?

apps/opentelemetry_api_experimental/src/otel_instrument.erl
```diff
@@ -29,7 +29,7 @@
                 ?KIND_OBSERVABLE_GAUGE | ?KIND_UPDOWN_COUNTER | ?KIND_OBSERVABLE_UPDOWNCOUNTER.
 -type unit() :: atom(). %% latin1, maximum length of 63 characters
 -type value_type() :: ?VALUE_TYPE_INTEGER | ?VALUE_TYPE_FLOAT.
--type callback() :: fun().
+-type callback() :: fun(([t()]) -> ok).
```
ferd, line 32: It seems that most of the callbacks in tests return `{number(), state()} | {Meter, state()}` or some variation thereof, and not just `ok`?

tsloughter: Yup, this actually relates to your other comment about `sync_record`. The `ok` and `sync_record` are from when this was not using returns of the callback to record the measurements. So `sync_record` needs to be removed and this updated to be returning the list of measurements.

apps/opentelemetry_experimental/src/otel_meter_server.erl
tsloughter: This particular one is a single instrument with a single callback, so there is no new instruments.

Further down is the line where it inserts a new callback with multiple instruments registered to it.

It could be a bag so the same `fun` could be used by multiple instruments separately, which makes sense since the funs can't interfere with each other and they take an args term which could change what they do.

ferd (approved):
Nice. I like also having explicit args for the callback since it will let people use fully-qualified anonymous functions rather than inlined funs, which means it'll be friendlier to hot code loading as well.

PR #451, opened 2022-09-07
Fix InstrumentationScope in OTLP protos

tsloughter:


tsloughter:
Basically `instrumentation_scope` was supposed to be named `scope`.

Making an issue to switch to using gpb generated records instead of maps so we catch this shit at compile time.

PR #456, opened 2022-09-09
OTLP Metrics exporting

tsloughter:
#455 

This is rough because I still don't know if we want to be passing metrics around like this instead of as a table but I think it is an ok first step.


apps/opentelemetry_exporter/src/otel_otlp_common.erl
```diff
@@ -0,0 +1,103 @@
...
+%% Unless required by applicable law or agreed to in writing, software
+%% distributed under the License is distributed on an "AS IS" BASIS,
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+%%
+%% @doc
+%% @end
+%%%-------------------------------------------------------------------------
+-module(otel_otlp_common).
```
ferd, line 18: would it make sense to call this `otel_otlp_convert` or something like that rather than `..._common` since the latter will likely trend towards being a 'util' module that's a sort of dumping ground?

ferd: I guess not necessarily, otherwise the whole `to_proto` function in `otel_otlp_metrics` sort of feels like it's in the wrong module.

ferd (approved):

PR #461, opened 2022-09-19
Add otel_monitor

derekkraan:
Following up on this PR: https://github.com/open-telemetry/opentelemetry-erlang-contrib/pull/109

This PR adds a new module, `otel_monitor`. We associate the span with a monitor ref in an ets table, and end the span when we detect that the process has died.

Looking forward to feedback for this one.

### DIFF
--- a/apps/opentelemetry/src/otel_monitor.erl
+++ b/apps/opentelemetry/src/otel_monitor.erl
@@ -1,0 +1,77 @@
+%%%------------------------------------------------------------------------
+%% Copyright 2020, OpenTelemetry Authors
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
+%%
+%% http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing, software
+%% distributed under the License is distributed on an "AS IS" BASIS,
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+%%
+%% @doc
+%% Process that can optionally monitor the process a span is in and end the
+%% span if the process stops for any reason with the span still unfinished.
+%% @end
+%%%-------------------------------------------------------------------------
+-module(otel_monitor).
+
+-behaviour(gen_server).
+
+-export([start_link/0,
+				 monitor/1]).
+
+-export([init/1,
+				 handle_call/3,
+				 handle_cast/2,
+				 handle_info/2]).
+
+-include("otel_tracer.hrl").
+
+-define(SERVER, ?MODULE).
+-define(TABLE, ?MODULE).
+
+start_link() ->
+  gen_server:start_link({local, ?SERVER}, ?MODULE, nil, [{name, ?SERVER}]).
+
+init(nil) ->
+  _TableId = ets:new(?TABLE, [set, named_table]),
+	{ok, nil}.
+
+handle_call({monitor, SpanCtx, Pid}, _From, State) ->
+  Ref = erlang:monitor(process, Pid),
+	true = ets:insert(?TABLE, {Ref, SpanCtx}),
+	{reply, ok, State}.
+
+handle_info({'DOWN', Ref, process, _Pid, normal}, State) ->
+  case ets:take(?TABLE, Ref) of
+		[] -> nil;
+		[{_Ref, SpanCtx}] -> otel_span:end_span(SpanCtx)
+	end,
+  {noreply, State};
+
+handle_info({'DOWN', Ref, process, _Pid, {shutdown, _}}, State) ->
+  case ets:take(?TABLE, Ref) of
+		[] -> nil;
+		[{_Ref, SpanCtx}] -> otel_span:end_span(SpanCtx)
+	end,
+  {noreply, State};
+
+handle_info({'DOWN', Ref, process, _Pid, Reason}, State) ->
+  case ets:take(?TABLE, Ref) of
+		[] -> nil;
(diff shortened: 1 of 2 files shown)


tsloughter:
Just had a thought, maybe this should be part of the sweeper?

apps/opentelemetry/src/otel_monitor.erl
```diff
@@ -0,0 +1,77 @@
...
+
+handle_call({monitor, SpanCtx, Pid}, _From, State) ->
+  Ref = erlang:monitor(process, Pid),
+	true = ets:insert(?TABLE, {Ref, SpanCtx}),
+	{reply, ok, State}.
+
+handle_info({'DOWN', Ref, process, _Pid, normal}, State) ->
+  case ets:take(?TABLE, Ref) of
+		[] -> nil;
+		[{_Ref, SpanCtx}] -> otel_span:end_span(SpanCtx)
```
tsloughter, line 52: Could be more than 1 span being monitored in the same process.

apps/opentelemetry/src/otel_monitor.erl
```diff
@@ -0,0 +1,77 @@
...
+      otel_span:add_event(SpanCtx, 'Process died', #{<<"reason">> => iolist_to_binary(io_lib:format("~p", [Reason]))}),
+			otel_span:end_span(SpanCtx)
+	end,
+  {noreply, State}.
+
+handle_cast(_Msg, State) ->
+    {noreply, State}.
+
+monitor(SpanCtx) ->
+  ok = gen_server:call(?SERVER, {monitor, SpanCtx, self()}),
```
tsloughter, line 76: User probably doesn't want to crash if the monitor fails. I'd say probably just want a `cast` unless you think some users will want to be able to rely on this feature in such a way that it should return `false` if the call fails?

apps/opentelemetry/src/otel_monitor.erl
derekkraan: Yes, but each span will get its own ref, no? I guess this is what I will find out when I write tests...

apps/opentelemetry/src/otel_monitor.erl
derekkraan: I like that idea, returning true / false based for success / failure.

derekkraan:
My first instinct would be to keep it self-contained and not include it in the sweeper. It's also a different use-case to the sweeper. The usage patterns of the sweeper are also quite different; it does a lot of work sometimes, while the monitor is doing a small amount of work all of the time, and it will also end up in the hot path of people's code, so it should be fast.

Perhaps a cast is also better from the perspective of speed...

apps/opentelemetry/src/otel_monitor.erl
tsloughter: Ooh, my mistake. I was thinking it was like my implementation which monitors all spans in a process, so it looks up all spans for a pid, not by ref.

Which is a separate question I have, is it really worth doing it per-span, are there spans that a user wouldn't want ended when a process crashes?

derekkraan: If you are asking why I didn't write a PR that monitors every span's process, it's because bigger changes like that are less likely to be accepted. I am aiming with this PR for the incremental approach: first offer the api optionally, if it turns out that everyone uses it all the time, then contemplate making it mandatory.

But if you are comfortable going directly to monitoring every span's process, then I am happy coding that up.

I'm not sure whether it's cheaper to have a single monitor per pid and juggling spans associated with that pid in an ets table, or just creating a new monitor and ets row per span.

tsloughter: No, I'm asking why not monitor every span in a single process the user requests.

If the user requests to monitor Span X in process P do they really not want Span Y that is also active in process P to be ended when the process dies? If they do they have to call `monitor(X)` and `monitor(Y)`. The API I had in mind was just `monitor()`, and then every span in `self()` is monitored. I suppose also a `monitor(Pid)` makes sense to include if we go that route.

tsloughter: I'm asking in slack. It is hard to know what is best without people's input on how they'd use it, or if they are already using their own solution, like you have been.

hkrutzer: >If the user requests to monitor Span X in process P do they really not want Span Y that is also active in process P to be ended when the process dies? If they do they have to call monitor(X) and monitor(Y). 

I would want to have this applied to every span.

tsloughter:
@derekkraan What do you think of the alternative of monitoring every span in the process that is monitored?

I also wonder about the api being an option passed to `with/start_span`. It would still monitor all spans in the process but make it more obvious when a user is looking at the API and `with_span`/`start_span` that they can do it.

derekkraan:
Re: monitoring every span, sounds good, and it appears this is what everyone wants, but I think then the exception event should only be added to the outermost one? Or the one that has been explicitly monitored? What do you think?

I do think there should also be an option to `with/start_span`. I was initially treating that as something that could be added later but if there is already agreement on what that should look like (not that hard perhaps, just `monitor: true`?), then I can certainly look at adding that in this PR.

hkrutzer:
>the exception event should only be added to the outermost one? Or the one that has been explicitly monitored? What do you think?

Yes outermost seems fine.

tsloughter:
I don't think outermost is a good idea. It would be a bit of a pain to implement. And really I would think the inner most would make the most sense as its where the failure is more likely to have happened?

I think it is better to just mark any active span in a process with the information.

hkrutzer:
If the innermost is easier I think that's fine too, any trace viewer should handle that fine.

tsloughter:
My latest https://github.com/open-telemetry/opentelemetry-erlang/pull/602

PR #468, opened 2022-09-23
Logging Handler

tsloughter:
This is a very simple version of what is needed to support the logging spec and is in part duplicating code from the batch span processor.

It has no overload protection and this will wait until we know what is planned for log sampling in the spec as it will overlap.

Oh, and no tests. I've only manually tested by bringing up the collector and a rebar3 shell in which I add the handler and then see logs going to the collector through the collector's console output where it will log each log it receives.


apps/opentelemetry_exporter/src/opentelemetry_exporter.erl
```diff
@@ -273,6 +273,24 @@ export(metrics, Tab, Resource, #state{protocol=grpc,
...
+        {ok, _Response, _ResponseMetadata} ->
+            ok;
+        {error, {Status, Message}, _} ->
+            ?LOG_INFO("OTLP grpc export failed with GRPC status ~s : ~s", [Status, Message]),
+            error;
+        {http_error, {Status, _}, _} ->
+            ?LOG_INFO("OTLP grpc export failed with HTTP status code ~s", [Status]),
+            error;
+        {error, Reason} ->
+            ?LOG_INFO("OTLP grpc export failed with error: ~p", [Reason]),
```
ferd, line 291: is there any chance these logs messages are gonna be fed back into the OTel exporter and cause issues with write amplification?

apps/opentelemetry_exporter/src/otel_otlp_common.erl
```diff
@@ -64,12 +64,20 @@ to_any_value(Value) when is_map(Value) ->
 to_any_value(Value) when is_tuple(Value) ->
     #{value => {array_value, to_array_value(tuple_to_list(Value))}};
 to_any_value(Value) when is_list(Value) ->
-    case is_proplist(Value) of
+    case io_lib:printable_list(Value) of
```
ferd, line 67: This is a tricky thing because it relies on the `+pc <unicode|latin1>` value to work. `io_lib:printable_range/0` will define what you get. 

You want to use `io_lib:printable_unicode_list(Value)` here, which will assume unicode output and tell you if that unicode output is printable or not. I had code for the `flatlog` library in tests that ended up looking like:

```erlang
case io:printable_range() of
    latin1 -> io_lib:printable_list(S);
    unicode -> io_lib:printable_unicode_list(S)
end
``` 

Since we're shuttling bytes over the network and we don't care about whether the VM is configured to use only unicode or not, we should just flat out do as the docstring says, assume UTF8, and call `io_lib:printable_unicode_list(S)` here.

apps/opentelemetry_exporter/src/otel_otlp_common.erl
```diff
@@ -64,12 +64,20 @@ to_any_value(Value) when is_map(Value) ->
...
         true ->
-            #{value => {kvlist_value, to_key_value_list(Value)}};
-        false ->
-            #{value => {array_value, to_array_value(Value)}}
-    end.
+            #{value => {string_value, unicode:characters_to_binary(Value)}};
```
ferd, line 69: This may return a non-string value. Is this okay? Unicode conversion can return:

> [binary()](https://www.erlang.org/doc/man/erlang.html#type-binary) |
    {error, [binary()](https://www.erlang.org/doc/man/erlang.html#type-binary), RestData} |
    {incomplete, [binary()](https://www.erlang.org/doc/man/erlang.html#type-binary), [binary()](https://www.erlang.org/doc/man/erlang.html#type-binary)}

apps/opentelemetry_exporter/src/otel_otlp_logs.erl
```diff
@@ -0,0 +1,240 @@
...
+    Time = opentelemetry:timestamp(),
+    {SeverityNumber, SeverityText} = level_to_severity(Level),
+    Body1 = case format_msg(Body, Metadata, Config) of
+                S when ?IS_STRING(S) ->
+                    %% if body is a string, make it a single line
+                    T = lists:reverse(
+                          trim(
+                            lists:reverse(
+                              trim(S, false)), true)),
+                    re:replace(T,",?\r?\n\s*",", ",
```
ferd, line 78: if I'm reading this right, if you have two linebreaks this will leave one behind. You likely want `(?:\r\n|\n|\s)+` as a group for the linebreaks. Also since you're using `,?` and then replacing with `, `, this may inject commas into logs. To keep it the same we probably want to use `(,?)` and reinject that capture group.

This looks a bit like:

```
1> re:replace("a, \r\n  b  \n\n  c,d, e f  ", "(,?)(\\r\\n|\\n|\\s)+", "\\g{1} ", [global, {return, list}]).
"a, b c,d, e f "
``` 
which preserves the original meaning but collapses duplicated whitespace.

ferd (changes requested):

apps/opentelemetry_experimental/src/otel_log_handler.erl
```diff
@@ -0,0 +1,250 @@
...
+-include_lib("opentelemetry_api/include/opentelemetry.hrl").
+
+-export([start_link/2]).
+
+-export([log/2,
+         adding_handler/1,
+         removing_handler/1,
+         changing_config/3,
+         filter_config/1,
+         report_cb/1]).
```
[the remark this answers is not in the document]
tsloughter: These are OTP logger handler callbacks.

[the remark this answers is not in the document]
tsloughter: And note it is "adding_handler" and such because it doesn't actually do the "add".  It is the callback you get called when the handler is added, the actual add is done in OTP.

apps/opentelemetry_experimental/src/otel_log_handler.erl
```diff
@@ -0,0 +1,250 @@
...
+
+handle_event({call, From}, {changing_config, _SetOrUpdate, _OldConfig, NewConfig}, Data) ->
+    {keep_state, Data#data{config=NewConfig}, [{reply, From, NewConfig}]};
+handle_event({call, From}, {removing_handler, Config}, _Data) ->
+    %% TODO: flush
+    {keep_state_and_data, [{reply, From, Config}]};
+handle_event({call, From}, {filter_handler, Config}, Data) ->
+    {keep_state, Data, [{reply, From, Config}]};
+handle_event({call, From}, {filter_config, Config}, Data) ->
+    {keep_state, Data, [{reply, From, Config}]};
```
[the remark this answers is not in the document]
tsloughter: Na, it is supposed to return with the Config being used.

apps/opentelemetry_experimental/src/otel_log_handler.erl
```diff
@@ -0,0 +1,250 @@
...
+            =:= failed_not_retryable
+    catch
+        Kind:Reason:StackTrace ->
+            ?LOG_WARNING(#{source => exporter,
+                           during => export,
+                           kind => Kind,
+                           reason => Reason,
+                           exporter => ExporterModule,
+                           stacktrace => StackTrace}, #{report_cb => fun ?MODULE:report_cb/1}),
+            true
```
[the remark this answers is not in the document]
tsloughter: I wasn't considering it retryable is why I did `true`. Since it'd be something like a protobuf encoding exception.

apps/opentelemetry_experimental/src/otel_log_handler.erl
```diff
@@ -0,0 +1,250 @@
...
+
+exporting({timeout, export_logs}, export_logs, _) ->
+    {keep_state_and_data, [postpone]};
+exporting(enter, _OldState, _Data) ->
+    keep_state_and_data;
+exporting(internal, export, Data=#data{exporter=Exporter,
+                                       resource=Resource,
+                                       config=Config,
+                                       batch=Batch}) when map_size(Batch) =/= 0 ->
+    _ = export(Exporter, Resource, Batch, Config),
```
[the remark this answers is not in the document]
tsloughter: Interesting idea... I planned to simplify it by breaking out the shared logic from the `batch_processor` module. If that works there too then it makes sense to do.

apps/opentelemetry_exporter/src/otel_otlp_common.erl
tsloughter: Ah, yea, we do want always to do `io_lib:printable_unicode_list(S)` because that is required by the protocol. So I'll update to that.

apps/opentelemetry_exporter/src/otel_otlp_common.erl
tsloughter: I always forget about that. I'll fix it, think it needs to be an exception and drop this value if it fails.

apps/opentelemetry_exporter/src/otel_otlp_logs.erl
tsloughter: I took this from logger https://github.com/erlang/otp/blob/master/lib/kernel/src/logger_formatter.erl#L75-L84

apps/opentelemetry_experimental/src/otel_log_handler.erl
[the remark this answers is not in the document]
ferd: oh well there goes my memory of how to implement this stuff. Disregard then.

apps/opentelemetry_exporter/src/otel_otlp_logs.erl
ferd: alright, that works then. Getting consistency in.

apps/opentelemetry_exporter/src/otel_otlp_common.erl
tsloughter: Updated. I also updated how binary strings are encoded to work properly with the spec. If the binary is not found to be a unicode string it is encoded as a `bytes_value` now.

ferd (approved):
Major concerns addressed. Still curious about the logging handler failing and possibly generating more logs to handle and the risk this might represent.

apps/opentelemetry_exporter/src/opentelemetry_exporter.erl
tsloughter: oh they definitely will

tsloughter: Maybe a `domain` can be used here but then by default all these logs would be lost.. ugh

PR #469, opened 2022-09-25
add to/from_string functions and documentation

jowrjowr:
This provisionally solves a big problem I have been having with stringifying an otel trace and passing it around. The :otel_propagator_text_map inject/extract functions were completely unknown to me and this will codify it a lot better, add some documentation, and explain some (what I consider) unusual output from them.

I'm currently using the extract/inject functionality in a work project so I know it works, but it was ever so confusing to understand and this will help the next people who have to deal.

Notes: 

This is all based on elixir slack opentelemetry channel discussions plus a LOT of my own personal experimentation under different cases. If I'm missing something important, I'm not seeing it right now.

Additionally there is a typespec bug in `:otel_propagator_text_map.extract/1` that's making noise. The `:otel_ctx.t()` spec is `map()` which is not correct as `:otel_propagator_text_map.extract([{"traceparent", "meow"}])` (a clearly malformed string span), but adjusting this to what I consider a more correct `map() | undefined` makes other things unhappy and pulling at that string feels yakshave-y right this second.

If I could find the spec of the string I could setup more formal validation of the input. Or just plug and chug?

### DIFF
--- a/apps/opentelemetry_api/lib/open_telemetry/tracer.ex
+++ b/apps/opentelemetry_api/lib/open_telemetry/tracer.ex
@@ -221,4 +221,63 @@ defmodule OpenTelemetry.Tracer do
   def update_name(name) do
     :otel_span.update_name(:otel_tracer.current_span_ctx(), name)
   end
+
+  @doc """
+  Takes the span context of an active process and returns it as a string.
+
+  Function is complementary with `from_string/1`.
+
+  Example usage context:
+
+  You have a function that broadcasts with pubsub, and executes on another node. Or
+  perhaps an oban job. Or some sort of microservice architecture.
+
+  The consuming node will not have any of the OTel process information so you cannot successfully
+  pass the full context you would normally get from `OpenTelemetry.Ctx.get_current/0`.
+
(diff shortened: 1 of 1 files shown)


tsloughter:
You are right that the type spec is wrong, but it shouldn't be `undefined`. The return for `extract` should be `otel_ctx:token()`. The return for `extract_to` remains `otel_ctx:t()`. It doesn't matter that `:otel_propagator_text_map.extract([{"traceparent", "meow"}])` won't extract a trace context, it still is running `otel_ctx:attach/1` at the end.

But these functions `to_string` and `from_string` can't work. Aside from the fact the propagators could also be adding other data, like baggage, to the text map, or the fact that it assumes `traceparent` is the key name, there are propagators like B3Multi that don't return a single string for the trace context, so returning `string()` wouldn't work.

I think better would be improved docs, or suggestions on how I can improve them.

Also wondering how you are using `inject` and `extract` in your work? Are you using it as just a string, meaning you aren't using propagation to inject/extract from a set of like HTTP headers? That certainly isn't expected, so makes sense it would seem more complicated than it needs to be.

PR #471, opened 2022-09-30
first pass at gradualizer cleanup

tsloughter:


tsloughter:
Gradualizer has found a number of useful things. Mostly it is just type spec cleanup but also some bugs waiting to happen. One big one it picked up on is we are using native erlang time unit for a variable named `_nano`. It didn't pick it up for that exact reason, but still, a confusing issue waiting to happen.

There are many more of these I hope to do because there are many more gradualizer errors.

ferd (approved):

PR #474, opened 2022-10-04
phase 2 of gradualization

tsloughter:
The final errors are ones I don't think can be resolved without updates to Gradualizer to be able to ignore errors by function and not only at the module level.

At that point we can ignore the ones that aren't fixable soon (I think the `undefined function module_info` may be one that we won't have to ignore) and add gradualizer to github actions CI.

Being able to ignore on a per-function basis would also allow getting rid of those `$1` additions to typespecs which we already have module attributes to ignore those issues in dialyzer. 

The remaining errors are below and a link to a discussion on the Gradualizer repo I started.

https://github.com/josefs/Gradualizer/discussions/474

```
/home/tristan/Devel/opentelemetry-erlang/_build/default/lib/opentelemetry_api/ebin/otel_propagator_b3.beam: The clause on line 90 at column 1 cannot be reached
/home/tristan/Devel/opentelemetry-erlang/_build/default/lib/opentelemetry/ebin/otel_exporter.beam: Call to undefined function grpcbox:module_info/0 on line 73 at column 40
/home/tristan/Devel/opentelemetry-erlang/_build/default/lib/opentelemetry/ebin/otel_resource.beam: Nonexhaustive patterns on line 77 at column 1
Example values which are not covered:
	#resource{}
	#resource{}
/home/tristan/Devel/opentelemetry-erlang/_build/default/lib/opentelemetry_experimental/ebin/otel_metric_reader.beam: The operator '++' on line 175 at column 27 is expected to have type _TyVar-576460752303409663 which is too precise to be statically checked
/home/tristan/Devel/opentelemetry-erlang/_build/default/lib/opentelemetry_experimental/ebin/otel_metric_reader.beam: The operator '++' on line 188 at column 87 is expected to have type _TyVar-576460752303409215 which is too precise to be statically checked
```


apps/opentelemetry/include/otel_span.hrl
```diff
@@ -18,48 +18,48 @@
 
 -record(span, {
           %% 128 bit int trace id
-          trace_id                :: opentelemetry:trace_id() | undefined,
+          trace_id                :: opentelemetry:trace_id() | undefined | '_',
```
ferd, line 21: ugh yeah this whole file's annotations are sort of annoying. I wish there was a way for records to be annotated as match-spec/select-spec friendly.

apps/opentelemetry_api/src/otel_utils.erl
```diff
@@ -26,3 +35,15 @@ format_exception(Kind, Reason, StackTrace) ->
...
+format_binary_string(Format, Data) ->
+    ?assert_type(unicode:characters_to_binary(io_lib:format(Format, Data)), binary()).
+
+-spec format_binary_string(io:format(), [term()], [{chars_limit, integer()}]) -> binary().
+format_binary_string(Format, Data, Options) ->
+    ?assert_type(unicode:characters_to_binary(io_lib:format(Format, Data, Options)), binary()).
+
+-spec assert_to_binary(unicode:chardata()) -> binary().
+assert_to_binary(String) ->
+    ?assert_type(unicode:characters_to_binary(String), binary()).
```
ferd, line 49:
```suggestion
-spec format_binary_string(io:format(), [term()]) -> binary().
format_binary_string(Format, Data) ->
    assert_to_binary(io_lib:format(Format, Data)).

-spec format_binary_string(io:format(), [term()], [{chars_limit, integer()}]) -> binary().
format_binary_string(Format, Data, Options) ->
    assert_to_binary(io_lib:format(Format, Data, Options)).

-spec assert_to_binary(unicode:chardata()) -> binary().
assert_to_binary(String) ->
    Bin = ?assert_type(unicode:characters_to_binary(String), binary()),
    true = is_binary(Bin),
    Bin.
```

Would this be safer to actually make the runtime semantics match the type analysis's own semantics? The concern I have is that by asserting the type on many conversions like that but actually still having a chance to return tuples, this can create a false sense of security about what the return values of the call might be and cause subtle bugs (e.g. filtering on `is_tuple/1` in a guard with mixed data). By having the assertion be an actual assertion at runtime too, bad conversions are caught at the call-site rather than later down a dependency chain.

ferd: A clearer error message could be obtained with `is_binary(Bin) orelse error({bad_binary_conversion, String})` as well, or something to that effect.

apps/opentelemetry_experimental/include/otel_metrics.hrl
```diff
@@ -5,7 +5,7 @@
 -define(AGGREGATION_TEMPORALITY_UNSPECIFIED, aggregation_temporality_unspecified).
 
 -record(meter, {module                  :: module(),
-                instrumentation_library :: otel_tracer_server:instrumentation_library() | undefined,
+                instrumentation_scope  :: opentelemetry:instrumentation_scope() | undefined,
```
ferd, line 8:
```suggestion
                instrumentation_scope   :: opentelemetry:instrumentation_scope() | undefined,
```

ferd (changes requested):
Looks good, lots of subtle things caught. Got a few changes suggested.

Also a bit curious about whether we should use `binary()` in asserts or go for the more accurate `unicode:unicode_binary()` -- can Gradualizer understand that it is a subset of `binary()` and that functions that accept binaries should accept unicode variations?

apps/opentelemetry/include/otel_span.hrl
tsloughter: oooh, maybe there is a better way. I commented in the gradualizer discussion () about if its a possibility to just ignore those. Since its only in a matchspec it should even be possible to not end up ignoring errors if for some reason a user is actually use those values in their record and not just in match specs.

apps/opentelemetry_api/src/otel_utils.erl
tsloughter: Good point about assert not asserting. I think we'll need a couple functions for use in different contexts. I meant to limit the use of these to only when I knew the values were already checked. Since in the public facing API we don't want to `error` it needs to be handled anywhere there might be a failure like this (or a return of an error). So I think I'll add a version that returns `{ok, binary} | {error, bad_binary_conversion}` and another that asserts that can be used in the SDK.

tsloughter:
I try to use `unicode_binary` or `latin1_binary` anywhere I know a specific encoding is required. I think only baggage is limited to latin1 values.

tsloughter:
Just realized use of the header makes a compile time dependency of gradualizer for each app. I'd rather not do that, so going to copy in the header to the api.

ferd (approved):

PR #477, opened 2022-10-06
Fix exporter stable

### LINKED ISSUE #476: Input checking
We are getting a lot of errors that look like this:

```
info [info] exporter threw exception: exporter=:opentelemetry_exporter exception error: bad argument
  in function  byte_size/1
     called as byte_size({error,<<"%PDF-1.4\n%">>,
                                <<"âãÏÓ\n7 0 obj\n<</Length 694/Filter/FlateDecode>>stream\nx"...>>})
     *** argument 1: not a bitstring
  in call from opentelemetry_exporter_trace_service_pb:e_type_string/3 (/app/deps/opentelemetry_exporter/src/opentelemetry_exporter_trace_service_pb.erl, line 942)
  in call from opentelemetry_exporter_trace_service_pb:e_mfield_key_value_value/3 (/app/deps/opentelemetry_exporter/src/opentelemetry_exporter_trace_service_pb.erl, line 870)
  in call from opentelemetry_exporter_trace_service_pb:e_mfield_event_attributes/3 (/app/deps/opentelemetry_exporter/src/opentelemetry_exporter_trace_service_pb.erl, line 778)
  in call from opentelemetry_exporter_trace_service_pb:e_field_event_attributes/3 (/app/deps/opentelemetry_exporter/src/opentelemetry_exporter_trace_service_pb.erl, line 784)
  in call from opentelemetry_exporter_trace_service_pb:encode_msg_event/3 (/app/deps/opentelemetry_exporter/src/opentelemetry_exporter_trace_service_pb.erl, line 315)
  in call from opentelemetry_exporter_trace_service_pb:e_mfield_span_events/3 (/app/deps/opentelemetry_exporter/src/opentelemetry_exporter_trace_service_pb.erl, line 811)
  in call from opentelemetry_exporter_trace_service_pb:e_field_span_events/3 (/app/deps/opentelemetry_exporter/src/opentelemetry_exporter_trace_service_pb.erl, line 817) 
```

I believe this is caused by invalid data being entered into otel (looks like via event attributes in this case).

As far as I know, the data shapes are very clear as to what is and is not acceptable input for the otel-erlang library, but this is not being checked at the moment a function is being called. This leads to errors surfacing much later, which is near impossible to debug.

Is my suspicion correct? The other alternative is that there is a bug in the otel_exporter_trace_service but I find that unlikely.

Generally we should check the input to our library carefully at the boundary, to prevent this kind of thing.

So I think this issue is really asking for: comprehensive input checking for public functions in this library.

I would be willing to help work on this if I had a good idea of where to start and confirmation that this is indeed the source of the issue. I do see some things that look like input checking but it's not entirely clear to me where they are insufficient or whether they are intended for this purpose.

tsloughter:
Think I need to break this out to `opentelemetry_exporter_experimental` but first wanted to open this to see if it fixes #476 for @derekkraan


tsloughter:
I still need to update the test profile to add the EXPERIMENTAL flag so it passes tests.

ferd (changes requested):
I'm guessing you're going to need the experimental flag on whatever is running the dialyzer task as well in order to keep passing tests, though I'm guessing this can also point at a runtime error if the check isn't conditionally turned to a no-op there either?

ferd (approved):

PR #481, opened 2022-10-20
use self() for grpc channel name

tsloughter:
using ?MODULE limited us to a single grpcbox client channel. We may want to not have a channel for each Signal (traces, meitrcs, logs) if they are the same configuration, but for now this at least allows each to export -- before only 1 could init their exporter because it would conflict with the other.

Reuse of the channel is not as simple as using ?MODULE and if it is already started use that channel because the endpoint may be different. Not only that, the configuration to the endpoint may be different, so a Channel name would have to be all the distiguishing attributes of a Channel... Just easier to create one for each Signal.

### DIFF
--- a/apps/opentelemetry_exporter/src/opentelemetry_exporter.erl
+++ b/apps/opentelemetry_exporter/src/opentelemetry_exporter.erl
@@ -146,7 +146,8 @@
               endpoint/0,
               protocol/0]).
 
--record(state, {protocol :: protocol(),
+-record(state, {channel :: term(),
+                protocol :: protocol(),
                 channel_pid :: pid() | undefined,
                 headers :: headers(),
                 compression :: compression() | undefined,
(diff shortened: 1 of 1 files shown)


apps/opentelemetry_exporter/src/opentelemetry_exporter.erl
```diff
@@ -170,11 +171,17 @@ init(Opts) ->
                                    undefined -> ChannelOpts;
                                    Encoding -> maps:put(encoding, Encoding, ChannelOpts)
                                  end,
-            case grpcbox_channel:start_link(?MODULE,
+
+            %% Channel name can be any term. To separate Channels per
+            %% the process calling the exporter (like the batch processor,
+            %% metric reader and log handler) use the current pid
+            Channel = self(),
```
ferd, line 178: would it make sense to use `{self(), ?MODULE}` to also avoid any sort of problems with multiple sorts of callers? I guess it wouldn't because the channelpid is stored but yeah.

ferd (approved):

apps/opentelemetry_exporter/src/opentelemetry_exporter.erl
tsloughter: It wouldn't hurt but don't think it'd add anything yea

PR #483, opened 2022-10-31
better support for multiple tracer providers and processors

tsloughter:
tracer providers should be able to be started by users and run independently with their own procesors, and multiple of each processor.

The SDK must register a simple_one_for_one that starts tracer providers under the name otel_tracer_provider_sup. Then a global tracer provider is started under the name otel_tracer_server_global to be used when getting tracers that aren't yet cached.

The key thing is that simple and batch processors must return a modified config with `reg_name` set in the map. This is then in the OnStart and OnEnd lambda of functions to run and their configs so it can call them.

My main problem with this strategy is it requires a name and that makes config a little awkward and possibly not compatible with what will be the config file for SDKs that is being worked on -- but if `name` is required for providers and processors in the config file then this is a good solution actually, so I may want to wait on merging this until that is decided.


tsloughter:
Broke a number of things, unsurprising, and still need tests of when things go wrong (like trying to add a processor with the same name as an existing one, same with providers).

tsloughter:
I'm getting rid of the naming for processors and maybe providers (at least not required). Will do that in a commit on this PR.

tsloughter:
I'd like to get rid of the need to name the ets tables in the batch processor but haven't figured that out and not sure there is a way.

The reasons being it currently passes the reference to the ets table to the exporter, if we want to be able to do concurrent exports then we can't be restricted to 2 tables, so can't just wait for it to finish and then reuse that table.

It also benefits from not having to do a `ets:tab2list` in order to pass off the spans to the exporter, instead allows the exporter to go through the ets table content in an order that matches the protobuf schema.

Would also be nice if ets table names could be dropped, or turned into not a named table, but they can't. I think we are stuck with this.

tsloughter:
This is ready now.

No longer is a name needed from the user. But processors do get a registered name so that they are able to be restarted by the supervisor without a pid change having to happen in the Tracer (which is stored in a persistent term so changing would be terrible).

apps/opentelemetry/src/otel_batch_processor.erl
```diff
@@ -34,9 +34,12 @@
          on_start/3,
          on_end/2,
          force_flush/1,
+         report_cb/1,
+
+         %% deprecated
```
ferd, line 39: are these deprecated for this app or for the whole API in general? I assume the former and it's a question of structure?

May be worth adding 
```erlang
-deprecated({set_exporter, 1, "use otel_tracer_provider instead"}).
``` 
or whatever the proper call would be?

apps/opentelemetry/src/otel_simple_processor.erl
```diff
@@ -29,9 +29,12 @@
          on_start/3,
          on_end/2,
          force_flush/1,
+         report_cb/1,
+
+         %% deprecated
```
ferd, line 34: same note here; worth using the deprecation attribute?

apps/opentelemetry/src/otel_simple_processor.erl
```diff
@@ -52,18 +55,37 @@
...
 
-start_link(Opts) ->
-    gen_statem:start_link({local, ?MODULE}, ?MODULE, [Opts], []).
+    RegisterName = ?NAME_TO_ATOM(?MODULE, Name),
+    Config1 = Config#{reg_name => RegisterName},
+    {ok, Pid} = gen_statem:start_link({local, RegisterName}, ?MODULE, [Config1], []),
+    {ok, Pid, Config1}.
 
-%% @equiv set_exporter(Exporter, [])
+%% @deprecated Please use {@link otel_tracer_provider}
```
ferd, line 76: worth using the module directive for these as well I think

apps/opentelemetry/src/otel_simple_processor.erl
```diff
@@ -52,18 +55,37 @@
...
 -spec set_exporter(module(), term()) -> ok.
 set_exporter(Exporter, Options) ->
-    gen_statem:call(?MODULE, {set_exporter, {Exporter, Options}}).
+    gen_statem:call(?REG_NAME(global), {set_exporter, {Exporter, Options}}).
+
+%% @deprecated Please use {@link otel_tracer_provider}
+-spec set_exporter(atom(), module(), term()) -> ok.
+set_exporter(Name, Exporter, Options) ->
+    gen_statem:call(?REG_NAME(Name), {set_exporter, {Exporter, Options}}).
```
ferd, line 88: Given the definition:
```erlang
-define(REG_NAME(Name), list_to_atom(lists:concat([?MODULE, "_", Name]))).
```
Since the type spec specifies an `atom()` for `Name` this `lists:concat/1` call is bound to fail. Something does not work here.

The macro needs to change to have a change to a list first?

apps/opentelemetry/src/otel_tracer_server.erl
```diff
@@ -117,6 +127,53 @@ update_force_flush_error(Reason, ok) ->
...
+        _ ->
+            init_processors_(SpanProcessorSup, Rest)
+    end.
+
+init_processor(SpanProcessorSup, ProcessorModule, Config) ->
+    %% start_link is an optional callback for processors
+    case lists:member({start_link, 1}, ProcessorModule:module_info(exports)) of
+        true ->
+            try
+                case supervisor:start_child(SpanProcessorSup, [ProcessorModule, Config]) of
```
ferd, line 161: you can use `try <exp> of ...` here to skip a nested `case`, and it will avoid capturing errors within the matching clauses which are our code.

apps/opentelemetry_api/include/opentelemetry.hrl
```diff
@@ -13,6 +13,12 @@
 %% limitations under the License.
 %%%------------------------------------------------------------------------
 
+%% convert 2 strings (one meant to be like a namspace and the other a unique name)
+%% to an atom for use registering a process
+-define(REG_NAME(Name), list_to_atom(lists:concat([?MODULE, "_", Name]))).
```
ferd, line 18: this here thing is used in code with atoms?

apps/opentelemetry_api/src/otel_tracer_provider.erl
```diff
@@ -21,56 +21,78 @@
...
+        gen_server:call(Server, force_flush)
     catch exit:{noproc, _} ->
             %% ignore because likely no SDK has been included and started
             ok
     end.
+
+%%
+
+maybe_to_reg_name(Name) when is_atom(Name) ->
+    ?REG_NAME(Name);
```
ferd, line 94: atom here, macro expects strings.

ferd (changes requested):
There are a few minor improvements and one major inconsistency I picked up.

apps/opentelemetry_api/src/otel_tracer_provider.erl
tsloughter: The macro expects anything that works in `lists:concat` which atoms do.

apps/opentelemetry_api/include/opentelemetry.hrl
tsloughter: Yea, `lists:concat` works with strings, atoms, ints, etc

apps/opentelemetry/src/otel_batch_processor.erl
tsloughter: Ah yup, meant to look that up, thought it existed.

tsloughter: And it is deprecated for the whole API in general. We need to get rid of its usage by users (mostly done in tests) because it now requires a bad hack to work -- there is a special "global" name for a processor if it is the only processor on the global tracer provider.

apps/opentelemetry/src/otel_tracer_server.erl
tsloughter: Updated.

apps/opentelemetry_api/src/otel_tracer_provider.erl
ferd: ah, I had no idea lists:concat handled atoms. Alright then.

ferd (approved):

apps/opentelemetry_api/src/otel_tracer_provider.erl
tsloughter: Yea, I learned it from reading OTP logger code where it is used this way :)

PR #490, opened 2022-11-09
feat: add elixir semantic conventions

yordis:


tsloughter:
Thanks! Looks good except the Erlang version uses atoms for the keys and the Elixir one should do the same.

I'm not positive atoms are the best choice, since they have to later be converted anyway, but it saves memory at least. @bryannaegele might have an opinion?

Could you also add to the README how the user would add `use OpenTelemetry.Semantic...` to use the macros in their code?

tsloughter (approved):

PR #493, opened 2022-11-17
register Instruments with meter provider when created

tsloughter:
when an Instrument is added it is matched against existing Views to create a ViewAggregation for each match (and for each Reader).

ViewAggregations table is updated to be a bag to simplify the updating of Instrument to ViewAggregation mapping. Now it is just an insert for the newly matched Views if a View is added and matches against existing Instruments.


tsloughter:
Closing. I want to include renaming the functions to make clear a side effect is happening.

PR #495, opened 2022-11-21
fix the Erlang header file generated in semconv app

tsloughter:


ferd (approved):
lmao wow I didn't see that go through on review. something about how characters look vs expectations?

PR #496, opened 2022-11-21
Metrics macro api

tsloughter:
This is inefficient how it looks up the instrument and then sends a message with the measurement, but I wanted to get this done in multiple PRs instead of one giant one.

So for now the way the macros work is an instrument is looked up by its Meter+Name and then a call to `record` sends a message to the `otel_meter_server` as before.

Another difference is the Instruments are registered with `otel_meter_server` when created instead of matching with Views when a recording is done. This allows for Views to be added dynamically.

### DIFF
--- a/apps/opentelemetry_api_experimental/lib/open_telemetry/counter.ex
+++ b/apps/opentelemetry_api_experimental/lib/open_telemetry/counter.ex
@@ -1,0 +1,26 @@
+defmodule OpenTelemetryAPIExperimental.Counter do
+  @moduledoc """
+
+  """
+
+  defmacro create(name, value_type, opts) do
+    quote bind_quoted: [name: name, value_type: value_type, opts: opts] do
+      :otel_meter.create_counter(
+        :opentelemetry_experimental.get_meter(__MODULE__),
+        name,
+        value_type,
+        opts
+      )
+    end
+  end
+
+  defmacro add(name, number, attributes) do
+    quote bind_quoted: [name: name, number: number, attributes: attributes] do
+      :otel_counter.add(
+        :otel_meter.lookup_instrument(:opentelemetry_experimental.get_meter(__MODULE__), name),
+        number,
+        attributes
+      )
+    end
+  end
+end
--- a/apps/opentelemetry_api/src/opentelemetry.erl
+++ b/apps/opentelemetry_api/src/opentelemetry.erl
@@ -455,7 +455,7 @@ verify_and_set_term(Module, TermKey, Behaviour) ->
             true;
         false ->
             ?LOG_WARNING("Module ~p does not exist. "
-                         "A noop ~p will be used until a module is configured.",
+                         "A noop ~p will be used.",
                          [Module, Behaviour]),
             false
     end.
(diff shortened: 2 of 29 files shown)


apps/opentelemetry_api_experimental/lib/open_telemetry/counter.ex
```diff
@@ -0,0 +1,26 @@
...
+  defmacro add(name, number, attributes) do
+    quote bind_quoted: [name: name, number: number, attributes: attributes] do
+      :otel_counter.add(
+        :otel_meter.lookup_instrument(:opentelemetry_experimental.get_meter(__MODULE__), name),
+        number,
+        attributes
+      )
+    end
+  end
+end
```
yordis, line 26:
```suggestion
defmodule OpenTelemetryAPIExperimental.Counter do
  @moduledoc """

  """

  defmacro create_counter(name, value_type, opts) do
    quote bind_quoted: [name: name, value_type: value_type, opts: opts] do
      :otel_meter.create_counter(
        :opentelemetry_experimental.get_meter(__MODULE__),
        name,
        value_type,
        opts
      )
    end
  end

  defmacro add_counter(name, number, attributes) do
    quote bind_quoted: [name: name, number: number, attributes: attributes] do
      :otel_counter.add(
        :otel_meter.lookup_instrument(:opentelemetry_experimental.get_meter(__MODULE__), name),
        number,
        attributes
      )
    end
  end
end
```

I prefer when the functions are self-sufficient and treat the module as a namespace that I could often ignore together.

:2cents: but 🤷🏻 whatever works, you get used to it.

tsloughter: Hm, interesting point. I guess we should go with what is most common in Elixir libraries. I'll go look at some like prometheus.ex and telemetry metrics.

yordis: Except for functions like `new` where you create creating structs out of the module (so it does matter a lot the name of the module), I think it is a typical pattern for the most part.

The biggest concern is when you require to `import` functions since the more generic you go, the harder it is because things are too generic. Or when you have a macro that also expands things in the scope.

I tend to avoid all together functions that are incapable of being self-sufficient. At some point, collisions will happen, and in the worst case, I can not even control it because it is in a 3rd party package ... so it becomes tricky to figure out the right solution.

Also, it is inlined with what the erlang package is doing. So it would be nice to avoid that much translation.

ferd (approved):
Seems good to me. Note that I haven't spent time checking out the Elixir-side of macro naming since there's an ongoing conversation there, and I mostly validated the Erlang side of things.

apps/opentelemetry_api_experimental/lib/open_telemetry/counter.ex
tsloughter: I'm going to merge this but not consider this issue settled. I'd like more input before making a decision but don't want to hold up this PR , since its not being released, and even if it was is still experimental :)

yordis: 🚀 🚀 🚀

PR #501, opened 2022-11-25
Export otel_ctx token

bullno1:
`otel_ctx:token/0` should be exported as it is a return type and it can be saved in some state record.

### DIFF
--- a/apps/opentelemetry/rebar.config
+++ b/apps/opentelemetry/rebar.config
@@ -1,5 +1,5 @@
 {erl_opts, [debug_info]}.
-{deps, [{opentelemetry_api, "~> 1.1"}]}.
+{deps, [{opentelemetry_api, "~> 1.2.0-beta"}]}.
 
 {profiles,
  [{docs, [{edoc_opts,
--- a/apps/opentelemetry_api/src/opentelemetry_api.app.src
+++ b/apps/opentelemetry_api/src/opentelemetry_api.app.src
@@ -1,6 +1,6 @@
 {application, opentelemetry_api,
  [{description, "OpenTelemetry API"},
-  {vsn, "1.1.1"},
+  {vsn, "1.2.0-beta"},
   {registered, []},
   {applications,
    [kernel,
--- a/apps/opentelemetry_exporter/README.md
+++ b/apps/opentelemetry_exporter/README.md
@@ -43,7 +43,7 @@ config :opentelemetry_exporter,
   otlp_headers: [{"x-honeycomb-dataset", "experiments"}]
 ```
 
-Furthermore, you need to configure Elixir apps to load `opentelemetry_exporter` before `opentelemtry`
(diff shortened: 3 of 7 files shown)


apps/opentelemetry_api/src/otel_propagator_text_map.erl
```diff
@@ -146,16 +146,16 @@ inject_from(Context, Module, Carrier, CarrierSetFun) when is_atom(Module) ->
 inject_from(Context, {Module, Options}, Carrier, CarrierSetFun) ->
      Module:inject(Context, Carrier, CarrierSetFun, Options).
 
--spec extract(otel_propagator:carrier()) -> otel_ctx:t().
+-spec extract(otel_propagator:carrier()) -> otel_ctx:token().
```
tsloughter, line 149: Good catch. But I think `extract` should still return a Context, so to be correct it should return the Context instead of the result of `attach` in `extract_to`.

apps/opentelemetry_api/src/otel_ctx.erl
```diff
@@ -45,9 +45,10 @@
 -type value() :: term().
 
 %% at this time the "token" is actually the context map itself
--type token() :: term().
+-opaque token() :: t().
```
tsloughter, line 48: I wanted to keep `token` as `term` since it could change in the future and be anything. Opaque will allow that for external use but it still means changing the type if we change it to an actual token of sorts ever. I don't know, I guess I'm token on this one.

tsloughter (approved):

apps/opentelemetry_api/src/otel_propagator_text_map.erl
```diff
@@ -159,7 +159,8 @@ extract(Propagator, Carrier) ->
 extract(Propagator, Carrier, CarrierKeysFun, CarrierGetFun) ->
     Context = otel_ctx:get_current(),
     Context1 = extract_to(Context, Propagator, Carrier, CarrierKeysFun, CarrierGetFun),
-    otel_ctx:attach(Context1).
+    otel_ctx:attach(Context1),
+    Context1.
```
tsloughter, line 163: Was about to merge this when I noticed this is now returning `Context1` when before wasn't it returning `Context`?

I guess it was never under test. 

I can't remember if it should return the previous Context (since the new one is in the pdict) or the new Context...

tsloughter: Technically this is a breaking API change to return `Context1` instead of `Context`, so I think we need to stick to `Context`.

tsloughter, line 163:
```suggestion
    Context.
```

tsloughter: Here I go again.. now I realize if it is meant to return something to reattach an old context then it should be the token, the result of `attach` and your original change to update the type spec is correct! haha, sorry about all this

tsloughter (approved):

PR #502, opened 2022-11-27
Use the same ets tables for all Readers

tsloughter:
These patches remove the per-Reader ETS tables and instead puts a Reader reference in the ViewAggregation and adds the Reader reference as part of the key in the Metrics table.

Now with just the Meter and name of an instrument a measurement can be taken by looking up the Metrics to update in the ViewAggregation table and then updating the Metrics table.

So 2 tables are hit for a measurement taken.


apps/opentelemetry_experimental/src/otel_meter_server.erl
```diff
@@ -210,23 +229,26 @@ handle_call({get_meter, Scope}, _From, State=#state{shared_meter=Meter}) ->
...
-handle_cast({record, Measurement}, State=#state{readers=Readers}) ->
-    _ = handle_measurement(Measurement, Readers),
+handle_cast(_, State) ->
     {noreply, State}.
 
 %% TODO: Uncomment when we can drop OTP-23 support
 %% handle_info({'DOWN_READER', Ref, process, _Pid, _} , State=#state{readers=Readers}) ->
 handle_info({'DOWN', Ref, process, _Pid, _} , State=#state{readers=Readers}) ->
+    %% TODO: remove ViewAggregation and Metrics tab entries
+    %% Or better, re-use them if/when the Reader restarts.
```
ferd, line 251: any risks in not removing these?

ferd (approved):

apps/opentelemetry_experimental/src/otel_meter_server.erl
tsloughter: Oops, that is an old comment. Now that each reader gets a `reference` which it reuses we don't want to remove these entries because they are reused.

ferd (approved):

apps/opentelemetry_experimental/src/otel_aggregation_histogram_explicit.erl
```diff
@@ -59,35 +59,57 @@ aggregate(Table, Key, Value) ->
...
-                                                                     sum=Sum+Value};
-                               (H=#explicit_histogram_aggregation{key=K,
-                                                                  sum=Sum}) when K =:= Key ->
-                                    H#explicit_histogram_aggregation{bucket_counts=BucketCounts,
-                                                                     sum=Sum+Value}
-                            end),
+
+            %% since we need the Key in the MatchHead for the index to be used we
+            %% can't use `ets:fun2ms' as it will shadow `Key' in the `fun' head
+            MS = [{{explicit_histogram_aggregation,Key,'_','_','_','_','_','$1','$2','$3'},
```
ferd, line 65: scarier (maybe?) option:
```erlang
[{Head#explicit_histogram_aggregation{key=Key}, Guards, Body}
  || {Head, Guards, Body} <- ets:fun2ms(....)]
```

tsloughter: I like it, but because the fun is so large I think it makes it less readable in the end. Hard to decide.

PR #505, opened 2022-12-02
bump api, sdk, exporter versions to beta releases

tsloughter:
Because of the large changes to support multiple tracer providers and batch processors I think it makes sense to release as a beta first.

The only concern I have is getting people to use it so we can then release the stable releases..


ferd (approved):
Yeah I'd like to reach a point with my side project where I'm using this, but I may not get there until the winter break in all likelihood.

Issue #510, opened 2022-12-06
tracestate propagation


bullno1:
It seems that the tracestate created by a sampler is not propagated into a span context.

This can be seen in: https://github.com/open-telemetry/opentelemetry-erlang/blob/main/apps/opentelemetry/src/otel_span_utils.erl#L75

`NewSpanCtx#span_ctx` should also adopt the `TraceState` returned from the sample.

Unless I misunderstand the semantic of tracestate, I think it should be propagated from parent to children.

If this is confirmed to be a bug, I'll submit a PR.

PR #511, opened 2022-12-06
Assign tracestate from sampler to new span

bullno1:
Fix https://github.com/open-telemetry/opentelemetry-erlang/issues/510


tsloughter (approved):
Thanks!

PR #514, opened 2022-12-06
pass name, not instrument, in metric macros to record measurements

tsloughter:
The existing tests using the macros cover this change to just using the name to take measurements.

I'm undecided on whether to remove the old API, like `otel_counter:add/3`, and only support use by name or to keep it and have an interface you can do either by name or by instrument.


tsloughter:
Note I'm going to open a PR to move the `value_type` check to the aggregator, so that won't be lost when using the name only API instead of passing the Instrument.

ferd (approved):

PR #516, opened 2022-12-07
remove value_type from instruments

### LINKED ISSUE #515: Metrics: Get rid of ValueType?
Currently an Instrument has a `value_type` property of `integer` or `float`.

This is useful in a sense because if a counter or updown counter is an integer we can use `ets:update_counter` instead of `ets:select_replace`.

But since the metrics API will be focused around passing the Instrument atom name only and not the Instrument record we will not have that record to know if the user defined it as an integer or float type.

One option is to use `update_counter` if the value passed by the user to `add` is an integer. The issue with this is the value in the ETS table may be a float, so `update_counter` will still fail. If removing `value_type` is deemed worth it we could either document that if you add with a float and then an integer on the same instrument you will be dropping those values.

tsloughter:
Closes #515


tsloughter:
Actually dealing with losing values can be resolved by a strategy used in deadtrickster's counters library https://github.com/deadtrickster/counters.erl/blob/master/src/counters/counters_counter.erl -- use a field for integers and a field for floats and then add them when collected.

apps/opentelemetry_experimental/src/otel_aggregation_sum.erl
```diff
@@ -47,36 +48,80 @@ aggregate(Tab, Key, Value, _Options) ->
...
+          {#sum_aggregation{key='$1',
+                            start_time_unix_nano='$2',
+                            checkpoint='_',
+                            int_value='$3',
+                            float_value='$4'},
+           [{'=:=', {element, 1, '$1'}, {const, Name}},
+            {'=:=', {element, 3, '$1'}, {const, ReaderPid}}],
+           [{#sum_aggregation{key='$1',
+                              start_time_unix_nano='$2',
+                              checkpoint={'+', '$3', '$4'},
```
ferd, line 122: Oh okay that's how we avoid dropping data. If there is a non-default floating point value, we coerce the whole thing to also be a float and give it [implicit] limits in accuracy. Clever, works nice with the previous clause guarding it so int-only values remain safe at all sizes.

apps/opentelemetry_experimental/test/otel_metrics_SUITE.erl
```diff
@@ -239,34 +324,32 @@ provider_test(_Config) ->
...
-                             value_type = ValueType,
                              unit = CounterUnit}, Counter),
 
     ?assertEqual(ok, otel_counter:add(Counter, 2, #{<<"c">> => <<"b">>})),
     ?assertEqual(ok, otel_counter:add(Counter, 3, #{<<"a">> => <<"b">>, <<"d">> => <<"e">>})),
     ?assertEqual(ok, otel_counter:add(Counter, 4, #{<<"c">> => <<"b">>})),
     ?assertEqual(ok, otel_counter:add(Counter, 5, #{<<"c">> => <<"b">>})),
 
-    %% this measurement will be ignored because counter is of type integer
+    %% converts counter to a float valuer
```
ferd, line 343:
```suggestion
    %% converts counter to a float value
```

ferd (approved):

PR #519, opened 2022-12-15
otel_span:set_status support for status code

Kuroneer:
The ?set_status macro in otel_tracer.hrl indicates that otel_span:set_status/1 allows you to provide both a opentelemetry:status() or a opentelemetry:status_code()

This commit fixes the opentelemetry:status_code() case

Also, since the typing already supports undefined, translate it to UNSET

### DIFF
--- a/apps/opentelemetry_api/src/otel_span.erl
+++ b/apps/opentelemetry_api/src/otel_span.erl
@@ -288,16 +288,18 @@ record_exception(SpanCtx, Class, Term, Message, Stacktrace, Attributes) when is_
 record_exception(_, _, _, _, _, _) ->
     false.
 
--spec set_status(SpanCtx, Status) -> boolean() when
-      Status :: opentelemetry:status() | undefined,
+-spec set_status(SpanCtx, StatusOrCode) -> boolean() when
+      StatusOrCode :: opentelemetry:status() | undefined | opentelemetry:status_code(),
       SpanCtx :: opentelemetry:span_ctx().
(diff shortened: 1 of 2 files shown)


tsloughter (approved):
Thanks!

Kuroneer:
> @Kuroneer can you sign the CLA?

I have raised a question to my employer to see if I can sign it as an individual contributor to avoid the bureaucracy. I doubt there'll be any problem, and even less with a patch of this size, but I'd rather err on the safe side.

Kuroneer:
> @Kuroneer can you sign the CLA?

Done

Kuroneer:
IMHO, the priority logic should be in only one place (in this case, the sdk), it makes no sense to implement it both in the API and the SDK. The SDK already limits the transitions to UNSET -> * and ERROR -> OK.

The documentation says that it should be ignored, and if I read it correctly, the SDK just discards the invalid transitions, so I think the behaviour matches the documentation.

Regarding this change, I'd rather had limited the atoms to those included in the lib, but as per the `spec`, `undefined` was already tolerated. As a user of the lib, my expectation would be to have `undefined` == `UNSET` to avoid introducing breaking changes.

Kuroneer:
IMHO, the priority logic should be in only one place (in this case, the sdk), it makes no sense to implement it both in the API and the SDK (which already limits the transitions to the valid ones only). 

The documentation says that it should be ignored, and if I read the code correctly, the SDK just discards the invalid transitions, so I think the behaviour matches the documentation.

Regarding this change, I'd rather had limited the atoms to those included in the lib, but as per the `spec`, `undefined` was already tolerated. As a user of the lib, my expectation would be to have `undefined` == `UNSET` to avoid introducing breaking changes.

PR #531, opened 2022-12-28
Store keys for otel_resource attributes as atoms

Kuroneer:
Imported attribute keys (from opentelemetry_semantic_conventions) are atoms, so otel_resource should support them.
Also, as otel_resource_detector shouldn't know the type of the key, add an indirection level for the key check.

### DIFF
--- a/apps/opentelemetry/src/otel_resource.erl
+++ b/apps/opentelemetry/src/otel_resource.erl
@@ -22,9 +22,10 @@
          create/2,
          merge/2,
          schema_url/1,
-         attributes/1]).
+         attributes/1,
+         is_key/2]).
 
--type key() :: unicode:latin1_binary().
+-type key() :: unicode:latin1_binary() | atom().
 %% values allowed in attributes of a resource are limited
 -type value() :: unicode:latin1_binary() | integer() | float() | boolean().
 -type schema_url() :: uri_string:uri_string().
@@ -48,9 +49,14 @@ create(Map, SchemaUrl) when is_map(Map) ->
 create(List, SchemaUrl) when is_list(List) ->
     List1 = lists:filtermap(fun({K, V}) ->
                                     %% TODO: log an info or debug message when dropping?
-                                    case check_key(K) andalso check_value(V) of
-                                        {true, Value} ->
-                                            {true, {to_binary(K), Value}};
+                                    case try_check_key(K, true) of
+                                        {true, Key} ->
+                                            case check_value(V) of
+                                                {true, Value} ->
+                                                    {true, {Key, Value}};
+                                                _ ->
+                                                    false
+                                            end;
                                         _ ->
                                             false
                                     end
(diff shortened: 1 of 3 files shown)


apps/opentelemetry/src/otel_resource.erl
```diff
@@ -48,11 +49,18 @@ create(Map, SchemaUrl) when is_map(Map) ->
 create(List, SchemaUrl) when is_list(List) ->
     List1 = lists:filtermap(fun({K, V}) ->
                                     %% TODO: log an info or debug message when dropping?
-                                    case check_key(K) andalso check_value(V) of
-                                        {true, Value} ->
-                                            {true, {to_binary(K), Value}};
+                                    try check_key(K) of
+                                        {true, Key} ->
```
tsloughter, line 53: It is better to keep the Key as an atom if it is provided as one.

There is the issue of missing duplicates but we should be phasing out use of non-atom keys everywhere to avoid this. Meaning examples and what not so users don't use non-atom keys.

tsloughter:
Thanks! I had one comment about wanting to keep atom keys as atoms to save memory and copying, but besides that this is great.

apps/opentelemetry/src/otel_resource.erl
Kuroneer: I did not know if the attributes were required to be binary for marshalling, so I tried to keep them that way (I agree with favoring atoms when possible). Tomorrow I'll update the PR

Kuroneer: Should I keep the `atom_to_binary(X, latin1)` transformation to check for utf8 here or it'd be an acceptable error during the marshalling (or even to be transmitted as utf8 through the wire and leave the problem to the collector)?

tsloughter: Hm, good question. For keys I think they do have to be dropped if not latin1. So I guess so. Not a performance issue for resource sice its created once at startup, but then copied around a lot.

Kuroneer:
I'm not entirely convinced about the `binary_to_existing_atom` in 0d20b0b398aeecaad26cdd6ced6f066f4d87c82e, but otherwise env vars (strings) may not be merged with the ones from semantic conventions.

A different approach would be to unconditionally create atoms from the keys in the OTEL env vars, but I'm hesitant to go that way.

tsloughter:
Should just use `binary_to_atom`. This is something that runs once on boot. If a node is taken down from atom exhaustion then at least it happens at boot time ;)

apps/opentelemetry/src/otel_resource.erl
```diff
@@ -70,6 +76,17 @@ attributes(#resource{attributes=Attributes}) ->
 attributes(_) ->
     undefined.
 
+-spec attributes_is_key(key(), t()) -> boolean().
```
tsloughter, line 79: I think this would be better has just `is_key`.
```suggestion
-spec is_key(key(), t()) -> boolean().
```

tsloughter: I'll merge and cut a new release once that is done.

Kuroneer:
Sorry for the rebase, but that way I squash all the changes and update it with main

apps/opentelemetry/src/otel_resource.erl
Kuroneer: Done

tsloughter (approved):
Thanks!

PR #537, opened 2023-01-12
Using opentelemetry_semantic_conventions for record_exception

wingyplus:
Use constant from opentelemetry_semantic_conventions in record_exception instead of binary (or string in Elixir). This changes apply to both Erlang and Elixir API.


tsloughter (commented):
Thanks!

tsloughter (approved):

PR #540, opened 2023-01-16
Have set_status check is_recording in all cases

Kuroneer:
Very minor: ?is_recording macro is missing when setting status as undefined (noticed when reading the code again from https://github.com/open-telemetry/opentelemetry-erlang/pull/519)

### DIFF
--- a/apps/opentelemetry_api/src/otel_span.erl
+++ b/apps/opentelemetry_api/src/otel_span.erl
@@ -297,7 +297,7 @@ set_status(SpanCtx=#span_ctx{span_sdk={Module, _}}, Code) when ?is_recording(Spa
                                                                 Code =:= ?OTEL_STATUS_OK orelse
                                                                 Code =:= ?OTEL_STATUS_ERROR)->
     Module:set_status(SpanCtx, opentelemetry:status(Code));
-set_status(SpanCtx=#span_ctx{span_sdk={Module, _}}, undefined) ->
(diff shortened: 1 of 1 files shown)


apps/opentelemetry_api/src/otel_span.erl
```diff
@@ -297,8 +297,8 @@ set_status(SpanCtx=#span_ctx{span_sdk={Module, _}}, Code) when ?is_recording(Spa
                                                                 Code =:= ?OTEL_STATUS_OK orelse
                                                                 Code =:= ?OTEL_STATUS_ERROR)->
     Module:set_status(SpanCtx, opentelemetry:status(Code));
-set_status(SpanCtx=#span_ctx{span_sdk={Module, _}}, undefined) ->
-    Module:set_status(SpanCtx, opentelemetry:status(?OTEL_STATUS_UNSET));
+set_status(SpanCtx, undefined) ->
```
tsloughter, line 300: Thanks! But lets just do the `is_recording` check here instead of first having to construct the `status` record.

Kuroneer: Sure thing, I thought it was more readable the other way around (and I doubt it would have had any meaningful impact)

tsloughter: Yea, impact is basically nothing, I just figure its a good habit for a library like this -- unless it hurts readability a lot.

tsloughter (approved):

PR #542, opened 2023-01-26
add support for running eqwalizer

tsloughter:


rebar.config
```diff
@@ -1,5 +1,9 @@
 {erl_opts, [debug_info]}.
-{deps, []}.
+{deps, [{eqwalizer_support,
+         {git_subdir,
+          "https://github.com/whatsapp/eqwalizer.git",
+          {branch, "main"},
+          "eqwalizer_support"}}]}.
```
ferd, line 6: This is possibly going to prevent us from publishing packages since it'll have a git dep in the lockfile.

You may want to wall this off under a non-default profile?

tsloughter: Na, currently publishing packages by cd'ing into each app -- since some are published with `mix` and some with `rebar3`.

If it was all rebar3 I'd be publishing from the top level. I suppose I could do that for the ones that are rebar3, but just haven't bothered to.

ferd (approved):

PR #549, opened 2023-03-04
use unique httpc profile per otlp exporter

tsloughter:


tsloughter:
I'm unsure if this is the appropriate strategy -- both how httpc profiles are done and the use of inet6fb4 by default -- so looking for feedback.

The only time ipv6 would be used in a possibly unexpected manner is if the user gives a domain for the collector that both resolves successfully to an ipv6 address and it is able to successfully open a connection to that address.

If for some reason a user really wants they can override with `http_options`.

apps/opentelemetry_exporter/src/opentelemetry_exporter.erl
```diff
@@ -198,23 +199,37 @@ init(Opts) ->
...
                         headers=Headers,
                         compression=Compression,
                         protocol=http_json}}
     end.
 
+%% use a unique httpc profile per exporter
+start_httpc(Opts) ->
+    HttpcProfile = list_to_atom(lists:concat([?MODULE, "_", erlang:pid_to_list(self())])),
+    %% by default use inet6fb4 which will try ipv6 and then fallback to ipv4 if it fails
+    HttpcOptions = lists:ukeymerge(1, maps:get(httpc_options, Opts, []), [{ipfamily, inet6fb4}]),
```
ferd, line 223:
```suggestion
    HttpcOptions = lists:ukeymerge(1, 
        lists:usort(maps:get(httpc_options, Opts, [])),
        [{ipfamily, inet6fb4}]
    ),
```
for `lists:ukeymerge/1` to work fine, both input lists need to have unique keys and be sorted.

ferd (changes requested):
Had a small change to request. May be worth considering that on many providers that do not support IPv6, there may be a perf cost to first trying and waiting to fail on DNS resolution with ipv6 before falling back to ipv4. Making it overridable is fine though.

apps/opentelemetry_exporter/src/opentelemetry_exporter.erl
tsloughter: Ah right! Thanks

tsloughter:
I figure perf cost isn't an issue since connection should just be done once on startup. But yea, if someone does end up on a system that keeps closing the socket and it has to be reopened then it'll be good they can override it.

ferd (approved):

Issue #517, opened 2022-12-09
Add general and source code attributes to spans


tsloughter:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/span-general.md#source-code-attributes

bullno1:
I kinda need this too.

This is only possible in macro right?

tsloughter:
Correct, it would need to be done in the macros for `with_span` and `start_span`.

bullno1:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/span-general.md#general-thread-attributes

On the topic of `thread.id` attribute, `erlang:system_info(scheduler_id)` is suggested.
While this is kind of correct, I'm not sure how useful it is compared to say: `self()`.
But formatting `self()` can have some overhead that is not always desirable.

bullno1:
On the same topic `code.function` also requires some string/binary construction from `?FUNCTION_NAME` and `?FUNCTION_ARITY`.
It's probably negligible but it's not nothing, even in the non-recording case.

There are 2 ways to go about this:

* Just format anyway.
* Introduce some sort of "lazy attributes" that's only evaluated by the span processor. `{'$lazy', fun()}`?

tsloughter:
I think just format (`lists:concat(?FUNCTION_NAME, "/", ?FUNCTION_ARITY)`) it anyway and we should look at proposing `?FUNCTION_NAME_STRING` that includes arity to OTP.

PR #550, opened 2023-03-04
Add source code attributes to Elixir tracer macros

Cantido:
This change adds thread & source code attributes spans created with the Elixir API, as described here: <https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/span-general/#general-thread-attributes>. This is related to #517, but it does not close it since I wasn't able to do the Erlang part (I don't know Erlang). I can certainly do the Elixir part though :smile:.

I tested this in a project with Jaeger, and the attributes were all correct, despite my lack of skill with macros.

### DIFF
--- a/apps/opentelemetry_api/lib/open_telemetry/tracer.ex
+++ b/apps/opentelemetry_api/lib/open_telemetry/tracer.ex
@@ -48,8 +48,7 @@ defmodule OpenTelemetry.Tracer do
     quote bind_quoted: [
             name: name,
             start_opts: opts,
-            code_attrs: code_attrs,
-            thread_id_key: thread_id_key
+            code_attrs: code_attrs
           ] do
       start_opts =
         Map.new(start_opts)
--- a/apps/opentelemetry_api/lib/open_telemetry/tracer.ex
+++ b/apps/opentelemetry_api/lib/open_telemetry/tracer.ex
@@ -28,7 +28,6 @@ defmodule OpenTelemetry.Tracer do
         end
 
       source_attrs = %{
-        Conventions.thread_id() => :erlang.system_info(:scheduler_id),
         Conventions.code_function() => code_function,
         Conventions.code_namespace() => __CALLER__.module,
         Conventions.code_filepath() => __CALLER__.file,
@@ -44,10 +43,12 @@ defmodule OpenTelemetry.Tracer do
   """
   defmacro start_span(name, opts \\ quote(do: %{})) do
     code_attrs = code_attributes() |> Macro.escape()
+    thread_id_key = Conventions.thread_id()
 
-    quote bind_quoted: [name: name, start_opts: opts, code_attrs: code_attrs] do
+    quote bind_quoted: [name: name, start_opts: opts, code_attrs: code_attrs, thread_id_key: thread_id_key] do
       start_opts =
         Map.new(start_opts)
+        |> Map.put(thread_id_key, :erlang.system_info(:scheduler_id))
         |> Map.update(:attributes, code_attrs, &Map.merge(&1, code_attrs))
 
       :otel_tracer.start_span(
@@ -65,10 +66,12 @@ defmodule OpenTelemetry.Tracer do
   """
   defmacro start_span(ctx, name, opts) do
     code_attrs = code_attributes() |> Macro.escape()
+    thread_id_key = Conventions.thread_id()
 
-    quote bind_quoted: [ctx: ctx, name: name, start_opts: opts, code_attrs: code_attrs] do
+    quote bind_quoted: [ctx: ctx, name: name, start_opts: opts, code_attrs: code_attrs, thread_id_key: thread_id_key] do
       start_opts =
         Map.new(start_opts)
+        |> Map.put(thread_id_key, :erlang.system_info(:scheduler_id))
         |> Map.update(:attributes, code_attrs, &Map.merge(&1, code_attrs))
 
       :otel_tracer.start_span(
@@ -104,11 +107,13 @@ defmodule OpenTelemetry.Tracer do
   """
   defmacro with_span(name, start_opts \\ quote(do: %{}), do: block) do
     code_attrs = code_attributes() |> Macro.escape()
+    thread_id_key = Conventions.thread_id()
 
     quote do
       code_attrs = unquote(code_attrs)
       start_opts =
         Map.new(unquote(start_opts))
+        |> Map.put(unquote(thread_id_key), :erlang.system_info(:scheduler_id))
         |> Map.update(:attributes, code_attrs, &Map.merge(&1, code_attrs))
 
       :otel_tracer.with_span(
@@ -129,11 +134,13 @@ defmodule OpenTelemetry.Tracer do
   """
   defmacro with_span(ctx, name, start_opts, do: block) do
     code_attrs = code_attributes() |> Macro.escape()
+    thread_id_key = Conventions.thread_id()
 
     quote do
       code_attrs = unquote(code_attrs)
       start_opts =
         Map.new(unquote(start_opts))
+        |> Map.put(unquote(thread_id_key), :erlang.system_info(:scheduler_id))
         |> Map.update(:attributes, code_attrs, &Map.merge(&1, code_attrs))
 
       :otel_tracer.with_span(
--- a/apps/opentelemetry_api/lib/open_telemetry/tracer.ex
+++ b/apps/opentelemetry_api/lib/open_telemetry/tracer.ex
@@ -16,12 +16,39 @@ defmodule OpenTelemetry.Tracer do
       end
   """
 
+  require OpenTelemetry.SemanticConventions.Trace
+  alias OpenTelemetry.SemanticConventions.Trace, as: Conventions
+
+  defmacrop insert_code_attributes(opts) do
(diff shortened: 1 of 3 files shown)


apps/opentelemetry_api/lib/open_telemetry/tracer.ex
```diff
@@ -16,12 +16,39 @@ defmodule OpenTelemetry.Tracer do
...
+      source_attrs = %{
+        Conventions.thread_id() => :erlang.system_info(:scheduler_id),
+        Conventions.code_function() => code_function,
+        Conventions.code_namespace() => __CALLER__.module,
+        Conventions.code_filepath() => __CALLER__.file,
+        Conventions.code_lineno() => __CALLER__.line
+      }
+
+      Map.new()
+      |> Map.update(:attributes, source_attrs, &Map.merge(&1, source_attrs))
```
tsloughter, line 39:
```suggestion
      Map.update(:attributes, source_attrs, source_attrs)
```

Isn't this the same?

tsloughter: Oh, I misread entirely :)

tsloughter: At least I think... the `&1` applies to the `&Map.merge/2`, right? And no thte pipe?

But I realize I still don't see where the original opt attributes are being merged with.

tsloughter:
Nice!

Could you add a simple test? Could even just extend one of the existing tests in `test/otel_tests.exs` to check these attributes are included in addition to attributes it passes to a start.

apps/opentelemetry_api/lib/open_telemetry/tracer.ex
Cantido: whoopsie, you're absolutely right, I was toying with that while debugging and removed the original opt attributes.

Cantido:
How can I check the attributes set in the current span? I'm having trouble finding an example.

tsloughter:
This test adds attributes and then checks they are in the outputted span https://github.com/open-telemetry/opentelemetry-erlang/blob/main/test/otel_tests.exs#L33-L46 -- could just add to that test that the source code attributes are also there.

Cantido:
Do I need to do anything special to run the tests locally? When I try to run `mix test`, I get this:

```console
$ mix test
===> Analyzing applications...
===> Compiling opentelemetry
=WARNING REPORT==== 5-Mar-2023::12:19:44.385057 ===
FORMAT ERROR: "OTLP exporter module `opentelemetry_exporter` not found. Verify you have included the `opentelemetry_exporter` dependency." - [opentelemetry_exporter]
There are no tests to run
```

Cantido:
I also checked CI, and I ran `mix test --no-start test/otel_tests.exs` and got another error.

```console
$ mix test --no-start test/otel_tests.exs
===> Analyzing applications...
===> Compiling opentelemetry

== Compilation error in file test/otel_tests.exs ==
** (RuntimeError) error parsing file /home/rosa/workspace/opentelemetry-erlang/_build/test/lib/opentelemetry/src/otel_tracer.hrl, got: {:error, :enoent}
    (elixir 1.14.3) lib/record/extractor.ex:84: Record.Extractor.read_file/2
    (elixir 1.14.3) lib/record/extractor.ex:50: Record.Extractor.extract_record/2
    test/otel_tests.exs:12: (module)
    (elixir 1.14.3) lib/kernel/parallel_compiler.ex:449: Kernel.ParallelCompiler.require_file/2
    (elixir 1.14.3) lib/kernel/parallel_compiler.ex:342: anonymous fn/5 in Kernel.ParallelCompiler.spawn_workers/7
```

tsloughter:
You may need to first run `rebar3 as test compile`.

apps/opentelemetry_api/lib/open_telemetry/tracer.ex
```diff
@@ -88,12 +133,21 @@ defmodule OpenTelemetry.Tracer do
...
   """
   defmacro with_span(ctx, name, start_opts, do: block) do
+    code_attrs = code_attributes() |> Macro.escape()
+    thread_id_key = Conventions.thread_id()
+
     quote do
+      code_attrs = unquote(code_attrs)
+      start_opts =
+        Map.new(unquote(start_opts))
+        |> Map.put(unquote(thread_id_key), :erlang.system_info(:scheduler_id))
```
tsloughter, line 143: Isn't this adding the thread id attribute to `start_opts` instead of the `attributes`?

Also, is the `Map.new` needed? 

And I'd suggest moving the `thread_id` to `code_attributes()` so it doesn't have to be duplicated, and may mean this whole pipeline can be reduced to a `Map.update`. The fewer map creations/updates the better :)

Cantido: Good catch! I'll fix that. And I'll see how removing the `Map.new` turns out.

I moved `thread_id` out of `code_attributes` because I was worried that I was getting the compile-time scheduler instead of the runtime one, so I wanted to move it to inside the `quote` block.

tsloughter: Oh, the `__CALLER__` calls work because they are macros but `:erlang.system_info(:scheduler_id)` is going to be run at compile time?

Sounds right. I get confused with this macro stuff, esp when its macros in macros :)

Cantido: Yep! I'm trying to wrap my head around macros myself, I don't work with them much.

So I tried removing the `Map.new` and the tests failed, there was a test that passes start opts as a keyword list. Since `opts` is user input and this function is dealing with it as a map, I think it's a good defensive move.

Cantido:
I got the tests figured out! Now they're actually, y'know, testing this change. Thank you for the guidance :). This should be all set for another once-over.

tsloughter:
@Cantido hey, sorry, can you add a changelog entry for this and then I'll merge it?

Or actually, I'll go merge your one changing the macros to functions and then you'll want to rebase on that.

Cantido:
Sure thing!On Mar 21, 2023, at 4:05 PM, Tristan Sloughter ***@***.***> wrote:﻿
@Cantido hey, sorry, can you add a changelog entry for this and then I'll merge it?
Or actually, I'll go merge your one changing the macros to functions and then you'll want to rebase on that.

Cantido:
Sure thing!

tsloughter:
I'm stalling a little on this because I learned almost no language does this by default.

I'm leaning towards it still being good to do always since it is only part of the macro. Manual calls to the `otel_tracer` module to start a span will not add them because it can't even if we wanted it to.

apps/opentelemetry_api/lib/open_telemetry/tracer.ex
```diff
@@ -16,17 +16,51 @@ defmodule OpenTelemetry.Tracer do
...
+    thread_id_key = Conventions.thread_id()
+
+    quote bind_quoted: [
+            name: name,
+            start_opts: opts,
+            code_attrs: code_attrs,
+            thread_id_key: thread_id_key
+          ] do
+      code_attrs = Map.put(code_attrs, thread_id_key, :erlang.system_info(:scheduler_id))
+
```
tsloughter, line 55: I know you mentioned not thinking it should be the scheduler id but the pid before. I wonder if we shouldn't just leave out the thread attribute for now. I'd like to get the pid in but not sure what attribute key to use for it yet and don't want to have our hands tied by backwards compatibility if we add this.

tsloughter: @Cantido do you think you'll have time to update this PR to remove this?

Cantido: I can certainly do that!

apps/opentelemetry_api/lib/open_telemetry/tracer.ex
```diff
@@ -88,12 +144,22 @@ defmodule OpenTelemetry.Tracer do
   See `start_span/2` and `end_span/0`.
   """
   defmacro with_span(ctx, name, start_opts, do: block) do
+    code_attrs = code_attributes() |> Macro.escape()
+    thread_id_key = Conventions.thread_id()
+
     quote do
+      code_attrs =
+        Map.put(unquote(code_attrs), unquote(thread_id_key), :erlang.system_info(:scheduler_id))
```
tsloughter, line 152: Looks like thread_id is still in here?

Cantido: My bad, I haven't had a chance to properly sit down with this and get any context back. I promise I will get to that soon.

btkostner:
Is there any work I can do here to move this forward? I've been recently migrating a code base from Spandex to OTel and getting some code attributes for easy debugging has been rather painful. It'd be really nice to get this built in and supported everywhere.

bryannaegele:
It may need to be restarted. Not sure if anything has changed in the conventions.

Implementation - we can't depend on the semconv library in the API or SDK so the attributes need to be copied in and hard coded.

Cantido:
My apologies for letting this slide. I'd be happy to take some time to remove the semconv library and make the requested changes if I can find the time.

Cantido:
Closing, this is replaced by #808

PR #551, opened 2023-03-07
Elixir semcomv: use functions instead of macros

Cantido:
Macros are harder to use, since we must require the module. Requiring the module also pollutes the current context, which makes them less suitable for being used in other macros. The Elixir guide and hexdocs also both recommend against using macros unless absolutely necessary, see [Library Guidelines - Avoid Macros](https://hexdocs.pm/elixir/library-guidelines.html#avoid-macros). In this case, functions will work great!

If there are any performance concerns, we can inline the functions, and I believe the JIT compiler would inline these anyway because they're returning constants.

The only consequence of this change to users of this library will be compiler warnings saying that the `require` statement is unused, otherwise it will work exactly the same.

### DIFF
--- a/apps/opentelemetry_semantic_conventions/README.md
+++ b/apps/opentelemetry_semantic_conventions/README.md
@@ -24,6 +24,5 @@ You can then use the macros for the attribute keys:
 You could also use `OpenTelemetry.SemanticConventions.Logs`, `OpenTelemetry.SemanticConventions.Resource` or
 `OpenTelemetry.SemanticConventions.Trace` modules.
 
-    iex> require OpenTelemetry.SemanticConventions.Logs
-    ...> OpenTelemetry.SemanticConventions.Logs.event_name()
+    iex> OpenTelemetry.SemanticConventions.Logs.event_name()
     :"event.name"
--- a/apps/opentelemetry_semantic_conventions/lib/open_telemetry/semantic_conventions/logs.ex
+++ b/apps/opentelemetry_semantic_conventions/lib/open_telemetry/semantic_conventions/logs.ex
@@ -2,27 +2,23 @@ defmodule OpenTelemetry.SemanticConventions.Logs do
   @doc """
   The schema url for telemetry resources.
 
-      iex> require OpenTelemetry.SemanticConventions.Logs
-      ...> OpenTelemetry.SemanticConventions.Logs.logs_schema_url()
+      iex> OpenTelemetry.SemanticConventions.Logs.logs_schema_url()
       "https://opentelemetry.io/schemas/1.13.0"
   """
   @spec logs_schema_url :: String.t()
-  defmacro logs_schema_url do
+  def logs_schema_url do
     "https://opentelemetry.io/schemas/1.13.0"
   end
-
   @doc """
   The name identifies the event
 
-      iex> require OpenTelemetry.SemanticConventions.Logs
-      ...> OpenTelemetry.SemanticConventions.Logs.event_name()
+      iex> OpenTelemetry.SemanticConventions.Logs.event_name()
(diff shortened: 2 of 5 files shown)


bryannaegele:
A function returning a constant value is effectively a constant and the value will just be inlined by the compiler, so this seems fine.

tsloughter:
So the Elixir compiler will do inlining?

tsloughter:
As in, pretty sure the Erlang compiler will never inline, even when the function body is a literal, since the code could change at runtime.

bryannaegele:
Yeah. It's like with unicode. The values are created at compile time with a
for loop creating a function for each value. And I think a memoization of
sorts happens in the BEAM itself. Like I know that if you have a function
which returns a map, and then you modify it, the map is only allocated once
and then referenced.

bryannaegele:
This is what I'm referring to where the keys are globally allocated and just referenced.

```
When the value for an existing key is updated, the key tuple is not updated, allowing the key tuple to be shared with other instances of the map that have the same keys. In fact, the key tuple can be shared between all maps with the same keys with some care. To arrange that, define a function that returns a map. For example:

new() ->
    #{a => default, b => default, c => default}.
Defined like this, the key tuple {a,b,c} will be a global literal. To ensure that the key tuple is shared when creating an instance of the map, always call new() and modify the returned map:

    (SOME_MODULE:new())#{a := 42}.
Using the map syntax with small maps is particularly efficient. As long as the keys are known at compile-time, the map is updated in one go, making the time to update a map essentially constant regardless of the number of keys updated. The same goes for matching. (When the keys are variables, one or more of the keys could be identical, so the operations need to be performed sequentially from left to right.)
```

[Unicode fun](https://github.com/elixir-lang/elixir/blob/main/lib/elixir/unicode/unicode.ex)

I have no idea if just doing them via attributes might be easier. I feel like this whole thing could be scripted from the spec to something like this.

https://elixir-lang.org/getting-started/module-attributes.html#as-constants

yordis:
It should not break anything, but it will require some refactoring (removing the `require` statement), and I am not sure how slower it will be since it would become a function call at that point instead of a compile-time replacement for the actual value 🤷🏻 no clue

Cantido:
We can inline a function with a module attribute: ` @compile {:inline, myfun: 1}`.

That said, at pre-1.0 and without benchmarks, I think the optimization discussion is a little premature.

tsloughter:
@Cantido true, but its a little different since its potentially adding overhead to code that already works.

But I'm coming at this from Erlang, if removing the require makes for cleaner and easier to use Elixir than I'm ok with the change.

tsloughter (approved):

yordis (approved):
We need to follow up in the `contrib` repo ... (which reminds me I dropped the ball ...)

albertored:
Sorry, joined lately to the discussion just to add that now being functions it is no longer possible to pattern match on them.

Before one can do
```elixir
defmodule Mod do
    require OpenTelemetry.SemanticConventions.Trace

    def arn(%{OpenTelemetry.SemanticConventions.Trace.aws_lambda_invoked_arn() => arn}), do: arn                                                            
    def arn(_), do: nil                                                  
end
```
now this is no longer possible and a module attribute should be created for each function
```elixir
defmodule Mod do
    @arn OpenTelemetry.SemanticConventions.Trace.aws_lambda_invoked_arn()

    def arn(%{@arn => arn}), do: arn                                                            
    def arn(_), do: nil                                                  
end
```

tsloughter:
Ah darn, that'd be nice to have kept.

albertored:
Yes, adding a module or adding a macro-copy of each function with a `_` prefix/suffix

PR #555, opened 2023-03-10
documents valid span statuses

marcdel:
In hindsight the `:unset` was a clue, but it wasn't clear to me what the valid options are for status -- might be nice to spell it out explicitly. I could also see it [here](https://github.com/open-telemetry/opentelemetry-erlang/blob/v1.2.1/apps/opentelemetry_api/lib/open_telemetry.ex#LL106-L106) instead 🤷🏻‍♂️

### DIFF
--- a/apps/opentelemetry_api/lib/open_telemetry.ex
+++ b/apps/opentelemetry_api/lib/open_telemetry.ex
@@ -98,10 +98,11 @@ defmodule OpenTelemetry do
 
   @typedoc """
   An optional final status for this span. Semantically when Status
-  wasn't set it means span ended without errors and assume `unset`.
+  wasn't set it means span ended without errors and assume `:unset`.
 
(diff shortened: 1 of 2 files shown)


bryannaegele:
There is some nuance here that I think needs to be captured if adding more info here. `:unset` is not a status that users can set, so that isn't something we want to say is a valid argument to `set_status` as it will be ignored.

marcdel:
Ahh, okay I see [this](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md?plain=1#L591) now:
```
An attempt to set value `Unset` SHOULD be ignored.
```

marcdel:
I updated the doc to remove `:unset` as a valid parameter value to `set_status`. I added a couple other things, but let me know if it's too much!

tsloughter (approved):

PR #560, opened 2023-03-18
metrics: support delta/cumulative conversions in sum aggregator

tsloughter:
Cumulative aggregate is now also reset on each collection. A previous checkpoint field is tracked in order to add the current value to the previous in the case of needing the cumulative and subtracting in the case you have a cumulative but need a delta (example being an Observable which reports the cumulative value in each callback and needed it reported as a delta).


tsloughter:
Still not a fan of how we have to checkpoint and collect in separate steps but I can't think of another way that doesn't involve new ETS features (specifically a `select_replace` that returns the new, or even the old, value).

ferd (approved):
That strikes me as an entirely safe pattern, classic with ETS operations that don't support atomicity.

PR #561, opened 2023-03-21
metrics: fix observable callbacks to return a list of results

tsloughter:
the instrument could have a result for multiple sets of attributes, so a list of observations must be returned in every case.


tsloughter:
Think I'm happy enough with these that I plan to start documenting stuff next.

ferd (approved):

PR #556, opened 2023-03-10
looks like strict option wants to be on erlef/setup-beam not top level

marcdel:
~🤞🏻~ 🥳

PR #563, opened 2023-03-21
github action: correct usage of version-type strict

tsloughter:


tsloughter:
Closing, #556 covers this

PR #568, opened 2023-03-30
Add tests of an example sampler based on attributes

tsloughter:


tsloughter:
Argh, `maps:intersect` was added in OTP-24. I might just make this test only run on 24+...

tsloughter:
Eh, I made it a more efficient matching instead of changing it to only run on 24+. Just means now I need to update the website docs.

apps/opentelemetry/test/attributes_sampler.erl
```diff
@@ -0,0 +1,47 @@
...
+    case has_match(Attributes, ConfigAttributes) of
+        true ->
+            {?DROP, [], []};
+        _ ->
+            {?RECORD_AND_SAMPLE, [], []}
+    end.
+
+has_match(A, B) ->
+    I = maps:iterator(A),
+    has_match_(maps:next(I), B).
```
ferd, line 40: Would it be faster if you picked the smallest map to iterate from to compare with the other, since you're going for exact matches? Something like:

```
{Min,Max} = if maps:size(A) < maps:size(B) -> {A, B};
               true -> {B, A}
end,
I = maps:iterator(Min),
has_match_(maps:next(I), Max).
```
That'd ensure always having a constantly shorter iteration, though I figured the configured map won't be too big and will often be smallest.

ferd (approved):

apps/opentelemetry/test/attributes_sampler.erl
tsloughter: Yea. I wasn't worried about performance. That's why it was originally using `intersection`.

tsloughter: But may as well since people may actually use this in their own sampler, so I updated it.

PR #569, opened 2023-04-02
Resource as an argument to starting the tracer provider

tsloughter:
Also a commit to remove the setting of the global tracer to `noop` when the SDK stops. I think this makes it safer to not potentially drastically harm performance for users on shutdown. The hope would be it is the last thing done but since this can't be guaranteed and technically could be stopped at anytime, so not having a global GC happen seems best.


tsloughter:
Eh.. this may be wrong, it requires making the resource type available in the API. Rethinking.

ferd (approved):

PR #572, opened 2023-04-10
resolve eqwalizer warnings in experimental applications

tsloughter:
The ets matchspec types are really out of hand...

### DIFF
--- a/apps/opentelemetry_api_experimental/include/otel_metrics.hrl
+++ b/apps/opentelemetry_api_experimental/include/otel_metrics.hrl
@@ -1,7 +1,7 @@
 -record(instrument, {module        :: module(),
                      meter         :: otel_meter:t(),
                      name          :: otel_instrument:name(),
-                     description   :: otel_instrument:description(),
+                     description   :: otel_instrument:description() | undefined,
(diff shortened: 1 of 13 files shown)


ferd:
Would it make sense to make an internal type like:

```
match_var() :: '_' | '$1' | '$2' | '$3' | '$4' | '$5' | '$6' | '$7' | '$8' | '$9'.
match_expr(A) :: undefined | match_var() | {const, A} | {'+', A | match_var(), A | match_var()}.
match_spec() :: match_expr(term()).
match_spec(A) :: match_expr(A).
```
and then just slap that in there? Like it won't tell you if you somehow start using the wrong type of matchspec, but it should make the checker a bit more silent while properly conveying intent? If you support integers in a matchspec you go `match_spec(integer())` and it'll start blowing up on `number()` being used and whatnot, and that should generally help.

tsloughter:
Yea, my initial thought was the warnings about using the wrong number, `$2`, and such would be useful. But I kinda doubt it now, so I'm liking that suggestion.

tsloughter:
Updated per your suggestion @ferd

ferd (approved):

Issue #573, opened 2023-04-17
[Question]: How to disable/enable tracer(s) in runtime


jamars:
Is it possible to disable/enable open-telemetry tracing while the application is running? I see the [spec](https://opentelemetry.io/docs/reference/specification/sdk-environment-variables/#general-sdk-configuration) defines `OTEL_SDK_DISABLED` environment variable as

```Boolean value. If “true”, a no-op SDK implementation will be used for all telemetry signals. Any other value or absence of the variable will have no effect and the SDK will remain enabled. This setting has no effect on propagators configured through the OTEL_PROPAGATORS variable.```

As far as I could see there is no reference to this env var in the code.

tsloughter:
We don't yet have `OTEL_SDK_DISABLED`. Should be simple enough to add, I can look at doing that soon, probably today.

PR #574, opened 2023-04-18
add support for OTEL_SDK_DISABLED environment variable

tsloughter:
Resolves https://github.com/open-telemetry/opentelemetry-erlang/issues/573

### DIFF
--- a/apps/opentelemetry/src/opentelemetry_app.erl
+++ b/apps/opentelemetry/src/opentelemetry_app.erl
@@ -26,26 +26,32 @@
 
 start(_StartType, _StartArgs) ->
     Config = otel_configuration:merge_with_os(
-             application:get_all_env(opentelemetry)),
+               application:get_all_env(opentelemetry)),
 
     %% set the global propagators for HTTP based on the application env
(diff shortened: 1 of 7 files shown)


apps/opentelemetry/src/opentelemetry_app.erl
```diff
@@ -28,12 +28,13 @@ start(_StartType, _StartArgs) ->
...
 
-    %% set global span limits record based on configuration
-    otel_span_limits:set(Config),
-
     %% set the global propagators for HTTP based on the application env
+    %% these get set even if the SDK is disabled
     setup_text_map_propagators(Config),
 
+    %% set global span limits record based on configuration
+    otel_span_limits:set(Config),
```
tsloughter, line 36: I reordered this because I was going to put a `case Disabled` around the parts that don't need to run if the SDK is disabled but then didn't because its not exactly required since the supervisor simply starts no children and that causes everything to fallback to no-ops.

Undecided.. Maybe I should check if its disabled here too and not do stuff like `get_resource`.

ferd (approved):

PR #579, opened 2023-04-29
update semconv to spec release 1.20.0

tsloughter:


ferd (approved):

ferd:
Woops, changes seemed straightforward (autogenerated code) but tests appear to fail here and there.

tsloughter:
Ah, yup, Elixir module ends up with the wrong name with this change.

ferd (approved):

andrewhr:
I'm not familiar with the codegen tool, but would be great if we could maintain old attributes as deprecated and maybe aliases, example:

```
@deprecated "use new_name/0 instead"
def old_name(), do: new_name()
```

That works even for removed attributes, which deprecated message will say it's removed. 

My concern is to make it backward compatible as possible, otherwise this library will end up locking dozens of other libraries, given our approach for Erlang/Elixir to be micro-libraries for each instrumentation.

apps/opentelemetry_semantic_conventions/generate.sh
```diff
@@ -2,8 +2,8 @@
 
 set -ex
 
-semconv_version='1.13.0'
-docker_img_vsn='0.14.0'
+semconv_version='1.20.0'
```
andrewhr, line 5: In the meantime, we got 1.21.0 released 😬

bryannaegele:
We could probably keep a list of them, mark them there and in the changelog, and then set a policy for removal. This doesn't get updated often so it doesn't seem like a big burden to add/modify.

PR #581, opened 2023-05-03
metrics: get the Resource from the Provider/server when registering a Reader

tsloughter:


apps/opentelemetry_experimental/src/otel_meter_server.erl
```diff
@@ -372,15 +373,15 @@ handle_measurement(#measurement{instrument=#instrument{meter=Meter,
...
     update_aggregations(Value, Attributes, Matches, MetricsTab).
 
 handle_measurement(Meter, Name, Number, Attributes, ViewAggregationsTab, MetricsTab) ->
-    Matches = ets:lookup_element(ViewAggregationsTab, {Meter, Name}, 2),
+    Matches = ets:match(ViewAggregationsTab, {{Meter, Name}, '$1'}),
     update_aggregations(Number, Attributes, Matches, MetricsTab).
 
 update_aggregations(Value, Attributes, ViewAggregations, MetricsTab) ->
-    lists:foreach(fun(ViewAggregation=#view_aggregation{}) ->
+    lists:foreach(fun([ViewAggregation=#view_aggregation{}]) ->
```
ferd, line 384: Is this the change that currently makes 2 tests fail? It looks like the previous test just returned unaggregated metrics (2,5) and the new one sum the values (7). Test needs an update to match this change?

ferd (changes requested):

apps/opentelemetry_experimental/src/otel_meter_server.erl
tsloughter: Hm, don't think so. And only 1 test fails locally (thought I'd run them all and they all had passed..) so weird.

tsloughter: Nm, figured it out. Was just the test using different meters between non-macro calls and macro calls. The instrument names are unique per-Meter.

PR #583, opened 2023-05-05
Changed a couple of include_lib directives to include

srstrong:
We're running a bezel-based build system, and it doesn't like these couple of include_lib directives - I believe they technically should just be includes anyway, so this shouldn't cause issues for other build systems (e.g., rebar3)


srstrong:
CLA Auth now completed...

tsloughter (approved):
Yup, `include` is right here. Thanks!

PR #586, opened 2023-05-10
get meter for application instrmentation scope in Elixir macros

tsloughter:


tsloughter:
This change was made in the Erlang macros but not the Elixir macros.

I'm working on adding tests to the top level `tests/otel_tests.exs` so that metrics are actually tested in Elixir, but didn't think it should hold up this change to experimental.

ferd (approved):

PR #589, opened 2023-05-16
bump rebar3 and OTP versions in CI

tsloughter:
OTP-26 might be too new for the setup to actually work, but opening this for when it does.


tsloughter:
Works now but removed testing Elixir on 26 because 1.13 wasn't able to be installed for it.

ferd (approved):

PR #602, opened 2023-08-08
Another attempt at span monitoring

tsloughter:
Things I don't like:

- The monitor option being part of start span but applying to all spans started in that same process is odd.
  - One solution is to have it only end the 1 span but monitoring all spans in a process makes more sense 
  - Another is to have a `monitor(Pid)` function in the API that calls the SDK. Reason I don't like that is it has to lookup what SDK is being used.
- Because `pid` isn't updated at any point this changes how we will be suggesting child spans are created for remote processes from https://opentelemetry.io/docs/instrumentation/erlang/manual/#spans-in-separate-processes to having to instead use a new API (see the new `start_span` macro added in this PR) to pass the parent context when starting the child in the new process. 

I am still kicking around where I stand on these issues and there are probably more, but opening this for feedback.

### DIFF
--- a/apps/opentelemetry/include/otel_span.hrl
+++ b/apps/opentelemetry/include/otel_span.hrl
@@ -59,7 +59,18 @@
           %% trace flags lowest bit is 1 but simply not propagated.
           is_recording            :: boolean() | undefined | '_',
 
-          instrumentation_scope :: opentelemetry:instrumentation_scope() | undefined | '_'
+          instrumentation_scope   :: opentelemetry:instrumentation_scope() | undefined | '_',
+
+
+          %% this is the Erlang process the span is or has was last active in.
+          %% It is used for the optional process monitoring feature where a process
+          %% can be monitored and have all spans
+          %% active in that process be ended if the process exits for any reason.
+          pid                     :: pid() | undefined,
+
+          %% the span processors to run on start and end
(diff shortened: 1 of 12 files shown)


ferd:
Would using the monitor(Pid) call for a given span allow the process touching the span to define how or when it wants it tracked? I could imagine the monitor option just deferring to an API Call for that, but as you said, the biggest issue is maintaining compatibility by adding the call?

tsloughter:
@ferd not sure what you mean about touching a span.

ferd:
> @ferd not sure what you mean about touching a span.

Mostly creating a span in one process and then handing it over to another for most of its lifecycle

tsloughter:
That would depend on updating the `pid` field of the span data in the ets table when it gets handed over.

albertored:
I haven't yet looked at the implementation details but from an end-user point of view if `monitor` is an option on span creation I would expect only that span is monitored, if instead a `monitor(pid)` API is exposed then I would expect all spans handled by that process to be monitored.

Said so, I agree that monitoring only a subset of the spans handled by a process doesn't make much sense.

> Another is to have a monitor(Pid) function in the API that calls the SDK. Reason I don't like that is it has to lookup what SDK is being used.

This is not so clear to me but I'm not very familiar with the internals

PR #603, opened 2023-08-14
Support otp 22

tsloughter:


tsloughter:
Note that the top level readme already says requirements is 22+.

bryannaegele (approved):

ferd:
It's weird that the build works on the default profile, but fails on the test profile, and only on OTP-22. No compiler options changed in between, I'm curious if there'd be a weird issue with symlinks or something.

tsloughter:
Yea, I don't know whats going on, but also pretty sure I've seen it before.

tsloughter:
I have NO CLUE what is going on. I can't get 22 to work in CI. I tested locally and it works fine. I've been at this so long I want to suggest ditching 22 in CI.

ferd (approved):
I'm good with this. I imagine calling this "best effort support" and sometimes having to go back and maintain it is fine.

andrewhr:
Hey @tsloughter, I'm not sure what happened but API 1.2.2 silently broke our setup. Looking at the PR it should be a noop, [but package diff shows that more code is present in this version than CHANGELOG would tell](https://diff.hex.pm/diff/opentelemetry_api/1.2.1..1.2.2). 

It looks like more things got added by accident during publication. That's correct?

tsloughter:
I don't see what is added but not in the changelog?

Maybe the call to provider sup start? But it's in a deprecated function. Do you use it?

andrewhr:
@tsloughter there are changes like this one on `src/otel_tracer_provider.erl`:

```erlang
  start(Name, Config) ->
-   supervisor:start_child(otel_tracer_provider_sup, [Name, Config]).
+   otel_tracer_provider_sup:start(Name, Config).
```

Which were not part of the diff in this PR. I assume this `supervisor:start_child/2` call came earlier and probably it's part of the cause we need to update both API and SDK?

Anyways, looks like the discussion inside the issue is already covering that! 🙇

PR #604, opened 2023-08-17
Experimental fixes

albertored:
A couple of improvements on the experimental applications.

Different contributions are split in different commits so if you prefer to accept only some of them I can discard the unwanted ones:

- [ ] Improvement on typespecs
- [ ] Easier way for creating observable instrumenters without callback
- [ ] Add instrument_unit to view criteria


apps/opentelemetry_api_experimental/src/otel_meter.erl
```diff
@@ -124,31 +149,40 @@ scope(Meter={Module, _}) ->
...
     Module:register_callback(Meter, Instruments, Callback, CallbackArgs).
 
 record(Meter={Module, _}, Name, Number, Attributes) ->
     Module:record(Meter, Name, Number, Attributes).
+
+validate_name(Name) ->
+    NameString = atom_to_list(Name),
+    case re:run(NameString, ?INSTRUMENT_NAME_REGEX) of
+        {match, _} -> ok;
+        nomatch -> error(invalid_name)
```
tsloughter, line 187: Is this caught anywhere? We can't crash in otel.

apps/opentelemetry_api_experimental/test/otel_metrics_SUITE.erl
```diff
@@ -1,93 +0,0 @@
--module(otel_metrics_SUITE).
```
tsloughter, line 1: Why was this module deleted?

apps/opentelemetry_api_experimental/src/otel_meter.erl
albertored: No, you're right, I planned to ask for this and then forgot. 
What is the best approach here? The spec says:

> If the instrument name does not conform to this syntax, the Meter SHOULD emit an error notifying the user about the invalid name. It is left unspecified if a valid instrument is also returned.

We can log an error but then? Discard the instrument? Somehow fix the name? Return an instrument with an invalid name?

About this, I am also wondering if this is the correct place to perform the validation. At the moment this gets executed also when no SDK is used so it's not really noop, we can move it in `otel_meter` implementations but then every implementation should duplicate this logic

apps/opentelemetry_api_experimental/test/otel_metrics_SUITE.erl
albertored: I removed it because:
-  it's not executed in CI, its name is the same as the one in `opentelemetry_experimental` app and rebar executes that one 2 times (see for instance [here](https://github.com/open-telemetry/opentelemetry-erlang/actions/runs/5924757019/job/16062893521?pr=605#step:8:45))
- the single test it contains is no more valid (still using instrument records wiht `value_type`)
- that single test is covering things already covered by tests of `opentelemetry_experimental` app

apps/opentelemetry_api_experimental/src/otel_meter.erl
tsloughter: Hm, I recall this being a conversation in the metrics spec before.... I don't remember the outcome, hehe, but will try to find out.

albertored, line 187: @tsloughter with regards to https://github.com/open-telemetry/opentelemetry-erlang/pull/604#discussion_r1300342971, what about a solution like the following?

```suggestion
validate_name(otel_meter_noop, _Name) ->
    ok;
validate_name(_Module, Name) ->
    NameString = atom_to_list(Name),
    case re:run(NameString, ?INSTRUMENT_NAME_REGEX) of
        {match, _} -> ok;
        nomatch -> error(invalid_name)
```

This allows to:
* keep the validation in the API but without overhead in case of missing SDK
* ensure all SDK implementations have the name validated without code duplication

tsloughter: The interesting thing Go does is it still returns the instrument with the bad name. This allows for the case that the user happens to be using a metrics system that does accept the name. But they still give an error as well.

For them its just returning `instrument, err` and `err` can be checked. I'd rather not do `{ok, Instrument}` and `{error, Instrument}` so instead would log something and return the Instrument.

Other option is to return a noop instrument (in addition to the log).

I'm open to either, but still think the check should be only on the SDK side, instead of checking if its the noop meter.

I don't think we need to worry about multiple SDKs :), and either way, it is defined in the SDK spec, so it is something they are expected to do.

apps/opentelemetry_api_experimental/test/otel_metrics_SUITE.erl
tsloughter: lol, oh, well then, that is all good :)

apps/opentelemetry_api_experimental/src/otel_meter.erl
albertored: Coming from Elixir where ok/error tuples are everywhere I'm not against that option but since it's not so used in otel erlang libs I agree on not going for it.

An invalid name can only cause problems once the metrics are exported (I think) and this is out of our boundaries so I would go for log and return an actual instrument with an invalid name.

Ok for moving it to the SDK

albertored:
@tsloughter review suggestion applied and rebased on main, tests pass locally

albertored:
I think the failing test is a little bit flaky, it fails (not always) also on different PR (see for instance https://github.com/open-telemetry/opentelemetry-erlang/pull/609)

tsloughter:
Yup, I'm rerunning it.Should be fine.

apps/opentelemetry_experimental/src/otel_meter_default.erl
```diff
@@ -64,6 +67,19 @@ register_callback(_, _, _, _) ->
...
+validate_name(Name) when is_atom(Name) ->
+    Re = "^[A-Za-z]+[A-Za-z0-9_.\-]{0,62}$",
+    NameString = atom_to_list(Name),
+    case re:run(NameString, Re) of
+        {match, _} ->
+            ok;
+        nomatch ->
+            ?LOG_ERROR("Invalid instrument name, should be an atom matching '~s', but got '~s'", [NameString]),
+            ok
+    end;
```
tsloughter, line 79:
```suggestion
    case re:run(NameString, Re, [{capture, none}]) of
        match ->
            ok;
        nomatch ->
            ?LOG_ERROR("Invalid instrument name, should be an atom matching '~s', but got '~s'", [NameString]),
            ok
    end;
```

tsloughter (approved):

tsloughter:
Hey, can you send another PR with an update to the changelog? Sorry, I always forget to check this until I hit the merge button :)

albertored:
Also changes to experimental apps go into the main changelog?

PR #605, opened 2023-08-20
Add User-Agent header to exporter requests

### LINKED ISSUE #473: OTLP exporters should emit standard user agent string
As per the following change in the specification (1.14 releasing Oct 4):

> OpenTelemetry protocol exporters SHOULD emit a User-Agent header to at a minimum identify the exporter, the language of its implementation, and the version of the exporter. For example, the Python OTLP exporter version 1.2.3 would report the following:
> ```OTel OTLP Exporter Python/1.2.3```

See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#user-agent

albertored:
Fixes #473

### DIFF
--- a/apps/opentelemetry_exporter/src/opentelemetry_exporter.erl
+++ b/apps/opentelemetry_exporter/src/opentelemetry_exporter.erl
@@ -371,9 +371,20 @@ headers_to_grpc_metadata(Headers) ->
 
 %% make all headers into list strings
 headers(List) when is_list(List) ->
-    [{unicode:characters_to_list(X), unicode:characters_to_list(Y)} || {X, Y} <- List];
+    Headers =[{unicode:characters_to_list(X), unicode:characters_to_list(Y)} || {X, Y} <- List],
+    add_user_agent(Headers);
 headers(_) ->
-    [].
+    add_user_agent([]).
+
+add_user_agent(Headers) ->
+    case lists:search(fun({Header, _}) -> string:to_lower(Header) == "user-agent" end, Headers) of
+        {value, _} -> Headers;
+        false -> [{"User-Agent", user_agent()} | Headers]
+    end.
+
+user_agent() ->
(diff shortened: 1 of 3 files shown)


apps/opentelemetry_exporter/src/opentelemetry_exporter.erl
```diff
@@ -371,9 +371,21 @@ headers_to_grpc_metadata(Headers) ->
...
 headers(List) when is_list(List) ->
-    [{unicode:characters_to_list(X), unicode:characters_to_list(Y)} || {X, Y} <- List];
+    Headers =[{unicode:characters_to_list(X), unicode:characters_to_list(Y)} || {X, Y} <- List],
+    add_user_agent(Headers);
 headers(_) ->
-    [].
+    add_user_agent([]).
+
+add_user_agent(Headers) ->
+    Normalize = fun(Str) -> string:replace(string:to_lower(Str), "-", "_", all) end,
```
tsloughter, line 380: Why is this replacing all `-` with `_`? It then tests that it is equal to `user-agent` so won't that always fail if the header exists?

Looks like there isn't a test for this case?

albertored: Copy-pasted from [here](https://github.com/open-telemetry/opentelemetry-erlang-contrib/blob/main/utilities/opentelemetry_instrumentation_http/src/opentelemetry_instrumentation_http.erl#L18) (too much blindly). In this case it makes no sense, it should only be case-insensitive

albertored:
@tsloughter moving into draft because I need to verify how to add the user agent also on grpc calls

albertored:
@tsloughter we need to fix `grpcbox` for this to work correctly, see linked PR

albertored:
@tsloughter with the recent bump of grpcbox this should be ready to be merged

tsloughter (approved):

PR #606, opened 2023-08-21
Use min/max in histogram match_spec

### LINKED ISSUE #524: Histogram min/max in matchspec
In OTP-26 we can have a much simpler histogram matchspec because min/max have been added to guards: https://github.com/erlang/otp/pull/6567

albertored:
Fixes #524


tsloughter (approved):

tsloughter:
Thanks!

PR #607, opened 2023-08-21
add tracestate module and record

tsloughter:
tracestate is now a record with members element which are validated when added to the tracestate.


apps/opentelemetry_api/src/otel_tracestate.erl
```diff
@@ -0,0 +1,147 @@
...
+-type t() :: #tracestate{}.
+
+-export_type([t/0,
+              members/0]).
+
+%% See https://www.w3.org/TR/trace-context/#tracestate-header
+%% for the limits and string requirements that make up the regexes
+-define(MAX_MEMBERS, 32).
+-define(KEY_MP, element(2, re:compile("^([a-z][_0-9a-z\-\*\/]{0,255})|([a-z0-9][_0-9a-z-*/]{0,240}@[a-z][_0-9a-z-*/]{0,13})$"))).
+-define(VALUE_MP, element(2, re:compile("^([ -+--<>-~]{0,255}[!-+--<>-~])$"))).
```
ferd, line 49: not a blocker here, but since you use `?KEY_MP` and `?VALUE_MP` directly in the calls, I believe the recompilation will happen every time and provide limited benefits.

I realize this is code moved from another module and not a blocker, but I wanted to point it out.

apps/opentelemetry_api/src/otel_tracestate.erl
```diff
@@ -0,0 +1,147 @@
...
+    parse_pairs(Pairs, []);
+parse_pairs(_) ->
+    [].
+
+parse_pairs([], Acc) ->
+    Acc;
+parse_pairs([Pair | Rest], Acc) ->
+    case split(string:trim(Pair)) of
+        {K, V} ->
+            case re:run(K, ?KEY_MP) =/= nomatch
```
ferd, line 127: given what was pointed out earlier, it may make more sense to pass `?KEY_MP` and `?VALUE_MP` as parameters to this function so they're compiled once per list, particularly if there's no other caching (pdict or persistent terms) to save the easy re-compiling at each iteration.

ferd (approved):

apps/opentelemetry_api/src/otel_tracestate.erl
tsloughter: Ugh. Should I replace it with like:

```
-define(VALUE_MP, {re_pattern,1,0,0,
                <<69,82,67,80,152,0,0,0,16,0,0,0,1,0,0,0,255,255,255,
                  255,255,255,255,255,0,0,0,0,0,0,1,0,0,0,64,0,0,0,0,0,
                  0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,131,0,
                  84,27,133,0,76,0,1,110,0,0,0,0,255,239,255,223,255,
                  255,255,255,255,255,255,127,0,0,0,0,0,0,0,0,0,0,0,0,0,
                  0,0,0,104,0,0,0,255,110,0,0,0,0,254,239,255,223,255,
                  255,255,255,255,255,255,127,0,0,0,0,0,0,0,0,0,0,0,0,0,
                  0,0,0,120,0,76,25,120,0,84,0>>}).
```

No idea how often or what versions that would break on. Maybe rare to never?

ferd: tests should catch it I imagine

tsloughter: true

PR #609, opened 2023-08-22
Fix attributes protobuf encoding

albertored:
Attributes with array values are wrongly encoded by exporter:

```erlang
1> otel_otlp_common:to_attributes(#{
        bin_list => [<<"a">>, <<"b">>],
        int_list => [1, 2],
        float_list => [1.2, 3.4],
        bool_list => [true, false]
    }).

[#{key => <<"int_list">>,
   value => #{value => {string_value,<<1,2>>}}},
 #{key => <<"float_list">>,
   value =>
       #{value =>
             {array_value,#{values =>
                                [#{value => {double_value,1.2}},
                                 #{value => {double_value,3.4}}]}}}},
 #{key => <<"bool_list">>,
   value =>
       #{value =>
             {array_value,#{values =>
                                [#{value => {string_value,<<"true">>}},
                                 #{value => {string_value,<<"false">>}}]}}}},
 #{key => <<"bin_list">>,
   value => #{value => {string_value,<<"ab">>}}}]
```

See [specification about attributes](https://opentelemetry.io/docs/specs/otel/common/#attribute)

At the moment I choose to emit an error for nested arrays and omit them from the export but not sure it is the best option


apps/opentelemetry_exporter/src/otel_otlp_common.erl
```diff
@@ -42,9 +47,14 @@ to_instrumentation_scope_proto(#instrumentation_scope{name=Name,
 -spec to_attributes(opentelemetry:attributes_map() | otel_attributes:t() | undefined) -> [opentelemetry_exporter_trace_service_pb:key_value()].
 to_attributes(Attributes) when is_map(Attributes) ->
     maps:fold(fun(Key, Value, Acc) ->
-                      [#{key => to_binary(Key),
-                         value => to_any_value(Value)} | Acc]
-              end, [], Attributes);
+        case to_any_value(Value) of
+            invalid ->
+                ?LOG_ERROR("OpenTelemetry exporter: discarded invalid attribute, ~p", [{Key, Value}]),
```
tsloughter, line 52:
```suggestion
                ?LOG_DEBUG("OpenTelemetry exporter: discarded invalid attribute, ~p", [{Key, Value}]),
```
This can't be the error level. I know this needs to change in one of your other PRs that was merged and I was planning to send a PR for that have the discussion there but I guess may as well now.

I'd like to use an `otel` `domain` for the logs but then they won't get logged by default... (see https://github.com/erlang/otp/issues/6532) so haven't figured out what to do.

apps/opentelemetry_exporter/src/otel_otlp_common.erl
```diff
@@ -71,28 +81,22 @@ to_any_value(Value) when is_boolean(Value) ->
...
         false ->
-            #{value => {array_value, to_array_value(Value)}}
-    end.
+            case to_array_value(Value) of
+                invalid -> invalid;
+                ArrayValue -> #{value => {array_value, ArrayValue}}
+            end
+    end;
+to_any_value(Value) ->
+    #{value => {string_value, to_binary(io_lib:format("~p", [Value]))}}.
```
tsloughter, line 99: Why is this used now instead of `unicode:characters_to_binary`?

ferd: similarly, if using `~p` I would imagine wanting `~tp` as well at least.

albertored: This is not a new addition but only [old code that changed position](https://github.com/open-telemetry/opentelemetry-erlang/pull/609/files#diff-647e41ac745cfba551574cec0d8011c8fea5c0a9f12bc02e9d423af6f6f105cdL87). It is the catch-all clause for values that do not fall into the attributes allowed types.

Be aware that it's not calling `list_to_binary` but only a private `to_binary` [function](https://github.com/open-telemetry/opentelemetry-erlang/pull/609/files#diff-647e41ac745cfba551574cec0d8011c8fea5c0a9f12bc02e9d423af6f6f105cdL125) (already defined in the old code) that is exactly `unicode:characters_to_binary`

tsloughter: Ah, I see now. Got confused using the diff.

But the use of `unicode:character_to_binary` to tell if a list is a string is still removed, right? That is what I noticed and lead me to misread this code as new.

albertored: Yes, it has been removed, that part was the bug.

Attribute values are correctly typed to not be charlist because in that case they are totally undistinguishable from list of integers (that is an allowed attribute value).

So I removed the conversion of charlist to string_value

albertored: The old code tried to covert every list to a binary string so not only list of integers but also list of binaries were wrongly encoded as string_value

tsloughter: Right. My thinking had been that a tuple had to be used if the user wanted an array of what are printable characters (integer).

tsloughter: In order to support iolists I figured this was the only option, require tuples for arrays.

albertored: Yea, I didn't get it. Then most of this PR can be reverted, I will rework it and see if remains something or if it's better to close it.

Is this documented somewhere? If not I can add it

tsloughter: Probably not documented so that'd be really great.

albertored: The only "problem" I see with this solution is from an Elixir user point of view. In Elixir charlist are almost never used so it feels strange to represent list of integers as tuples. New comers often do not ever know they exist, there are plenty of questions online like "why my list of integers is printed as a string?" 😄

Anyway the only other solutions are:
* do not allow charlists as attributes values as done is this PR
* allow charlists but with some tag, for instance `{string, "abc"}`

I would discard the second because it has the same problem as the list-as-tuple but for the Erlang side of the moon: in Erlang charlists are common so why I need to wrap them?

The first may be the better one but it breaks retrocompatibility.

Just sharing my thoughts

tsloughter: I actually don't care about supporting charlists alone, the point was to support iolists so the user didn't have to convert to a binary in process if it would be costly -- instead its left up to the exporter to deal with.

Its a tough one.

tsloughter: A side note. I'd like to get ppl creating attribute records instead of always passing a raw map or list to functions like `with_span` and `add_attributes`. This is to save the time spent creating that record (the searching for bad attributes, truncating...

And that just made me realize that if we are truncating in `otel_attributes` then we should be requiring a binary for all strings and not iolists, ugh. 

So now my issue is backwards compatibility.

albertored: Also at the moment `with_span` and friends only accept raw maps. I imagine this is because the attributes record is defined in sdk with the module doing the validation and truncation. But at the moment the raw map is the only way of adding attributes.

About backward compatibility: it is easier when working not in open source and when the only consumer of a lib is myself   :smile:

tsloughter: Argh, forgot `otel_attributes` was in the SDK and not the API. I think that may need to change... It can't be right since every other language has to have functions/methods for constructing attribute key/value pairs and sets of attributes in their API.

tsloughter: And yea, don't get me started on what I've come to consider mistakes and would change if it wasn't for backwards compatibility :)

albertored: A possible improvement can be to move the attributes record definition and construction in the API lib (validation and truncation included) and allow to pass the record as argument to `with_span` and similar functions.
If a raw map is instead used it is converted to the attributes record before being saved (into trace and metric ets). In this way we always work with  validated attributes and we don't need to duplicate the logic to fix them every time they need to be exported somehow.

The only point that is not so good is that attributes record creation is not a noop also when the SDK is missing. I also checked other languages (only Python in reality) and it seems to do the same.

tsloughter: One reason to construct an attributes record manually is to use in multiple signals, so then it'd somehow have to know if you aren't using any signal (traces, metrics, logs) to know if it should no-op.

Having the option to pass just a map/list instead of the record does at least give the user the ability to pass attributes only to a single signal in a call like `with_span` and have them ignored.

albertored: We can accept both raw maps/lists or attributes record and document that if a record is used the validation and truncation overhead is executed also in a noop case. If a user want to be sure to not add overhead when not needed then he should use maps/lists.

But he main point is that I would like to only save attributes in the validated record form so that exporters downstream do not need to re-validate attributes.

tsloughter: Did we have a path forward on this? hehe

albertored: No, sorry, too many open branches and I forgot about this one, I'll take a look in the next few days

albertored: Closing in favor of #633

PR #613, opened 2023-08-25
Fix temporality handling

albertored:
This PR contains some fixes on how temporality is handled, in particular:

* Readers should use a default cumulative temporality if not specified (see [spec](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#metricreader))
* Remove `TEMPORALITY_UNSPECIFIED` value since [it MUST not be used](https://github.com/open-telemetry/opentelemetry-proto/blob/d1468b7700309cec0a3fdfffbfba4e84acf94072/opentelemetry/proto/metrics/v1/metrics.proto#L253)
* Move the default instrument temporality definition into `otel_instrument` module
* Configure the previous default temporality mapping in test readers to avoid breaking tests


tsloughter:
Gauge and last value aggregation aren't "cumulative" so I think "unspecified" macro is still needed. It just isn't reported in the proto so it isn't used there, but makes things clearer in our code not to claim it is cumulative when it isn't.

apps/opentelemetry_exporter/test/opentelemetry_exporter_SUITE.erl
```diff
@@ -76,7 +76,7 @@ verify_metrics_export(Config) ->
                        name = <<"sum name">>,
                        description = <<"some sum description">>,
                        unit = kb,
-                       data = #sum{aggregation_temporality = 'AGGREGATION_TEMPORALITY_CUMULATIVE',
+                       data = #sum{aggregation_temporality = temporality_cumulative,
```
tsloughter, line 79: Is this right? Isn't `'AGGREGATION_TEMPORALITY_CUMULATIVE'` what the proto encoder expects?

albertored: yes, the value `temporality_cumulative` is converted to `'AGGREGATION_TEMPORALITY_CUMULATIVE'` in the map generated by the `otel_otlp_metrics:to_proto/2` function that is called by `opentelemetry_exporter:export/4`

albertored:
> Gauge and last value aggregation aren't "cumulative" so I think "unspecified" macro is still needed. It just isn't reported in the proto so it isn't used there, but makes things clearer in our code not to claim it is cumulative when it isn't.

I'm not able to find the piece of spec where this is explained but, if I remember correctly, the default temporality for sync instruments is delta while for async ones is cumulative.

Apart from that the instrument default temporality is only used for building the view aggregation record that then it is used by the reader to determine how to handle measurement based on his config (that is if a temporality transformation is needed or not). If this value is unspecified how can the reader handle the measurements?

tsloughter:
Ah, looking at the code now I see that the difference in delta and cumulative for last value is whether the `start_time_unix_nano` gets reset on collection.

Clearly don't have tests that verify the default for last value works since my guess is that it is always using the cumulative branch of `checkpoint` :)

albertored:
@tsloughter is this ok or are there things to fix?

tsloughter:
Oh, I almost forgot again, the CHANGELOG needs updating :)

albertored:
@tsloughter ready, I always forget too...

tsloughter (approved):

albertored:
changelog conflict fixed

PR #617, opened 2023-08-28
Allow nested attribute maps in trace attributes

ejscunha:


bryannaegele:
Hey @ejscunha. Maps are not an allowed value type for attributes in opentelemetry. The list of allowed types is what is allowed per the spec.

PR #618, opened 2023-08-29
move attribute smodule to api

tsloughter:


ferd (approved):

bryannaegele:
What's prompting the move?

tsloughter:
@bryannaegele it was in the wrong place. Not sure why it was ever in the SDK. But motivating it to happen now is the desire to support `#attributes` as the argument to stuff like `start_span` so the user can process the Attributes only once -- whether that is to use the same attribute set in spans or in both a span, metric and log.

bryannaegele (approved):

albertored:
At the moment attributes are stored (in spans and metrics ets tables) as provided by the user, should we cast them to attributes record before saving them?

tsloughter:
@albertored not sure what you mean. In the ets table the attributes are the record.

albertored:
@tsloughter my bad, span ETS table contains indeed attributes as records but the same is not true for metrics, there attributes are stored as raw maps. I think we should do the same for metrics (storing the record), I can open a PR for that if you think is right

PR #619, opened 2023-08-29
Add prometheus exporter

albertored:
First draft of the Prometheus exporter to collect some feedback.

Open points:

1. At the moment it is embedded in the `opentelemetry_experimental` app to speed up development, should it be a separate app?

2. As suggested in the spec I used a reader and an exporter but from the user point of view the exporter is masked and its config parameters are on the reader, for example:
   ```elixir
   config :opentelemetry_experimental,
     readers: [
       %{
         module: :otel_metric_reader_prometheus,
         config: %{
           add_scope_info: true,
           add_total_suffix: true,
           endpoint_port: 3000
         }
       }
     ]
   ```
   `default_temporality_mapping`, `export_interval_ms` and `exporter` options on the reader are overridden, should we emit a warning if the user configures them?

3. The encoding part works but it can be improved for sure, I still need to go through that module again and refactor it. Anyway, any suggestion is welcomed.

4. The `otel_metric_reader_prometheus` is in reality a supervisor that starts the reader itself and the httpd server if needed. The httpd server is started only if an `endpoint_port` is given to the configuration. We can maybe have a more explicit configuration field for this?

5. If the httpd server is not used it should be possible for the user to get the metrics in prometheus format so that he can expose them. This can be done by calling the `otel_metric_reader_prometheus:collect/1` function but it needs the reader pid as an argument, not so user-friendly.

6. How to implement the same behavior of `erlang:float_to_binary(2.0, [short])` that was added in OTP 25? (tests failing for this reason)

### DIFF
--- a/apps/opentelemetry_experimental/include/otel_metrics.hrl
+++ b/apps/opentelemetry_experimental/include/otel_metrics.hrl
@@ -115,7 +115,7 @@
         {
          name :: otel_view:name(),
          scope :: opentelemetry:instrumentation_scope() | undefined,
-         description :: otel_instrument:description(),
-         unit :: otel_instrument:unit(),
+         description :: otel_instrument:description() | undefined,
+         unit :: otel_instrument:unit() | undefined,
          data :: #sum{} | #gauge{} | #histogram{}
         }).
--- a/apps/opentelemetry_experimental/src/otel_metric_exporter_prometheus.erl
+++ b/apps/opentelemetry_experimental/src/otel_metric_exporter_prometheus.erl
@@ -1,0 +1,633 @@
+%%%------------------------------------------------------------------------
+%% Copyright 2022, OpenTelemetry Authors
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
+%%
+%% http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing, software
+%% distributed under the License is distributed on an "AS IS" BASIS,
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+%%
+%% @doc
+%% @end
+%%%-------------------------------------------------------------------------
+
+-module(otel_metric_exporter_prometheus).
+
+-record(opts,{
+    add_scope_info :: boolean(),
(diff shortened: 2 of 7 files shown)


eproxus:
I'm very interested in use case 5, since we already have our own web server and do not want to export an additional port.

albertored:
> I'm very interested in use case 5, since we already have our own web server and do not want to export an additional port.

Yea, I think this will be the most common use case, we need to find a way to make this easy for users

starbelly:
I am also interested in use case 5. Will evaluate switching to otel collector but for now this would go in nicely on my current project 😄

albertored:
@tsloughter I'll try this weekend hoping things have not changed too much. Unfortunately in the last months I was not able to follow the project with constancy

RoadRunnr:
@albertored I've rebased and modified your version a bit https://github.com/open-telemetry/opentelemetry-erlang/compare/main...travelping:opentelemetry-erlang:feature/prometheus-exporter-for-upstream

The reader logic can now be used from a cowboy handler as well (sample included).

albertored:
@tsloughter rebased and pushed with additions from @RoadRunnr, we should still resolve the points in my first comment before merging

albertored:
PS: [this commit](https://github.com/open-telemetry/opentelemetry-erlang/pull/619/commits/366871e3598079cad7016fa0c59ff73c1d64f8f4) is causing failures in the pipeline since `maps:iterator/2` was introduced in erlang 26. I don't know if there is an efficient way of doing it without that function (I can only think about converting map to list and order it before folding). If not we should revert that commit

RoadRunnr:
> PS: [this commit](https://github.com/open-telemetry/opentelemetry-erlang/pull/619/commits/366871e3598079cad7016fa0c59ff73c1d64f8f4) is causing failures in the pipeline since `maps:iterator/2` was introduced in erlang 26. I don't know if there is an efficient way of doing it without that function (I can only think about converting map to list and order it before folding). If not we should revert that commit

The ordering is only used for the tests to get defined ordering for comparing the results. It is therefore not to critical that they are super efficient. The simple sort of the map keys would be enough. I can supply a patch if you want?

> we should still resolve the points in my first comment before merging

point 5 (or use case 5) should be solved. A sample cowboy handler is included.

> The encoding part works but it can be improved for sure.

I do have additional changes pending in https://github.com/travelping/opentelemetry-erlang/compare/bbcfe2f6370c89fcd8c7ed5981fd8c171aeff4d0...0d5f677ccc64a069274689aeeee70663acb41eed

PR #620, opened 2023-08-30
move attribute processing to otel-attribute module

tsloughter:
This just cleans up some stuff that got put in otel_span because otel_attributes was in the SDK.

I checked that no lib in -contrib is calling the otel_span functions

### DIFF
--- a/apps/opentelemetry_api/src/otel_attributes.erl
+++ b/apps/opentelemetry_api/src/otel_attributes.erl
@@ -21,8 +21,16 @@
          set/2,
          set/3,
          dropped/1,
-         map/1]).
+         map/1,
+         is_valid_attribute/2,
+         process_attributes/1]).
 
+-define(is_allowed_key(Key), (is_atom(Key) orelse (is_binary(Key) andalso Key =/= <<"">>))).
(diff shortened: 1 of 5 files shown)


apps/opentelemetry_api/src/otel_attributes.erl
```diff
@@ -21,8 +21,16 @@
...
+         map/1,
+         is_valid_attribute/2,
+         process_attributes/1]).
 
+-define(is_allowed_key(Key), is_atom(Key) orelse (is_binary(Key) andalso Key =/= <<"">>)).
+-define(is_allowed_value(Value), is_atom(Value) orelse
+                                 is_boolean(Value) orelse
+                                 is_number(Value) orelse
+                                 is_binary(Value) orelse
+                                 is_list(Value)).
```
ferd, line 33: you might want to put both of these guard clauses in a wrapping `(...)` so they can be composed with others without messing with evaluation rules when `andalso` and `orelse` are called with them.

tsloughter: Good point.

PR #623, opened 2023-09-04
Improve ets queries performance

albertored:
Passing the key pattern to `ets:select` and `ets:select_replace` instead of using match conditions to select the correct elements improves performances allowing to use ETS indexing.

These are the results of a quick benchee on my laptop with some fake but similar data

```
Comparison: 
select_replace match         59.25
select_replace cond          52.14 - 1.14x slower +2.30 ms
select match                 45.49 - 1.30x slower +5.10 ms
select cond                  36.52 - 1.62x slower +10.50 ms
```

Here the full script for reference:

```elixir
:ets.new(:table, [:set, :named_table, {:keypos, 2}, :public])

instr_count = 100
reader_count = 10
attrs_count = 100

data =
  for instr <- 1..instr_count, reader <- 1..reader_count, attrs <- 1..attrs_count do
    {:metric, {"i#{instr}", "a#{attrs}", "r#{reader}"}, instr + reader + attrs}
  end

Benchee.run(
  %{
    "select match" => fn _ ->
      select = [{{:_, {"i33", :_, "r3"}, :_}, [], [:"$_"]}]
      [{:metric, {"i33", _, _}, _} | _] = :ets.select(:table, select)
    end,
    "select cond" => fn _ ->
      select = [
        {:"$1",
         [
           {:"=:=", "i33", {:element, 1, {:element, 2, :"$1"}}},
           {:"=:=", "r3", {:element, 3, {:element, 2, :"$1"}}}
         ], [:"$1"]}
      ]

      [{:metric, {"i33", _, _}, _} | _] = :ets.select(:table, select)
    end,
    "select_replace match" => fn _ ->
      select = [
        {{:"$1", {"i33", :"$2", "r3"}, :"$3"}, [], [{{:"$1", {{"i33", :"$2", "r3"}}, 0}}]}
      ]

      ^attrs_count = :ets.select_replace(:table, select)
    end,
    "select_replace cond" => fn _ ->
      select = [
        {{:"$1", :"$2", :"$3"},
         [
           {:"=:=", "i33", {:element, 1, :"$2"}},
           {:"=:=", "r3", {:element, 3, :"$2"}}
         ], [{{:"$1", :"$2", 0}}]}
      ]

      ^attrs_count = :ets.select_replace(:table, select)
    end
  },
  before_scenario: fn _ ->
    :ets.delete_all_objects(:table)
    :ets.insert(:table, data)
  end,
  time: 10
)

```


tsloughter:
Interesting. I could have swore I did this for a reason and wasn't worried about the perf since it wasn't in a hot path. But if it works the improved performance and readability is welcomed :)

albertored:
Yes, the gain is not so much but that plus the readability change made me open the PR. I think tests are failing for the usual flaky test

apps/opentelemetry_experimental/src/otel_aggregation_histogram_explicit.erl
```diff
@@ -189,9 +189,8 @@ checkpoint(Tab, #view_aggregation{name=Name,
                                            max='$7',
                                            sum='$8'
                                           },
-           [{'=:=', {element, 1, '$1'}, {const, Name}},
-            {'=:=', {element, 3, '$1'}, {const, ReaderId}}],
-           [{#explicit_histogram_aggregation{key='$1',
+           [],
+           [{#explicit_histogram_aggregation{key={{Name, '$1', {const, ReaderId}}},
```
tsloughter, line 193: Why is `Name` not `{const, Name}`?

tsloughter:
Sorry for the delay. Looks good except I had one question.

apps/opentelemetry_experimental/src/otel_aggregation_histogram_explicit.erl
albertored: Probably a typo in making the change. `const` in only needed for avoiding wrapping tuples into tuples or is there any other use?

apps/opentelemetry_experimental/src/otel_aggregation_last_value.erl
```diff
@@ -74,14 +73,13 @@ checkpoint(Tab, #view_aggregation{name=Name,
...
+    MS = [{#last_value_aggregation{key={Name, '$1', ReaderId},
                                    start_time_unix_nano='$3',
                                    last_start_time_unix_nano='_',
                                    checkpoint='_',
                                    value='$2'},
-           [{'=:=', {element, 1, '$1'}, {const, Name}},
-            {'=:=', {element, 3, '$1'}, {const, ReaderId}}],
-           [{#last_value_aggregation{key='$1',
+           [],
+           [{#last_value_aggregation{key={{Name, '$1', {const, ReaderId}}},
```
tsloughter, line 82:
```suggestion
           [{#last_value_aggregation{key={{{const, Name}, '$1', {const, ReaderId}}},
```

tsloughter:
I updated the branch to the latest main and it broke the Erlang build.

apps/opentelemetry_experimental/src/otel_aggregation_histogram_explicit.erl
ferd: It also likely avoids accidentally sneaking in magic atom names like `$_` or `$1` in the field you're looking for, which would suddenly change the semantics of the literal value into a matchspec special value.

PR #628, opened 2023-09-19
Support explicit_bucket_boundaries advisory parameters

albertored:
`ExplicitBucketBoundaries` is almost stable, already approved, only needs to be merged

See https://github.com/open-telemetry/opentelemetry-specification/pull/3694

The [added test](https://github.com/open-telemetry/opentelemetry-erlang/pull/628/files#diff-72303ef69e83043fb8789bf8144db3791f7de0dd28f43f8ce8242710cc8c6ab8R1134) should be explicative about the behaviour

### DIFF
--- a/apps/opentelemetry_experimental/src/otel_view.erl
+++ b/apps/opentelemetry_experimental/src/otel_view.erl
@@ -125,10 +125,10 @@ match_instrument_to_views(Instrument=#instrument{name=InstrumentName,
 
 %%
 
-aggragation_options(#{boundaries := _} = AggregationOptions, _AdvisoryParams) ->
+aggragation_options(#{explicit_bucket_boundaries := _} = AggregationOptions, _AdvisoryParams) ->
     AggregationOptions;
 aggragation_options(AggregationOptions, #{explicit_bucket_boundaries := Boundaries}) ->
-    maps:put(boundaries, Boundaries, AggregationOptions);
+    maps:put(explicit_bucket_boundaries, Boundaries, AggregationOptions);
 aggragation_options(AggregationOptions, _AdvisoryParams) ->
     AggregationOptions.
 
--- a/apps/opentelemetry_experimental/src/otel_view.erl
+++ b/apps/opentelemetry_experimental/src/otel_view.erl
@@ -78,7 +78,8 @@ new(Name, Criteria, Config) ->
 -spec match_instrument_to_views(otel_instrument:t(), [t()]) -> [{t() | undefined, #view_aggregation{}}].
 match_instrument_to_views(Instrument=#instrument{name=InstrumentName,
                                                  meter=Meter,
-                                                 description=Description}, Views) ->
+                                                 description=Description,
+                                                 advisory_params=AdvisoryParams}, Views) ->
     IsMonotonic = otel_instrument:is_monotonic(Instrument),
     Temporality = otel_instrument:temporality(Instrument),
     Scope = otel_meter:scope(Meter),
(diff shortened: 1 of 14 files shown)


albertored:
Fixes #629

RoadRunnr:
managed to crash it.

Give this sequence:

```erlang
    Meter = opentelemetry_experimental:get_meter(),
    true = otel_meter_server:add_view(
             #{instrument_kind => ?KIND_HISTOGRAM},
             #{aggregation_module => otel_aggregation_histogram_explicit,
               aggregation_options => #{boundaries => [10, 100]}}),
    Histogram = otel_histogram:create(Meter, histogram, #{}),
    true = otel_meter_server:add_view(
             #{instrument_name => histogram},
             #{
               aggregation_module => otel_aggregation_histogram_explicit,
               aggregation_options => #{boundaries => [10, 20, 30, 40, 50, 60, 70, 80, 100]}}),
    ok = otel_histogram:record(Histogram, 2000, #{<<"a">> => <<"1">>})
```

It will crash with:
```
{badarg,
    [{counters,add,
         [{write_concurrency,#Ref<0.388515526.2174353409.18119>},10,1],
         [{error_info,#{module => erl_erts_errors}}]},
     {otel_aggregation_histogram_explicit,aggregate,4,
         [{file,
              "/usr/src/erlang/opentelemetry-erlang/apps/opentelemetry_experimental/src/otel_aggregation_histogram_explicit.erl"},
          {line,175}]},
     {otel_aggregation,maybe_init_aggregate,4,
         [{file,
              "/usr/src/erlang/opentelemetry-erlang/apps/opentelemetry_experimental/src/otel_aggregation.erl"},
          {line,49}]},
```

Root problem is that the the second view does not update the counter bucket in the histogram view.

There is nothing in the OpenTelemetry specification that indicates that doing something like that is not permitted.
Even the code suggest that it should be fine to update views like that. Of course, I would expect that such an updated flushes any already collected data points.

albertored:
Thanks, I will look at it

tsloughter:
We likely don't want to support updating views -- I don't think any implementation does. But also don't want to crash :)

There has been talk about this in the metrics SIG, about whether you can even add views after the meter provider has started or not.

albertored:
There are a couple of problems revealed from this:

1. we are missing the name parameter in view stream configuration and wrongly using the instrument_name selector for that. Because of this your example is producing a duplicated view aggregation with different boundaries thus causing the crash. (I will open a PR to fix this)
2. We do not handle very well duplicated metrics, see https://opentelemetry.io/docs/specs/otel/metrics/sdk/#measurement-processing. In the case of the example you provided we are defining a duplicated metric with different boundaries, it is not exactly what is written in the spec but I think we should drop the second view

apps/opentelemetry_experimental/src/otel_view.erl
```diff
@@ -101,27 +103,35 @@ match_instrument_to_views(Instrument=#instrument{name=InstrumentName,
...
         Aggs ->
             Aggs
     end.
 
 %%
 
+aggragation_options(#{boundaries := _} = AggregationOptions, _AdvisoryParams) ->
+    AggregationOptions;
+aggragation_options(AggregationOptions, #{explicit_bucket_boundaries := Boundaries}) ->
+    maps:put(boundaries, Boundaries, AggregationOptions);
```
tsloughter, line 131: Can just rename `boundaries` to `explicit_bucket_boundaries` in `otel_aggregation_histogram_explicit`.

tsloughter (approved):
Had one suggestion.

albertored:
@tsloughter ready

PR #630, opened 2023-10-02
Fix flaky test

### LINKED ISSUE #616: Flappy Elixir test

albertored:
Fix #616 

The problem is that the 2 test modules were running concurrently and they are both modifying the `:opentelemetry` application environment. Sometimes the metric test sets the exporter to none and next span test fails.

### DIFF
--- a/test/otel_metric_tests.exs
+++ b/test/otel_metric_tests.exs
@@ -1,5 +1,5 @@
 defmodule OtelMetricTests do
-  use ExUnit.Case, async: true
+  use ExUnit.Case, async: false
 
   require OpenTelemetryAPIExperimental.Counter, as: Counter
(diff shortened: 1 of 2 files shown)


tsloughter (approved):
Ah! Thanks!

PR #631, opened 2023-10-02
Return ctx on attach

bryannaegele:
Given `attach/1` is doing a pdict put which returns `undefined` on success for the first call and the value on subsequent calls, always returning the ctx value which was passed rather than undefined is consistent. Selfishly, it will clean up a couple things I am doing where pipelining into attach from an `extract_to/2` call.

### DIFF
--- a/apps/opentelemetry/test/opentelemetry_SUITE.erl
+++ b/apps/opentelemetry/test/opentelemetry_SUITE.erl
@@ -745,7 +745,8 @@ attach_ctx(Config) ->
     Ctx = otel_ctx:get_current(),
 
     erlang:spawn(fun() ->
-                         otel_ctx:attach(Ctx),
+                         ReturnVal = otel_ctx:attach(Ctx),
+                         ?assertEqual(Ctx, ReturnVal),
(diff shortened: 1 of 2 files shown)


bryannaegele:
On the failing tests, I'm not sure what behavior we're looking for here when `detach` is called. 

```
-spec detach(token()) -> ok.
detach(Token) ->
    %% at this time `Token' is a context
    update_logger_process_metadata(Token),
    erlang:put(?CURRENT_CTX, Token).
```

I don't know why this would result in `assert :undefined = Tracer.current_span_ctx()` would be true. The current implementation seems like it's just a copy of attach.

tsloughter:
It is meant to return the previous context.

tsloughter:
It is in the spec: https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/context#attach-context

> The API MUST return a value that can be used as a Token to restore the previous Context.

bryannaegele:
Crap. Forgot about that and couldn't find it in the spec after it moved. :/

PR #632, opened 2023-10-03
Improve spec compliance

albertored:
1. Attributes are optional in Counter.add(), UpDownCounter.add() and Histo.record()
2. Counter.add() and Histogram.record() accept only positive numbers

For the second point I only log and discard the value, should we also return an error? If so it will happen only when the SDK is present so I think we are spec compliant. The following is the only part of spec I found mentioning this

> The increment value is expected to be non-negative. This API SHOULD be documented in a way to communicate to users that this value is expected to be non-negative. This API SHOULD NOT validate this value, that is left to implementations of the API.


tsloughter:
The checking of positive numbers is actually a mistake in the matrix. 

I see you put it in the SDK at least but the matrix I think is referring to the API.

We already drop negative values in aggregators if the aggregation is monotonic.

tsloughter:
Right, that part of the spec says:

> This API SHOULD NOT validate this value

albertored:
Yes, my interpretation of 
> this is left to implementation of the API
was that the validation should be done on SDK.

Not clear to me why this validation is done on aggregations. Monotonic is a property of the instrument not of the aggregation, in addition at the moment the check for positive values is done only on sum aggregation (checked from smartphone so I may have missed something)

tsloughter:
Ah yea, its only done on the sum aggregation right now.

CHANGELOG.md
```diff
@@ -49,6 +50,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
  - [Correctly record histogram values greater than last boundary](https://github.com/open-telemetry/opentelemetry-erlang/pull/614)
  - [Readers should use a default cumulative temporality if not specified](https://github.com/open-telemetry/opentelemetry-erlang/pull/613)
+ - [Counter.add() and Histogram.record() accept only positive numbers](https://github.com/open-telemetry/opentelemetry-erlang/pull/632)
```
tsloughter, line 53: You removed this right?

albertored: Not yet, I'm still not convinced about having it on aggregations instead of instruments, I checked a couple of implementations on other languages and all of them have this check on the instrument side:

- [python](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-sdk/src/opentelemetry/sdk/metrics/_internal/instrument.py#L150)
- [java](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkDoubleCounter.java#L40)
- [javascript](https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/sdk-metrics/src/Instruments.ts#L102)

tsloughter: Right, its just a matter of wording, not implementation. You check it in the SDK https://github.com/open-telemetry/opentelemetry-erlang/pull/632/files#diff-9fcf9e6f6f81d312bdec357c254b29a41e56886b859fae4250847e293f18128eR378 not in `Counter.add`. So `Counter.add` accepts a negative number but then it gets rejected in the SDK by `record`.

albertored: Right, I posted the comment without looking at my code and I forgot where the check is placed :-D
I still think that having this check here is clearer than implementing it in specific aggregators but I can move it if you prefer. 

But I may have misunderstood your comment, were you only suggesting about rewording the changelog entry?

tsloughter: Yea, just rewording.

albertored: Should we also remove the check on positive values on the sum aggregation? Now it is already checked elsewhere

albertored:
@tsloughter changelog updated. I also removed the check for positive values on the sum aggregation module since it is now already done elsewhere. If you prefer I can revert that last commit but to me it seems more clear in this way

tsloughter (approved):

tsloughter:
Well that is concerning... tests pass on 24 but not 26 and the failure is unexpected metric results.

tsloughter:
```
%%% otel_metrics_SUITE ==> float_updown_counter: FAILED
%%% otel_metrics_SUITE ==> 
Failure/Error: ?assertMatch([ ], lists : sort ( [ { 3.3 , # { } } , { 10.0 , # { << "c" >> => << "b" >> } } ] ) -- SortedDatapoints)
  expected: = [ ]
       got: [{3.3,#{}},{10.0,#{<<"c">> => <<"b">>}}]
      line: 305
   comment: [{5.4,#{}},{15.4,#{<<"c">> => <<"b">>}}]
```

So the test expects 3.3 and 10.0 but gets 5.4 and 15.4. I don't even see how those numbers are possible from the recordings made.

apps/opentelemetry_experimental/src/otel_meter_server.erl
```diff
@@ -374,15 +374,20 @@ handle_measurement(Meter, Name, Number, Attributes, ViewAggregationsTab, Metrics
...
-                                                                Attributes);
+    lists:foreach(fun([ViewAggregation=#view_aggregation{instrument=Instrument}]) ->
+                        maybe_init_aggregate(Value, Instrument, MetricsTab, ViewAggregation, Attributes);
                      (_) ->
                           ok
                   end, ViewAggregations).
 
+maybe_init_aggregate(Value, #instrument{kind=Kind} = Instrument, _MetricsTab, _ViewAggregation, _Attributes)
+        when Value < 0, Kind == ?KIND_COUNTER orelse Kind == ?KIND_HISTOGRAM ->
+    ?LOG_INFO("Discarding negative value for instrument ~s of type ~s", [Instrument#instrument.name, Kind]),
```
tsloughter, line 385: Wonder if this shouldn't be a debug log instead? To guard against a messed up dependency flooding info logs with logs not about actual functionality *shrug*

albertored:
Really strange indeed, I'll take a look

apps/opentelemetry_experimental/src/otel_meter_server.erl
albertored: Yea I'm always dubious about the level. I got your point but on the other side a debug log is hardly seen by the user. So I'm ok both with debug and info, we should also align this all over the code

albertored:
@tsloughter  tests are running in while loop since a while with the same version of CI (26.1.2) and they are consistently passing.  Can you try to re-trigger the CI and run them locally?

tsloughter:
Yea, they pass, makes me fear a race condition.

tsloughter:
Merged but opened https://github.com/open-telemetry/opentelemetry-erlang/issues/661

PR #635, opened 2023-10-10
fix(otel_batch_processor): don't divide `max_queue_size` by word-size

SergeTupchiy:
There is no need to convert `max_queue_size` to words, as it is compared with the number of objects in ETS table: https://github.com/open-telemetry/opentelemetry-erlang/blob/main/apps/opentelemetry/src/otel_batch_processor.erl#L278.

### DIFF
--- a/apps/opentelemetry/src/otel_batch_processor.erl
+++ b/apps/opentelemetry/src/otel_batch_processor.erl
@@ -190,11 +190,17 @@ callback_mode() ->
 idle(enter, _OldState, Data=#data{exporter=undefined,
                                   exporter_config=ExporterConfig,
                                   scheduled_delay_ms=SendInterval,
+                                  check_table_size_ms=CheckInterval,
                                   reg_name=RegName}) ->
     Exporter = init_exporter(RegName, ExporterConfig),
-    {keep_state, Data#data{exporter=Exporter}, [{{timeout, export_spans}, SendInterval, export_spans}]};
-idle(enter, _OldState, #data{scheduled_delay_ms=SendInterval}) ->
-    {keep_state_and_data, [{{timeout, export_spans}, SendInterval, export_spans}]};
+    {keep_state, Data#data{exporter=Exporter},
+     [{{timeout, export_spans}, SendInterval, export_spans},
+      {{timeout, check_table_size}, CheckInterval, check_table_size}]};
+idle(enter, _OldState, #data{scheduled_delay_ms=SendInterval,
+                             check_table_size_ms=CheckInterval}) ->
+    {keep_state_and_data,
+     [{{timeout, export_spans}, SendInterval, export_spans},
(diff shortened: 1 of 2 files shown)


bryannaegele:
I think that might actually be a bug in the code you linked as I think the intention is to limit memory consumption, not the number of spans. The rules around limits are focused around things that affect memory pressure.

I think the info check should actually be this:

`case ets:info(?CURRENT_TABLE(RegName), memory) of`

SergeTupchiy:
> I think that might actually be a bug in the code you linked as I think the intention is to limit memory consumption, not the number of spans. The rules around limits are focused around things that affect memory pressure.
> 
> I think the info check should actually be this:
> 
> `case ets:info(?CURRENT_TABLE(RegName), memory) of`

Hi @bryannaegele, 
Thanks for your reply. 
It used to be compared against memory size (in words) indeed: 
- https://github.com/open-telemetry/opentelemetry-erlang/commit/571d4af0b3c61ce29843d6edc4f09260e9b097d8#diff-5bd137db87b7d675fde66547dccf5b47aadebcd8f6f1401bc5d5a6e2ff3ebc7bR172
- https://github.com/open-telemetry/opentelemetry-erlang/commit/571d4af0b3c61ce29843d6edc4f09260e9b097d8#diff-f94a0d999fc7707f9c724767c27af9e5009313ecdcb865b0da2b5f364c45cec6L167

I'd be happy to fix it to use memory instead, but I think we also need to account for some pre allocated memory per table:
```
> ets:info('otel_batch_processor_otel_export_table1_<0.4149.0>', memory).
1377
> ets:info('otel_batch_processor_otel_export_table1_<0.4149.0>', size).  
0
```
Basically, an empty table already occupies more words than the default max_queue_size: `2048 div erlang:system_info(wordsize) = 256`

tsloughter:
Wow, yea, this got all messed up...

Hm, because spans have options for limits on attributes maybe we should just use the # of spans instead of size of the table?

bryannaegele:
What if we supported both? `integer` for number of entries, `{memory_unit, integer}` for memory size? I suspect most folks would want to use a memory cap as span sizes and volume can vary so greatly and should be the default method.

SergeTupchiy:
@tsloughter , @bryannaegele,
Found another related issue: check_table_size timeout has not been started at all, attempted to fix it in this commit: https://github.com/open-telemetry/opentelemetry-erlang/pull/635/commits/42b8376c09b31491c42519e1ba8e54b1a827eba2

tsloughter:
I'm thinking we should go with this and can add a `max_queue_size_bytes` if desired in the future if desired.

@bryannaegele does that sound good?

tsloughter (approved):

bryannaegele:
Yep. I'm fine with length

bryannaegele (approved):

PR #637, opened 2023-10-13
Fix log metadata and log exporter

SergeTupchiy:


apps/opentelemetry_api/src/otel_ctx.erl
```diff
@@ -147,9 +147,5 @@ text_map_injector(Key, ToText) ->
 text_map_injector_fun(TextMap, Key, ToText) ->
     TextMap ++ ToText(?MODULE:get_value(Key, undefined)).
 
-%%
-
-update_logger_process_metadata(undefined) ->
-    ok;
```
SergeTupchiy, line 153: `undefined` Ctx will be properly handled by `otel_tracer` which will eventually clear Logger metadata.

apps/opentelemetry_experimental/src/otel_log_handler.erl
```diff
@@ -189,8 +189,8 @@ exporting(enter, _OldState, _Data) ->
 exporting(internal, export, Data=#data{exporter=Exporter,
                                        resource=Resource,
                                        config=Config,
-                                       batch=Batch}) when map_size(Batch) =/= 0 ->
-    _ = export(Exporter, Resource, Batch, Config),
+                                       batch=Batch}) ->
+    map_size(Batch) =:= 0 orelse export(Exporter, Resource, Batch, Config),
```
SergeTupchiy, line 193: I think this quick fix is worthwhile, even though logs implementation is experimental, it make sense to keep it functional. :slightly_smiling_face:

apps/opentelemetry_experimental/src/otel_otlp_logs.erl
```diff
@@ -87,17 +87,15 @@ log_record(#{level := Level,
     DroppedAttributesCount = maps:size(Attributes) - length(Attributes1),
     Flags = 0,
 
-    LogRecord = case  Metadata of
+    LogRecord = case Metadata of
                     #{otel_trace_id := TraceId,
                       otel_span_id := SpanId} ->
-                        #{trace_id => TraceId,
-                          span_id => SpanId};
+                        #{trace_id => from_hex_str(TraceId, 128),
```
tsloughter, line 93: Instead of here we should probably make it a binary in `otel_span` instead of using a list and converting here.

apps/opentelemetry_api/src/otel_ctx.erl
tsloughter: Hm, I think this is right but trying to think of why it might have been done this way in the first place. I probably just made a mistake, just worried there is a case of an existing active span that could be you don't want to overwrite when a undefined context is then used...

apps/opentelemetry_experimental/src/otel_otlp_logs.erl
SergeTupchiy: There is probably some benefit in keeping a hex string (list) in logger metadata if someone else is interested in it, as it has better readability. For example, I want to trace `"00-0226551413cd73a554184b324c82ad51-b7ad6b71432023a2-01"` and I know that it's added to logger metadata, so I can grep my standard handler logs by `0226551413cd73a554184b324c82ad51`  without the need to convert it to a binary.

tsloughter: By binary I meant `<<"00-0226551413cd73a554184b324c82ad51-b7ad6b71432023a2-01">>`

SergeTupchiy: That will still require conversion to 16 (for trace ID) and 8 (for span ID)  byte binaries here :thinking: 
```
 B = <<"0226551413cd73a554184b324c82ad51">>.
<<"0226551413cd73a554184b324c82ad51">>
2> <<(binary_to_integer(B, 16)):128>>.
<<2,38,85,20,19,205,115,165,84,24,75,50,76,130,173,81>>
```
As we need `<<2,38,85,20,19,205,115,165,84,24,75,50,76,130,173,81>>` for proto, not `<<"0226551413cd73a554184b324c82ad51">>`

bryannaegele: Having it as the string binary representation in the metadata means we only have to convert it during proto mapping versus forcing every logger backend formatter handle it as a special case. We've thus far only done things that proto needs in the proto mapper.

bryannaegele: Another bit around that is we're supposed to support other exporting formats, all of which take a string afaik.

tsloughter: Right.

tsloughter: I think I'll merge this. This PR doesn't make the values into lists itself so doesn't need to be the place to resolve that issue.

PR #640, opened 2023-10-14
Fix transmitted time units for logs

filmor:


bryannaegele (approved):

ferd (changes requested):
The old code wasn't safe, but the new one isn't either. The implementation of the opentelemetry timestamp we use relies on a `native` time unit: 

https://github.com/open-telemetry/opentelemetry-erlang/blob/2c45b1b8dc6616bf50b2724d8a166776630c68b7/apps/opentelemetry_api/src/opentelemetry.erl#L305-L318

The native time unit is system-dependent and returns the most accurate result the VM sees. Basically, the old version of the code assumes that `native` means `nanosecond` and the new one assumes `native` means `microsecond` but neither is actually guaranteed.

The time unit returned by the timestamp should specify its unit.

The conversion function here should be `erlang:convert_time_unit(OBservedTime, native, nanosecond)` since we generate a native timestamp earlier.

bryannaegele:
I assumed `ObservedTime` was the timestamp added by `:logger` which uses `os:system_time(microsecond)`. Introduced OTP 21.3. Was that an incorrect assumption?

https://www.erlang.org/doc/man/logger.html#timestamp-0

filmor:
Yes, `ObservedTime` is extracted from the log record (https://www.erlang.org/doc/man/logger#type-metadata), so it is always going to be in milliseconds. I've tested this on Linux where `native` would be nanoseconds and it created correct records on export (which it didn't do before :)).

ferd:
> Yes, `ObservedTime` is extracted from the log record (https://www.erlang.org/doc/man/logger#type-metadata), so it is always going to be in milliseconds. I've tested this on Linux where `native` would be nanoseconds and it created correct records on export (which it didn't do before :)).

Well it's set to nanoseconds from microseconds here, not milliseconds?

PR #644, opened 2023-10-24
fix(otel_metrics): use `opentelemetry:timestamp/0`

SergeTupchiy:
There was inconsistency in timestamps handling: metrics used `system_time(nanosecond)`, while metrics exporter treated these values as `opentelemetry:timestamp()` (monotonic time in native units). 

This PR fixes it by using `opentelemetry:timestamp()`, most of the diff is caused by renaming `*time_unix_nano` to `time` to avoid confusions.


SergeTupchiy:
@tsloughter , @bryannaegele, @ferd 
Could you please take a look at this PR?
Alternative fix would be to avoid converting timestamps in `otel_otlp_metrics`: https://github.com/open-telemetry/opentelemetry-erlang/blob/main/apps/opentelemetry_experimental/src/otel_otlp_metrics.erl#L89
But I think the proposed fix is better as it follows the same approach used for traces: https://github.com/open-telemetry/opentelemetry-erlang/blob/main/apps/opentelemetry_api/src/opentelemetry.erl#L305

bryannaegele:
This seems good? I haven't been involved in the metrics but this seems cleaner and more consistent.

tsloughter:
Thanks, yea, I think this looks right.

The only thing I'm debating, which can be a separate issue/pr, is should we really be using `monotonic_time` instead of just `system_time`... I don't think we ever use `opentelemetry:timestamp` anywhere it gets used to compare or find a diff. Maybe safest to allow it to be used for comparisons/diffs.

tsloughter (approved):

PR #659, opened 2023-11-18
bump grpcbox dep to 0.17.0

tsloughter:


tsloughter:
Well damn... grpcbox and chatterbox only support 24, 25, 26 and I didn't realize the latest release actually broke 23 because of that.

I guess I need to add support for 23 to them before we bump this.

bryannaegele (approved):

PR #669, opened 2023-12-22
don't export unit fields in metrics when the unit is undefined

RoadRunnr:
The term `undefined` is used for metrics where not unit is declared. When exporting such a metric, we should not set the OTLP unit field to the string "undefined" as that it not a valid unit.


tsloughter (approved):

tsloughter:
Thanks! Can you sign the cla?

RoadRunnr:
CLA is signed

PR #671, opened 2023-12-27
Improve batch overload protection and log handler

SergeTupchiy:
This is quite huge PR, a short summary:

- Common batch processor functionality is moved to a single module: `otel_batch_olp` (olp - for overload protection).
- `otel_batch_olp` implementation is quite different from the previous `otel_batch_processor` and deserves additional explanations:
    - First of all, the previous olp  implementation (`max_queue_size` implemented as a periodic ETS tab size check) couldn't react to the changes timely due to its periodic nature. Even if the check interval is quite low, e.g. 1s,  an abrupt burst of events (be it trace spans or log messages) can insert much more records to the ETS table than the configured `max_queue_size`. The issue can be mitigated by draining ETS tab up to `max_queue_size` during the export, but it would only confirm the fact that overload protection is not perfect. 
    - Batch processor heavily relies on persistent_term. Though the stored terms (atoms) should fit a machine word and, thus, their quite frequent (periodic) update during ETS table rotation should not have triggered global GC, it still can be quite expensive if the number of other terms is large (https://www.erlang.org/doc/man/persistent_term.html):
        > Storing or updating a term (using [put/2](https://www.erlang.org/doc/man/persistent_term.html#put-2)) is proportional to the number of already created persistent terms because the hash table holding the keys will be copied. In addition, the term itself will be copied.
    - Due to the above reasons, an alternative solution is proposed which is based on atomic counters: https://github.com/SergeTupchiy/opentelemetry-erlang/blob/improve-batch-overload-protection-and-log-handler/apps/opentelemetry/src/otel_batch_olp.erl#L83
        The performance of atomic ops is comparable to persistent_term (at least on my basic hardware, more benchmarks are welcome), they are meant to be mutable and the proposed olp mechanism ensures that the table doesn't grow beyond `max_queue_size` value.
- `otel_batch_olp` is re-used in `otel_batch_processor` and `otel_log_handler`
- ETS tab export is changed to eliminate possible data loss due to certain race conditions (discussed in: https://github.com/open-telemetry/opentelemetry-erlang/pull/641).
- ETS tab keys are changed from instrumentation_scope to trace id / log timestamp. These values have higher entropy, and thus are better fit to be used as keys (storing a large number of objects under the same key in ETS bag can degrade performance). This was also discussed in: https://github.com/open-telemetry/opentelemetry-erlang/pull/641.

### DIFF
--- a/apps/opentelemetry/src/otel_batch_olp.erl
+++ b/apps/opentelemetry/src/otel_batch_olp.erl
@@ -1,0 +1,602 @@
+%%%------------------------------------------------------------------------
+%% Copyright 2023, OpenTelemetry Authors
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
+%%
+%% http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing, software
+%% distributed under the License is distributed on an "AS IS" BASIS,
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+%%
+%% @doc Common overload protection implementation
+%% @end
+%%%-------------------------------------------------------------------------
+
+-module(otel_batch_olp).
+
+-behaviour(gen_statem).
+
+-export([init_conf/1,
+         start_link/2,
+         insert_signal/2,
+         change_config/3,
+         force_flush/1]).
+
+%% gen_statem
+-export([init/1,
+         callback_mode/0,
+         init_exporter/3,
+         idle/3,
+         exporting/3,
+         terminate/3]).
+
+-export_type([otel_batch_olp_config/0,
+              otel_batch_olp_state/0,
+              max_queue_size/0,
+              otel_timeout_ms/0]).
+
+-include_lib("kernel/include/logger.hrl").
+
+%% Metrics are implmented separately
+-type otel_signal() :: traces | logs.
+
+-type otel_batch_olp_config() ::
+        #{reg_name := atom(),
+          cb_module := module(),
+          otel_signal := otel_signal(),
+          max_queue_size := max_queue_size(),
+          exporting_timeout_ms := pos_integer(),
+          scheduled_delay_ms := pos_integer(),
+          exporter := otel_exporter:exporter_config(),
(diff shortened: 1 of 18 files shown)


tsloughter:
It has been a week so wanted to say a big thanks! And let you know I hoped to get around to reviewing this next week at the latest.

tsloughter:
Hey @SergeTupchiy, sorry I haven't gotten this merged yet! Can you update to resolve the merge conflicts?

SergeTupchiy:
Hi @tsloughter,
Sure, hopefully I will manage to do it later this week. 
Also, I'm planning to remove this commit https://github.com/open-telemetry/opentelemetry-erlang/pull/671/commits/bf08e05ea85f10f254f93ea630320c8512d55eb2 from this branch and fix the relevant test case, as it was rejected in another PR: https://github.com/open-telemetry/opentelemetry-erlang/pull/670

tsloughter:
Great, thanks. And yes, please remove that commit but don't close that PR bc I don't think a final decision has been made.

SergeTupchiy:
> Great, thanks. And yes, please remove that commit but don't close that PR bc I don't think a final decision has been made.
Rebased, dropped 'cleanup persistent terms' commit from this branch and added a workaround fix for the failing test: https://github.com/open-telemetry/opentelemetry-erlang/pull/671/files#diff-ca895028100a6d11170ba82e246533e4bcaf1ffa2b15cfce32777a56e7cf8af4R122-R124

SergeTupchiy:
> @SergeTupchiy how does this PR align with #637? Is it building on that one?

Yes, this one includes changes done in #637 and makes the latter unnecessary.

tsloughter:
@SergeTupchiy sorry this still hasn't gotten a real review. I really want to get it in but it is large and changes a lot since it doesn't just change the logs handler to add olp but changes the span batch processor too. I'm wondering about breaking it up into smaller PRs so we can move it forward in pieces. What do you think?

tsloughter:
Ooh, so I think I get your ets table lookup strategy now. At first it worried me because I recalled we used to do similar (2-tuple with which to use kept in a counter) but then moved to the persistent term but I think I see now that it sort of blends the changes we made with the use of a reg_name with the original 2-tuple design.

Because we use reg_name now we don't have to put pid in the name anymore to make them unique but then require a name lookup instead of storing in the config/state.

Really wish we already had benchmarks so we could at least guarantee this doesn't have any regressions. Thinking about it I don't see why it would, but always better to measure of course :). Maybe I need to do some quick work on that before we can change the span batch processor.

PR #675, opened 2024-01-22
Add docs to `*_processor` modules

whatyouhide:
This is work that I quickly chatter about with @tsloughter.

PR #676, opened 2024-01-22
Document `opentelemetry:get_tracer/*` functions

whatyouhide:
Similar to #675.

### DIFF
--- a/apps/opentelemetry_api/src/opentelemetry.erl
+++ b/apps/opentelemetry_api/src/opentelemetry.erl
@@ -211,6 +211,8 @@ module_to_application(Name, Version, SchemaUrl) ->
                         Acc#{M => {Name, Version, SchemaUrl}}
                 end, #{}, Modules).
 
+%% @doc Returns the default global tracer.
+%% Defaults to `{otel_tracer_noop, []}'.
 -spec get_tracer() -> tracer().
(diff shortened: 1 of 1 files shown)


apps/opentelemetry_api/src/opentelemetry.erl
```diff
@@ -211,6 +211,8 @@ module_to_application(Name, Version, SchemaUrl) ->
                         Acc#{M => {Name, Version, SchemaUrl}}
                 end, #{}, Modules).
 
+%% @doc Returns the default global tracer.
+%% Defaults to `{otel_tracer_noop, []}'.
```
tsloughter, line 215:
```suggestion
%% Without the `opentelemetry' application, or another SDK, this defaults to `{otel_tracer_noop, []}'.
```

Not sure if my suggestion makes is confusing or helpful to the user...

whatyouhide: I’m not sure what you mean by "without the `opentelemetry` application". Do you mean something like "without having the `opentelemetry` application in your dependencies"?

tsloughter: Right, as a dependency (and started).

tsloughter (approved):

PR #677, opened 2024-01-22
use a generation integer to lock writes during each reader checkpoint

tsloughter:
more details to come in a revise commit message. This is a broken commit, only delta explicit histograms work


tsloughter:
Note it also doesn't actually delete the aggregates yet.

tsloughter:
Ok, this is ready for real reviews.

The issue this set out to solve is delta dropping metrics from export when they aren't recorded to in the collection interval. To do this a generation counter was added to the key of each metric in the metrics table. That counter is incremented (for deltas) on each collection so that only the current set of metrics are used -- `0` generation is always used for cumulative metrics.

The last generation is always kept around until after the current collection in order to facilitate conversion of a cumulative instrument to a delta export because it involves grabbing the previous value.

tsloughter:
Probably going to move the code that decides on the Generation depending on `forget` out of each individual aggregator but undecided since that ties the aggregator interface to having the Generation and I'd love to have a solution not reliant on this counter.

ferd:
I think this can work nicely so long as the collection phase happens sequentially. If there can be more than one collection phase overlapping, the 'forget' step is going to clobber a previous collection. This is because I think this is somewhat similar to a simplification of MVCC semantics (where you'd need to 'forget' based on the oldest 'live' counter currently in use. This could be done by incrementing one counter at start of collection for writes as right now, while storing a list of generations free for clean-up when done—ignoring all failure management).

If the collection step isn't structurally enforced by being done by a single process (per reader?), grabbing a name or lock of some sort would be required to get the equivalent semantics.

> Probably going to move the code that decides on the Generation depending on `forget` out of each individual aggregator but undecided since that ties the aggregator interface to having the Generation and I'd love to have a solution not reliant on this counter.

What you could do is tie an opaque term to each key, with an interface that enforces the behavior you want:
- acquire the opaque term before writing (from a writer-specific function)
- acquire the opaque term before collecting (from a collector-specific function)
- return the opaque term when done collecting (from a collector-specific function) which could then do the deletion.

These last two elements somewhat fit the `with_span` approach of the lib, and so they could all fit within a functional interface that ensures the right params are passed and the right semantics are enforced, kind of like with transactions.

---
Finally I'm thinking of a test for something like this that could verify the concurrency mechanisms:

1. start a collector process that just collects in a loop with a `receive` for a `done` message
2. start N writer processes that write values M times, with a call to `erlang:yield()` after each write
3. when the writers are done, send `done` to the collector, which returns a last collection, adds up the sum and batches of all values.
4. the test validates that the sum matches `N*M` such that all writes are accounted for, no more no less.
5. optionally make sure that all 'forget' steps have run and the tables are clean.

The yielding forces an increase in concurrency across operations and makes it more likely to show if any interleaving causes issues with the handling.

tsloughter:
To the first part: yes, it is sequential. The generation is controlled by the Reader which is a single process that will run the collections.

ferd (approved):
Should be alright given the sequential implementation; I'm assuming that further tests and/or extraction of patterns as discussed will be follow-up work.

PR #681, opened 2024-02-04
Add docs to `otel_ctx` and `otel_baggage`

whatyouhide:
Moar docs 🙃

### DIFF
--- a/apps/opentelemetry_api/src/otel_ctx.erl
+++ b/apps/opentelemetry_api/src/otel_ctx.erl
@@ -16,6 +16,10 @@
 %% are associated with a particular Trace or set of Baggage.
 %% `OpenTelemetry.Tracer' and `OpenTelemetry.Baggage' handle updating
 %% the Context.
+%%
+%% Functions in this module include variants that explicitly take a `Ctx'
+%% argument and variants that implicitly use the <i>current context</i>, which is
+%% the context stored in the process dictionary.
 %% @end
 %%%-------------------------------------------------------------------------
 -module(otel_ctx).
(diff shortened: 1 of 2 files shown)


apps/opentelemetry_api/src/otel_ctx.erl
```diff
@@ -41,11 +45,17 @@
          text_map_injector_fun/3]).
 
 -type t() :: map() | undefined.
+%% A context map.
+
 -type key() :: term().
+%% A context key.
```
tsloughter, line 51: For consistency with the rest of the project the comment about the type should go above the type, not below it.

tsloughter:
Looks good except for my comment about the comments going above the definitions.

tsloughter:
This reminded me we really need macros for Ctx and Baggage. I dislike that we have any functions that work on the pdict instead of explicit context :(. I think it'd be clearer if it was only macros that ever read the context from the pdict. Then those functions would be deprecated.

Of course, not something that has to be addressed in this PR. Though I am curious about your thoughts on that.

apps/opentelemetry_api/src/otel_ctx.erl
whatyouhide: @tsloughter "above the type" doesn't show up in edoc, that's why I went with below the type.

This is with the comment above the type:

CleanShot 2024-02-05 at 12 11 49@2x

tsloughter: Argh oops, ok, haha, nevermind then :)

tsloughter (approved):

whatyouhide:
@tsloughter if we're only ever meant to get the context from the process dictionary then yes, would be nice to deprecate the "explicit context" functions. I’m pretty much always for smaller API surfaces 😄

tsloughter:
Oh na, other way around. Deprecate the ones that do implicit context and only do that in macros that then call the functions with the context.

whatyouhide:
@tsloughter ah, even better then yes. 🙃

PR #684, opened 2024-02-04
Update and delete conventions following spec

joaothallis:
Create following the BREAKINGs of OTEL spec

https://github.com/open-telemetry/opentelemetry-specification/blob/main/CHANGELOG.md


tsloughter:
Cool. Was this done with the generation script? 

The app version would need to change too.

But I think this upgrade in semconv is still blocked for us by http stuff. @bryannaegele ?

I should probably try to knock that out before I start working again so we can upgrade semconv...

joaothallis:
> Cool. Was this done with the generation script?

No, I applied the changes manually verifying the spec. Is that [this script](https://github.com/open-telemetry/opentelemetry-erlang/blob/main/apps/opentelemetry_semantic_conventions/generate.sh)?

bryannaegele:
> But I think this upgrade in semconv is still blocked for us by http stuff. @bryannaegele ?

We still have to implement the env var checking. I don't know how we're going to deal with the new plus old. I threw out running the script, diffing, and adding deprecated stuff but that had a host of problems.

I know this is still an issue the wg knows of in general with no progress as yet. I think there are several other sigs frustrated by it.

tsloughter:
Ugh, we have got to do something about being stuck on old semconv. Next time we meet, @bryannaegele , lets be sure to write up a detailed issue and I'll do my best to knock it out. I keep losing context on the issue and feeling like I have to start over figuring out just what to do.

bryannaegele:
@joaothallis this is being addressed in https://github.com/open-telemetry/opentelemetry-erlang/pull/733

PR #686, opened 2024-02-06
Add docs to `otel_attributes`

whatyouhide:
Also had this one lying around.

### DIFF
--- a/apps/opentelemetry_api/src/otel_attributes.erl
+++ b/apps/opentelemetry_api/src/otel_attributes.erl
@@ -14,7 +14,7 @@
 %%
 %% @doc Functions to work with Attributes.
 %%
-%% An Attribute is a key-value pair with string keys.
+%% An Attribute is a key-value pair with string or atom keys.
 %% See <a href="https://opentelemetry.io/docs/specs/otel/common/#attribute">the specification</a>.
 %% @end
 %%%-------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


apps/opentelemetry_api/src/otel_attributes.erl
```diff
@@ -145,6 +191,7 @@ is_valid_atom_value(nil) ->
 is_valid_atom_value(Value) ->
     is_atom(Value) andalso (is_boolean(Value) == false).
 
+%% @private
```
whatyouhide, line 194: Given the signature, I think we want this to not be part of the public API? This works on the raw attribute map, not the `#attribute{}` record.

tsloughter: Yea.

apps/opentelemetry_api/src/otel_attributes.erl
```diff
@@ -55,19 +67,40 @@ new(_, CountLimit, ValueLengthLimit) ->
                 dropped=0,
                 map=#{}}.
 
-set(NewList, Attributes) when is_list(NewList) ->
-    set(maps:from_list(NewList), Attributes);
+%% @doc Sets the given key-value pairs in the given `Attributes'.
+%%
+%% `NewListOrMap' can be a list of key-value pairs or a map. If `NewListOrMap' is not a list
+%% or map, the function returns `Attributes' as is. Returns the updated `Attributes'.
```
tsloughter, line 73:
```suggestion
%% or map, the function returns `Attributes' as is, otherwise returns the updated `Attributes'.
```

apps/opentelemetry_api/src/otel_attributes.erl
```diff
@@ -55,19 +67,40 @@ new(_, CountLimit, ValueLengthLimit) ->
                 dropped=0,
                 map=#{}}.
 
-set(NewList, Attributes) when is_list(NewList) ->
-    set(maps:from_list(NewList), Attributes);
+%% @doc Sets the given key-value pairs in the given `Attributes'.
```
tsloughter, line 70:
```suggestion
%% @doc Sets the given key-value pairs in the given `Attributes'. Overrides 
%% the existing value for a given key if it already exists in `Attributes'.
```

tsloughter (approved):

PR #692, opened 2024-02-13
add exemplar reservoir support

tsloughter:
Opening as a draft because it only has SimpleFixed reservoir and it has no configuration for whether to use exemplars and how to filter.

Only the drop and simple fixed exemplar reservoir are supported by this PR, so patchstograms do not yet have exemplar support.

This patch also does not have a toggle for if exemplars are to be taken or not.


apps/opentelemetry_experimental/src/otel_aggregation_histogram_explicit.erl
```diff
@@ -301,20 +314,21 @@ datapoint(CollectionStartTime, #explicit_histogram_aggregation{
...
-                                 }) ->
+datapoint(ExemplarReservoir, ExemplarsTab, CollectionStartTime, #explicit_histogram_aggregation{
+                                                                   key=Key={_, Attributes, _, _},
+                                                                   explicit_bucket_boundaries=Boundaries,
+                                                                   checkpoint=#explicit_histogram_checkpoint{bucket_counts=BucketCounts,
+                                                                                                             min=Min,
+                                                                                                             max=Max,
+                                                                                                             sum=Sum,
+                                                                                                             start_time=StartTime}
+                                                                  }) ->
```
ferd, line 330: the indentation is unreadable here without zooming out on my display

ferd (commented):
At this point I have no significant comments, this all looks okay on the form. To properly evaluate the function though, I sort of need more background on examplars; can you direct me to the spec on these?

ferd (approved):

apps/opentelemetry_experimental/src/otel_aggregation_histogram_explicit.erl
tsloughter: Cleaned it up.

ferd (approved):

PR #709, opened 2024-03-13
make the console metric exporter handle all values

RoadRunnr:
The OTLP exporter handles all types of keys, but the debug exporter is a bit too simplistic and will choke on atoms as attribute keys.

Convert all values to something sensible for printing on the console.

### DIFF
--- a/apps/opentelemetry_experimental/src/otel_metric_exporter_console.erl
+++ b/apps/opentelemetry_experimental/src/otel_metric_exporter_console.erl
@@ -80,8 +80,12 @@ print_histogram_datapoint(Name, #histogram_datapoint{
     AttributesString = attributes_string(Attributes),
     io:format("~s{~s} ~p~n", [Name, AttributesString, Buckets]).
 
+attr_str(V) when is_list(V); is_binary(V) -> V;
+attr_str(V) when is_atom(V) -> atom_to_list(V);
+attr_str(V) -> io_lib:format("~0tp", [V]).
+
 %% need to handle non-string values
 attributes_string(Attributes) ->
     lists:join(", ", maps:fold(fun(K, V, Acc) ->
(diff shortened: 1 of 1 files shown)


apps/opentelemetry_experimental/src/otel_metric_exporter_console.erl
```diff
@@ -80,8 +80,12 @@ print_histogram_datapoint(Name, #histogram_datapoint{
     AttributesString = attributes_string(Attributes),
     io:format("~s{~s} ~p~n", [Name, AttributesString, Buckets]).
 
+attr_str(V) when is_list(V); is_binary(V) -> V;
+attr_str(V) when is_atom(V) -> atom_to_list(V);
+attr_str(V) -> io_lib:format("~0p", [V]).
```
ferd, line 85: this specific attribute will suffer on unicode content unless we pass in `~0tp`:

```
1> io:format("~0p~n", [{"😀"}]).
{[128512]}
ok
2> io:format("~0tp~n", [{"😀"}]).
{"😀"}
ok
```

RoadRunnr: Yes, I know. I was concerned that the minimum support Erlang release (22) might not have the `t` modifier. But I've check the docs and it does have it. So, I'll add it.

ferd (approved):
I'm assuming validation of keys prevents us from using tuples and other values at this point, though maybe @tsloughter wants to double-check.

tsloughter:
@ferd right.

RoadRunnr:
rebased to main, @ferd @tsloughter is there anything else missing that prevents merging this?

ferd:
yeah it needed a rebase, which is now done.

PR #717, opened 2024-03-26
add otel_span:start_config typespec and use it in the SDK

### LINKED ISSUE #708: broken typespec for otel_span:start_opts
@tsloughter commit e4c2c109 authored by you broke the typespec for the `Opts` argument of `otel_tracer:with_span` by changing all map members to mandatory, however in the API all of them are optional.

Declaring all map members as mandatory might be appropriate when the type is used for SDK functions, but for the API function all options are indeed optional. Please consider changing the type spec back and or use a different type for the SDK functions.

RoadRunnr:
fixes #708

In the API functions, all members of start_opts() are optional.

Before passing the options on the SDK all options are set to sensible
defaults. That makes is possible to have a dedicated start_config()
type where all values are mandatory and optimize the usage of the
settings slightly.

### DIFF
--- a/apps/opentelemetry/src/otel_span_ets.erl
+++ b/apps/opentelemetry/src/otel_span_ets.erl
@@ -55,7 +55,7 @@ start_link(Opts) ->
 
 %% @doc Start a span and insert into the active span ets table.
 -spec start_span(otel_ctx:t(), opentelemetry:span_name(), otel_sampler:t(), otel_id_generator:t(),
-                 otel_span:start_opts(), fun(), otel_tracer_server:instrumentation_scope() | undefined)
+                 otel_span:start_config(), fun(), otel_tracer_server:instrumentation_scope() | undefined)
                 -> opentelemetry:span_ctx().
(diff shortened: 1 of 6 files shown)


tsloughter:
Trying to decide if this is ok. The API already sort of makes a requirement of the SDK to accept start_opts so making a more restrictive version seems reasonable, just weird to essentially put an SDK type in the API.

RoadRunnr:
> Trying to decide if this is ok. The API already sort of makes a requirement of the SDK to accept start_opts so making a more restrictive version seems reasonable, just weird to essentially put an SDK type in the API.

Anything I can help with to get this approved/merged?

tsloughter:
Thanks for pinging. I did have one idea that is just a nitpick. Could you rename `sdk_start_opts` to `start_config`. They are no longer "optional" so makes sense to call it the config and I think the name makes more sense.

Do you think that makes sense?

RoadRunnr:
make sense to me, type renamed to `start_config` and commit message adjusted accordingly.

tsloughter (approved):

PR #720, opened 2024-04-01
move metrics and logs to separate exporter modules from traces

tsloughter:
Splitting traces, metrics and logs OTLP exporters across modules.

Opening as a draft since there is still a lot of duplication that can be resolved before merge.


tsloughter:
Tests pass so marking ready for review but I think it still needs some tweaks and added tests.

tsloughter:
Main motivation was the pain in getting setting the url correctly. Not only did it now need to change `init` to require a type argument but it just made the configuration code awkward. Just having them be separate modules greatly simplifies this. 

I don't expect them to diverge beyond configuration. So I should be able to have most everything in a common module.

ferd:
Sounds fair enough. Config variation absolutely makes sense there.

tsloughter:
@ferd I updated to move common functionality to a shared module.

apps/opentelemetry_experimental/src/otel_metric_exporter_console.erl
```diff
@@ -19,18 +19,16 @@
 -module(otel_metric_exporter_console).
 
 -export([init/1,
-         export/4,
+         export/3,
          force_flush/0,
          shutdown/0]).
 
--include_lib("opentelemetry_api_experimental/include/otel_metrics.hrl").
--include("otel_view.hrl").
```
ferd, line 27: should this module declare implementing the otel_exporter_metrics behavior?

ferd (approved):

apps/opentelemetry_experimental/src/otel_metric_exporter_console.erl
tsloughter: yup, updating

tsloughter:
Fix for this has been pushed. `force_flush` now uses `call_collect`.

For others: The issue was after the `force_flush`, unlike in other tests for metrics where we do a `receive` before any more metrics are recorded, it immediately does new recordings and then bumps a local Generation variable to 1 to check the ETS table. Unless everything had already run in `collect` before getting to the measurements with exemplars the Generation would still be 0 and not show up in the ETS table under what the test expected the Generation to be. Fixe by making `force_flush` call a synchronous `collect`.

ferd (approved):

PR #721, opened 2024-04-01
Add more context docs

whatyouhide:
cc @tsloughter 

Baby steps 🙃

### DIFF
--- a/apps/opentelemetry_api/src/otel_ctx.erl
+++ b/apps/opentelemetry_api/src/otel_ctx.erl
@@ -18,11 +18,8 @@
 %% the Context.
 %%
 %% For more information on what the Context is and what is its purpose,
-%% see the <a href="https://opentelemetry.io/docs/concepts/context-propagation/">
+%% see the <a href="https://opentelemetry.io/docs/specs/otel/context/#overview/">
 %% OpenTelemetry documentation</a>.
-%%
-%% Contexts are propagated around the system using <b>propagators</b>.
-%% See the {@link otel_propagator} behaviour for more information.
 %%
 %% Functions in this module include variants that explicitly take a `Ctx'
 %% argument and variants that implicitly use the <i>current context</i>, which is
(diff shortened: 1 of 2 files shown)


apps/opentelemetry_api/src/otel_ctx.erl
```diff
@@ -12,11 +12,18 @@
 %% See the License for the specific language governing permissions and
 %% limitations under the License.
 %%
-%% @doc Ctx is responsible for propagating values within a process that
+%% @doc The Context is responsible for propagating values within a process that
 %% are associated with a particular Trace or set of Baggage.
 %% `OpenTelemetry.Tracer' and `OpenTelemetry.Baggage' handle updating
 %% the Context.
 %%
+%% For more information on what the Context is and what is its purpose,
```
bryannaegele, line 20: This section is a bit misleading and an area where we'd like to put more clarity to. The context is a general purpose object for state. The propagator behaviour is arguably just for propagation across services. It's also not limited to use in a single process which the old language also implies.

https://opentelemetry.io/docs/specs/otel/context/#overview is the better link for what context is used for. We don't use that behaviour for any propagation within an application here or in any of the contrib libraries, so `propagated around the system` is ambiguous since context is propagated across process boundaries without any translation steps.

whatyouhide: Gotcha, updated the link and removed some wording @bryannaegele!

tsloughter: Yea, actually, these docs were phrased poorly in the first place.

I would not put it as "associated with a particular Trace or set of Baggage". Not only does this miss the fact that it is used by metrics and logs (and eventually events), it removes the general purpose nature that context technically has. Since it is currently part of the otel library -- I'd argue we should consider breaking it out -- it basically is tied to just those, but its not meant to be.

whatyouhide:
I'll rework this PR.

PR #722, opened 2024-04-01
Group modules in the docs for opentelemetry_api

whatyouhide:
Looks like this:

CleanShot 2024-04-01 at 21 47 24@2x

We can do this in other apps too. cc @tsloughter


tsloughter (approved):
Very nice, didn't know you could do this :)

PR #723, opened 2024-04-01
Make the otel_utils module private

whatyouhide:


ferd (approved):
No docs, omitting generating docs for the module makes sense. You'll seemingly need to rebase on `main` to make this mergeable though.

Issue #557, opened 2023-03-12
Suggestions for docs to work on


tsloughter:
I'm opening this issue both to list some of my thoughts on needed docs and to solicit suggestions on what docs people would like to see. These apply both to the website's Erlang docs, https://github.com/open-telemetry/opentelemetry.io/tree/main/content/en/docs/instrumentation/erlang, and hex.pm docs, mainly focusing on the API https://hexdocs.pm/opentelemetry_api/readme.html and SDK https://hexdocs.pm/opentelemetry/readme.html

- set_status 
  - [ ] API docs
  - [ ] website example
- [ ] samplers: website example

I'll keep adding as I find more.

albertored:
I think a big step forward would be to define a list of modules/functions that we want to be exposed to users, add a description to all of them and hide from hex.pm all other internal functions since at the moment all modules/functions are exposed and mostly of them are without any description.

tsloughter:
Agreed.

whatyouhide:
I’m picking this back up as I think right now it's the biggest issue with `opentelemetry-erlang` docs: it's really hard to figure out what's the API that end users are supposed to use.

So, for Erlang in `opentelemetry_api`, I think what we want to be public is:

- `opentelemetry`
- `otel_attributes`
- `otel_baggage`
- `otel_ctx`
- `otel_span`
- `otel_tracer`

What about the `otel_propagator*` ones? Or `otel_tracer_noop`/`otel_tracer_provider`? I don't want to go too deep into this... small steps 🙃

tsloughter:
I think all propagator ones need to be shown. It is a little awkward since only `otel_propagator_text_map` is to be called but the others give you the name of the module to use when configuring which propagator to use.

The tracer provider should be public too. We just need better docs around what it is and how to (rarely) use it bc it is used indirectly in our case. Technically you get a Tracer from calling TracerProvider and then use that Tracer to start a span, but we macro around that for the common case of wanting a Tracer from the global Provider named for the current Application.

`otel_tracer_noop` shouldn't be called by the user but may be used in configuration, so that is a tossup.

whatyouhide:
@tsloughter good call. We can easily show all the propagator modules and `otel_tracer_noop`, just hide all the functions in them, so that it's clear that users are not meant to *call* them?

tsloughter:
Sounds good.

PR #726, opened 2024-04-03
Add docs for all `otel_propagator*` modules

whatyouhide:
cc @tsloughter

This is some work towards https://github.com/open-telemetry/opentelemetry-erlang/issues/557.

### DIFF
--- a/apps/opentelemetry_api/src/otel_propagator_text_map.erl
+++ b/apps/opentelemetry_api/src/otel_propagator_text_map.erl
@@ -23,13 +23,13 @@
 %% {text_map_propagators, [trace_context, baggage]},
 %% '''
 %%
-%% The propagators are then used at the points that cross service
-%% communication is performed. By default `inject' and `extract' work on a
-%% generic list of 2-tuple's with binary string keys and values. A user
-%% defined function for setting a key/value in the carrier and for getting
-%% the value of a key from a carrier can be passed as an argument. For
-%% example, injecting and extracting to and from Hackney headers could be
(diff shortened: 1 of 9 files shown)


apps/opentelemetry_api/src/otel_propagator_text_map.erl
```diff
@@ -111,65 +112,87 @@
...
 
+%% @doc Extracts the current context from the provided `Carrier' using the
+%% given `Propagator' and functions to get all the keys and get the keys from the carrier.
 -spec extract(otel_propagator:t(), otel_propagator:carrier(), fun(), fun()) -> otel_ctx:token().
 extract(Propagator, Carrier, CarrierKeysFun, CarrierGetFun) ->
     Context = otel_ctx:get_current(),
     Context1 = extract_to(Context, Propagator, Carrier, CarrierKeysFun, CarrierGetFun),
     otel_ctx:attach(Context1).
 
+%% @private
```
tsloughter, line 184: The `_to` and `_from` functions shouldn't be private. If I could do it all over again these would be the only functions and the ones working on the pdict would be gone. So if anything.. those should be private, hehe, but probably not a good idea either.

whatyouhide: Ah okay, gotcha, I thought we wanted to move forward with these versions. Let's not make anything private for now then, I will push an update soon.

whatyouhide: @tsloughter done!

tsloughter (approved):

PR #727, opened 2024-04-05
Documentation for `otel_resource*` modules

whatyouhide:
Documents all those modules, how to use them, their functions, and their relationship.

### DIFF
--- a/apps/opentelemetry/src/otel_resource.erl
+++ b/apps/opentelemetry/src/otel_resource.erl
@@ -107,8 +107,8 @@ attributes(_) ->
 
 %% @doc Returns `true' if `Key' is valid and part of the given resource.
 -spec is_key(key(), t()) -> boolean().
-is_key(Key, #resource{attributes=Attributes}) ->
-    case try_check_key(Key, false) of
+is_key(K, #resource{attributes=Attributes}) ->
+    case try_check_key(K, false) of
         {true, Key} ->
             maps:is_key(Key, otel_attributes:map(Attributes));
         _ ->
--- a/apps/opentelemetry/src/otel_resource.erl
+++ b/apps/opentelemetry/src/otel_resource.erl
@@ -14,11 +14,19 @@
 %%
 %% @doc A Resource is attributes representing the entity producing
 %% telemetry. For example, by default the language (Erlang), name of this
-%% library (opentelemetry) and version of this library are included in
+%% library (`opentelemetry'), and version of this library are included in
 %% the Resource.
 %%
 %% This module provides the functional interface for working with the
 %% resource record.
+%%
+%% The `opentelemetry' library supports <i>resource detectors</i> to
+%% detect attributes to include in the Resource. See {@link otel_resource_detector}
+%% for the behaviour to detect resources, and the {@link otel_resource_app_env}
+%% and {@link otel_resource_env_var} modules for built-in implementations.
+%%
+%% See the <a href="https://opentelemetry.io/docs/concepts/resources/">
+%% OpenTelemetry Resource documentation</a> for more information on Resources.
 %% @end
 %%%-----------------------------------------------------------------------
 -module(otel_resource).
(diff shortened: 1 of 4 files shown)


apps/opentelemetry/src/otel_resource_env_var.erl
```diff
@@ -12,8 +12,15 @@
...
 %%
-%% @doc
+%% @doc Resource detector ({@link otel_resource_detector}) which adds attributes
+%% to the `Resource' based on environment variables.
 %%
+%% This resource detector reads the `OTEL_RESOURCE_ATTRIBUTES' environment
+%% variable and parses it as a comma-separated list of key-value pairs. For
+%% example, `key1=val1,key2=val2'.
+
+%% To use this module, add it to the list of `resource_detectors' in the
```
tsloughter, line 22: It is on by default so I think this should be reworded.

apps/opentelemetry/src/otel_resource_detector.erl
```diff
@@ -15,12 +15,16 @@
...
+%% {@link otel_resource:t()} made from those attributes.
 %%
-%% The state machine will spawn a process for each detector and collect the
-%% results of running each and merge in the order they are defined. Once in
-%% the `ready' state it will reply to `get_resource' calls with the final
-%% `otel_resource:t()'.
+%% This module is meant for users who intend to write their own resource
+%% detectors.
+%%
+%% This module is a state machine (started by the `opentelemetry' application)
```
tsloughter, line 23:
```suggestion
%% This behaviour is a state machine (started by the `opentelemetry' application)
```

apps/opentelemetry/src/otel_resource_app_env.erl
```diff
@@ -12,13 +12,23 @@
...
-%% Results in the `Resource' attributes `service.name' and `service.namespace'.
+%% ```
+%% [{service, #{name => "service-name",
+%%              namespace => "service-namespace"}]
+%% '''
+%%
+%% then it results in the `Resource' attributes `service.name' and `service.namespace'
+%% set to `service-name' and `service-namespace' respectively.
+%%
+%% To use this module, add it to the list of `resource_detectors' in the
```
tsloughter, line 30: On by default so this should also be reworded.

whatyouhide:
@tsloughter changes applied!

apps/opentelemetry/src/otel_resource_app_env.erl
```diff
@@ -12,13 +12,23 @@
...
+%% configuration under the `resource' key:
 %%
-%% Results in the `Resource' attributes `service.name' and `service.namespace'.
+%% ```
+%% [{service, #{name => "service-name",
+%%              namespace => "service-namespace"}]
+%% '''
+%%
+%% then it results in the `Resource' attributes `service.name' and `service.namespace'
+%% set to `service-name' and `service-namespace' respectively.
```
tsloughter, line 28:
```suggestion
%% set to `service.name' and `service.namespace' respectively.
```

apps/opentelemetry/src/otel_resource_detector.erl
```diff
@@ -36,6 +40,8 @@
          handle_event/4]).
 
 -callback get_resource(term()) -> otel_resource:t().
+%% Should return a resource identified by the given term. For example,
+%% the term could be the name of an environment variable.
```
tsloughter, line 44:
```suggestion
%% Function that takes the configuration of a resource detector 
%% and returns a resource.
```

apps/opentelemetry/src/otel_resource.erl
```diff
@@ -69,21 +88,27 @@ create(List, SchemaUrl) when is_list(List) ->
...
     Attributes;
 attributes(_) ->
     undefined.
 
+%% @doc Returns `true' if `Key' is valid and part of the given resource.
 -spec is_key(key(), t()) -> boolean().
-is_key(K, #resource{attributes=Attributes}) ->
-    case try_check_key(K, false) of
+is_key(Key, #resource{attributes=Attributes}) ->
+    case try_check_key(Key, false) of
```
tsloughter, line 111:
```suggestion
is_key(K, #resource{attributes=Attributes}) ->
    case try_check_key(K, false) of
```

`try_check_key` will convert to a binary, so either need to leave this as `K` or change `{true, Key}` to `{true, Key1}`.

apps/opentelemetry/src/otel_resource_app_env.erl
whatyouhide: @tsloughter mmm I think this is not correct? `service.name` gets set to `service-name` and `service.namespace` gets set to `service-namespace`, no?

apps/opentelemetry/src/otel_resource.erl
whatyouhide: Oh man yeah my bad. Will do it in another pass, no need for this now. 👍

apps/opentelemetry/src/otel_resource_app_env.erl
tsloughter: ooooo, I was missing the context in the diff :). Mind changing those names to something other than being the exact same as the key but with `-` instead of `.`?

tsloughter: like just `myservice` and `mynamespace`

whatyouhide: yes good call. 👍

whatyouhide: Done @tsloughter!

tsloughter (approved):

PR #728, opened 2024-04-07
Group and hide modules in "opentelemetry" app

whatyouhide:
- Hid several modules (`_utils` and friends) from the public API
- Grouped mods into sections in the docs

CleanShot 2024-04-07 at 16 24 37@2x

### DIFF
--- a/apps/opentelemetry/docs.config
+++ b/apps/opentelemetry/docs.config
@@ -2,3 +2,24 @@
 {extras, [<<"apps/opentelemetry/README.md">>, <<"apps/opentelemetry/LICENSE">>, <<"VERSIONING.md">>]}.
 {main, <<"readme">>}.
 {proglang, erlang}.
+{groups_for_modules, [
+    {'Span', [otel_span_ets,
+              otel_span_limits,
+              otel_span_sweeper]},
+    {'Span Processing', [otel_span_processor,
+                         otel_simple_processor,
+                         otel_batch_processor]},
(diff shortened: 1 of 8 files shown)


ferd (approved):

whatyouhide:
cc @tsloughter this should be GTG 🙃

apps/opentelemetry/docs.config
```diff
@@ -2,3 +2,24 @@
 {extras, [<<"apps/opentelemetry/README.md">>, <<"apps/opentelemetry/LICENSE">>, <<"VERSIONING.md">>]}.
 {main, <<"readme">>}.
 {proglang, erlang}.
+{groups_for_modules, [
+    {'Span', [otel_span_ets,
+              otel_span_limits,
```
tsloughter, line 7: This should maybe be private but I'm going to merge this and we can decide later.

tsloughter: In the case of the SDK it is tougher to decide what should/shouldn't be public but I think `otel_span_ets` is too much of how the internals work to be good for someone writing an extension to use. In a perfect world it would be transparent to both API user and user who writes some sort of extension how the spans are stored.

tsloughter (approved):

apps/opentelemetry/docs.config
tsloughter: oops that comment was supposed to just be on the line with `otel_span_ets`.

PR #729, opened 2024-04-09
Document sampler modules

whatyouhide:


apps/opentelemetry/src/otel_sampler.erl
```diff
@@ -100,5 +126,6 @@ should_sample({Sampler, _, Config}, Ctx, TraceId, Links, SpanName, Kind, Attribu
             Result
     end.
 
+%% @private
```
tsloughter, line 129:
```suggestion
```

apps/opentelemetry/src/otel_sampler.erl
```diff
@@ -59,15 +76,23 @@
...
 -type sampling_result() :: {
                             sampling_decision(),
                             opentelemetry:attributes_map(),
                             opentelemetry:tracestate() | otel_tracestate:members()
                            }.
+
 -opaque t() :: {module(), description(), sampler_opts()}.
+%% A sampler.
 
+%% @private
```
tsloughter, line 95: Torn on if this should be private. I guess for now it is best to have it private but there are places like this where we require configuration instead of just allowing the user to pass a sampler "object" that maybe we should allow either. For now this is good, just something I wanted to note for the future.

tsloughter (approved):

apps/opentelemetry/src/otel_sampler.erl
whatyouhide: Yeah I’m always for exposing less initially and then making more funs private rather than the other way around 👍

apps/opentelemetry/src/otel_sampler.erl
whatyouhide: This should be part of the public API?

whatyouhide: @tsloughter if so, I’m not sure how to document this as I’m not sure about the use case. Suggestions?

apps/opentelemetry/src/otel_sampler.erl
```diff
@@ -12,9 +12,16 @@
...
-%% A sampler is a function run on each started span that returns whether to
-%% record and propagate, only record or not record the span.
+%% @doc Behaviour for defining samplers.
+%%
+%% A sampler should provide a function run on each started span that returns whether to
+%% record and propagate, only record, or not record the span.
+%%
+%% For more information on the concept of <i>Sampling</i>, see
+%% <a href="https://opentelemetry.io/docs/concepts/sampling/">Sampling in the OpenTelemetry
+%% documentation</a>. For examples of configuring samplers or implementing your own sampler,
```
bryannaegele, line 22: I'd link to the spec, as well, for docs on what all the various things mean. https://opentelemetry.io/docs/specs/otel/trace/sdk/#sampling

The various built-ins and concepts are all linkable.

apps/opentelemetry/src/otel_sampler.erl
bryannaegele: Description is required in the spec. Every sampler must have one.

apps/opentelemetry/src/otel_sampler_always_off.erl
```diff
@@ -12,8 +12,7 @@
 %% See the License for the specific language governing permissions and
 %% limitations under the License.
 %%
-%% @doc
-%% This sampler always decides to neither record nor sample each span.
+%% @doc An {@link otel_sampler} that drops all spans.
```
bryannaegele, line 15: You could link to the docs for each of the built-ins https://opentelemetry.io/docs/specs/otel/trace/sdk/#built-in-samplers

apps/opentelemetry/src/otel_sampler.erl
whatyouhide: Sounds good, added!

apps/opentelemetry/src/otel_sampler.erl
whatyouhide: I’m confused about this: the *callback* is documented in `otel_sampler`. Do we need to document `description/1` in every sampler too?

tsloughter: This is the function to get the description of any sampler. So if a user has a sampler record they would call this function to get its description.

whatyouhide: @tsloughter right but they know that they can do that because all the samplers implement the `otel_sampler` behaviour, where the `description/1` callback is documented. No?

tsloughter: The user would call `otel_sampler:description(Sampler)`, not the individual implementation module's implementation.

tsloughter (approved):

PR #733, opened 2024-04-21
Sem con 1.27.0

bryannaegele:
With the new layout everything common has been extracted out by the wg to a registry. From that, there's the `attribute_group` model which I chose to just name `common` instead. `trace` moved to `span` now which I guess makes sense but I could go either way with overriding that name.

Everything is a breaking change, so we have some freedom here. I attempted the single file approach Java did but the docs just get really confusing about what's common, trace-specific, and metric-specific.

Erlang is not done. I think once the hrl template is fixed up and verify we have everything being generated in the right namespaces, we can get it out there and continue to refine the docs and add guides. That should unblock all the http, bandit, phoenix work to be done.

I also switched from the shell script to an elixir script (`elixir generate.exs`) to parallelize the builds.

image
image


tsloughter:
Should this be a draft PR and we not merge it until I can send you a patch with the Erlang update so they stay in sync?

bryannaegele:
Yeah. You can move it to draft and work on this branch directly. I won't modify anything while you work.

bryannaegele:
I'm open to feedback and discussion on namespacing and multiple files (semantic attributes and semantic resource attributes). It doesn't appear there's anything consistent being employed, so there's nothing prescriptive to guide or bind us.

* Java appears they've also abandoned the 2 file approach in the past month to [file/class per type](https://github.com/open-telemetry/semantic-conventions-java/tree/main/semconv/src/main/java/io/opentelemetry/semconv) who have taken an additional step to move everything experimental to an [`incubating` package](https://github.com/open-telemetry/semantic-conventions-java/tree/main/semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating) with a file/class per type. It 
* [JS](https://github.com/open-telemetry/opentelemetry-js/tree/main/packages/opentelemetry-semantic-conventions) (2 file approach)
* [Python](https://github.com/open-telemetry/opentelemetry-python/tree/main/opentelemetry-semantic-conventions/src/opentelemetry/semconv) is doing 3 namespaces (trace, metrics, resource) with no common file
* DotNet seems to be hand-curated and possibly out of date like us.

bryannaegele:
https://github.com/open-telemetry/semantic-conventions/pull/953

tsloughter:
@bryannaegele can you add to the readme how to run generate.ex

tsloughter:
So JS has no metrics? And Python is still on 1.21.

tsloughter:
Looks like python moving to a file per https://github.com/open-telemetry/opentelemetry-python/pull/3586/files

bryannaegele:
Since starting this 1.26.0 has been released along with the new weaver generator. Part of 1.26.0 was the finalization of the `registry` where all the attributes actually live and other spec sections reference. This greatly simplifies generation.

An important change that needs to be addressed for Erlang is `allow_custom_values` is being [removed for enum attributes](https://github.com/open-telemetry/weaver/issues/218). Previously, in both languages, we did not generate functions or macros for enum attributes, opting instead to just create typespecs, appending `| atom()` to anything that allowed custom values. To get the typing/experience better in Elixir I took advantage of the use of functions to accept an argument and mapping back the enum value [http example](https://github.com/open-telemetry/opentelemetry-erlang/pull/733/files#diff-e7bc30338544c60cc0ff8469d27139974bfe4ec7da09b40b90a3ec1490cc528bR234-R250). We'll have to figure something out for Erlang.

I was able to get the reference guides added pretty easily now, as well. Linking into them from the modules isn't working but they're there, making it unnecessary to go to the website now.

One outstanding issue is on function naming. Right now the default formatting creates functions with the ctx id's name prefixed, e.g. `HttpAttributes.http_response_size()`. A `split` function is [making its way into weaver](https://github.com/open-telemetry/weaver/pull/207) which would make it pretty easy to turn it into `HttpAttributes.response_size()`. I don't know which is really preferable since leaving it the way it is does allow a user to import everything and not have collisions. If we generated a single file, this would be how it's named anyhow, but I still feel breaking the modules out is best for documentation. Given that, I lean towards going with how it is.

apps/opentelemetry_semantic_conventions/lib/container_attributes.ex
```diff
@@ -0,0 +1,214 @@
...
+      :user
+      
+      iex> OpenTelemetry.SemanticConventions.ContainerAttributes.container_cpu_state(:custom_value)
+      :custom_value
+  """
+  @spec container_cpu_state(container_cpu_state()) :: :user | :system | :kernel | atom()
+  def container_cpu_state(option) do
+    :"container.cpu.state"
+
+    case option do
```
yordis, line 78:
```suggestion
  def container_cpu_state(option) do
    case option do
```

Are these atoms floating around intentional?

yordis (commented):
I wish the package didn't use atoms since there is no way to do dead-code elimination and they are limited

apps/opentelemetry_semantic_conventions/lib/container_attributes.ex
bryannaegele: Template typo

bryannaegele:
> I wish the package didn't use atoms since there is no way to do dead-code elimination and they are limited

Atoms are the most efficient implementation and there aren't that many in this package. The default VM limit is 1m and that can be raised if a user wants. Binaries in systems like this would destroy your application in no time.

apps/opentelemetry_semantic_conventions/lib/cloud_attributes.ex
```diff
@@ -0,0 +1,276 @@
...
+      :gcp_cloud_run -> :gcp_cloud_run
+      :gcp_kubernetes_engine -> :gcp_kubernetes_engine
+      :gcp_cloud_functions -> :gcp_cloud_functions
+      :gcp_app_engine -> :gcp_app_engine
+      :gcp_openshift -> :gcp_openshift
+      :ibm_cloud_openshift -> :ibm_cloud_openshift
+      :tencent_cloud_cvm -> :tencent_cloud_cvm
+      :tencent_cloud_eks -> :tencent_cloud_eks
+      :tencent_cloud_scf -> :tencent_cloud_scf
+      _ -> option
```
tsloughter, line 179: Sorry if we've already discussed this but I don't see the point in these functions ad why we don't provide functions or macros like:

`cloud_platform.aws_ec2`

That is the only way the user will catch mistakes as far as I can tell. `cloud_platform(:ec2)` will be accepted just fine despite being wrong.

We could then do a `cloud_platform/1` that just accepts an atom and returns the same atom, no `case`.

apps/opentelemetry_semantic_conventions/lib/cloud_attributes.ex
```diff
@@ -0,0 +1,276 @@
...
+
+  ### Example
+      iex> OpenTelemetry.SemanticConventions.CloudAttributes.cloud_platform(:alibaba_cloud_ecs)
+      :alibaba_cloud_ecs
+      
+      iex> OpenTelemetry.SemanticConventions.CloudAttributes.cloud_platform(:custom_value)
+      :custom_value
+  """
+  @spec cloud_platform(cloud_platform()) ::
+          :alibaba_cloud_ecs
```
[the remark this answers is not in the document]
bryannaegele: The type name uses the same name as the function name https://github.com/open-telemetry/opentelemetry-erlang/pull/733/files/82b1319c8f76fa3330c242dd1ce0e08e8beed688#diff-17c4fe0a1cbc4aea65b3be524c04ca6fe255283a17a29b15fb915ad9a3a35a0dR73

apps/opentelemetry_semantic_conventions/lib/cloud_attributes.ex
bryannaegele: That would require another module per enum, no? `CloudPlatformAttributes.cloud_platform.aws_ec2` isn't a valid.

I hear what you're saying. Another possible option is making it a map, then another that takes and returns whatever value for the fallback. A user would then at least have the option of `CloudPlatformAttributes.cloud_platform().aws_ec2` and `CloudPlatformAttributes.cloud_platform(custom_value)` and we could make the enums closed types with a fallback option that technically makes it open-ended.

bryannaegele: Technically, with the change to always allow custom values, `cloud_platform(:ec2)` is valid now. It doesn't _have_ to match what's in the spec.

apps/opentelemetry_semantic_conventions/lib/cloud_attributes.ex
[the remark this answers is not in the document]
yordis: Oooooooh 🤦🏻

apps/opentelemetry_semantic_conventions/lib/cloud_attributes.ex
bryannaegele: Ok. I put together that approach and it feels pretty nice. I'm open to any ideas for how to handle this for Erlang.

image

tsloughter: Why is the type a map?

tsloughter: I guess to guard against the case of `custom` being a value in a semconv you could do `def _custom...` or some such.

tsloughter: And for Erlang I'd probably do:

```
-define('cloud_platform.aws_ec2', aws_ec2).
-define(cloud_platform(Custom), Custom).
```

But I guess that'd have to be in a `cloud_platform_attributes.hrl`... so lots of files? Could prepend `cloud_platform_attributes.` to each I suppose.

tsloughter: And maybe we should consider returning a tuple of the key and value?

tsloughter:
Curious about Python's putting stuff under "incubating" https://github.com/open-telemetry/opentelemetry-python/tree/main/opentelemetry-semantic-conventions/src/opentelemetry/semconv/_incubating

Maybe instead of a separate package which we don't want to do we could still separate things by a directory or module. Like in erlang you'd have `include/incubating/metrics/db.hrl`?

bryannaegele:
> Maybe instead of a separate package which we don't want to do we could still separate things by a directory or module. Like in erlang you'd have include/incubating/metrics/db.hrl?

So far as I can tell, incubating is a superset of stable + experimental and they're asking instrumentation libraries to not use the incubating package. It looks like they're taking advantage of their conventions to only have one package but marking that as internal/unstable whereas Java is doing two packages.

We could put it under another namespace as you suggested and Elixir could offer a set of macro conveniences to template imports in if someone should want. That would make migration from experimental easier for the user. I think that's fine.

apps/opentelemetry_semantic_conventions/lib/cloud_attributes.ex
bryannaegele: I was trying to avoid the nested modules because it adds more files and the docs would get to a weird state where it appears as another namespace in the docs when we're just trying to approximate enum support. In practice, each of these namespaces are `groups` so it would be nice to keep that consistent to `group == module`.

> `-define('cloud_platform.aws_ec2', aws_ec2).`

That's sounds good to me and would keep the `single module per group` approach the same between both languages.

> And maybe we should consider returning a tuple of the key and value?

I'd disagree on this as I don't know of any languages that return the key and the value for an enum. Is there a use case where you'd need the key again when you accessed it by that key name? The grpc enums in rpc have been the main example I've been working from.

bryannaegele:
Ok. I think this is ready for formal review. Changes are as follows:
* Incubating attributes have been separated (Elixir == namespace, Erlang == folder)
* Enum attributes accept custom values via a 1-arity function in Elixir and 1-arity macro function in Erlang
* Enum attributes for Elixir are accessed with dot syntax from the attribute's returned map
* Old modules/header files are put back and all marked deprecated for backwards compatibility (determine a removal version later)
* Elixir modules are moved to a shorter namespace `OpenTelemetry.SemConv` to match [recommendations](https://github.com/open-telemetry/semantic-conventions/blob/v1.26.0/supplementary-guidelines/semantic_conventions_code_generation.md#semantic-conventions-artifact-structure). This is backward compatible since everything has moved to the registry and the old modules are deprecated.
* Docs now use tabsets with Elixir and Erlang examples
* Docs have been improved with additional info (value types, examples, etc)
* Acronyms are formatted properly now
* Migration instructions added
* Usage updated

I will fix up the conflicts before merging. This could use verifying of the generated docs to make sure everything looks pretty good. They don't have to be perfect since we can publish them again later. The main task is validating we're happy with the actual code.

.gitignore
```diff
@@ -18,4 +18,5 @@ _build
 rebar3.crashdump
 *~
 doc
-!_index.md
\ No newline at end of file
+!_index.md
+guides
```
joaothallis, line 22: What is the reason to ignore guides?

bryannaegele:
Added metrics. Should be at parity with [python](https://github.com/lmolkova/opentelemetry-python/pull/2) now so good to go @tsloughter.

bryannaegele:
Added a convenience for the semconv opt in env var

bryannaegele:
[Python](https://github.com/emdneto/opentelemetry-python-contrib/blob/1a1bf8196a7daf931977570d77b207e6c50846f7/opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py#L114-L120) is using `default` for stability opt-in which aligns well with the migration verbage so I went ahead and matched them.

apps/opentelemetry_semantic_conventions/include/incubating/container_attributes.hrl
```diff
@@ -0,0 +1,84 @@
...
+
+%% The CPU state for this data point.
+
+-define('container_cpu_state.user', 'user').
+
+-define('container_cpu_state.system', 'system').
+
+-define('container_cpu_state.kernel', 'kernel').
+
+-define(container_cpu_state(Custom), Custom).
```
tsloughter, line 40: Where are these used?  don't see a `CONTAINER_CPU_STATE`?

and why aren't they capitalized as `CONTAINER_CPU_STATE_SYSTEM`?

bryannaegele: Ok. Fixed up the enums.

```elixir
OpenTelemetry.SemConv.Incubating.DBAttributes.db_cassandra_consistencylevel()
:"db.cassandra.consistency_level"

OpenTelemetry.SemConv.Incubating.DBAttributes.db_cassandra_consistencylevel_values().all
:all

%{OpenTelemetry.SemConv.Incubating.DBAttributes.db_cassandra_consistencylevel() => OpenTelemetry.SemConv.Incubating.DBAttributes.db_cassandra_consistencylevel_values().all}
%{:"db.cassandra.consistency_level" => :all}

OpenTelemetry.SemConv.Incubating.DBAttributes.db_cassandra_consistencylevel_values(:custom_value)
:custom_value
```

```erlang
?DB_CASSANDRA_CONSISTENCYLEVEL.
'db.cassandra.consistency_level'

?'DB_CASSANDRA_CONSISTENCYLEVEL_VALUES.all'.
all

#{?DB_CASSANDRA_CONSISTENCYLEVEL => ?'DB_CASSANDRA_CONSISTENCYLEVEL_VALUES.all'}.
#{'db.cassandra.consistency_level' => all}

?DB_CASSANDRA_CONSISTENCYLEVEL_VALUES(custom_value).
custom_value
```

bryannaegele: And upon further investigation, `container.cpu.state` has [just been deprecated](https://github.com/open-telemetry/semantic-conventions/pull/1026) so it will show up as such in 1.27. It's still valid as of 1.26.

apps/opentelemetry_semantic_conventions/include/incubating/host_attributes.hrl
```diff
@@ -0,0 +1,107 @@
...
+
+-define('HOST_ARCH_VALUES.ppc32', 'ppc32').
+
+-define('HOST_ARCH_VALUES.ppc64', 'ppc64').
+
+-define('HOST_ARCH_VALUES.s390x', 's390x').
+
+-define('HOST_ARCH_VALUES.x86', 'x86').
+
+-define(HOST_ARCH_VALUES(Custom), Custom).
```
tsloughter, line 37: Maybe we shouldn't bother with this? Similar to Java doesn't bother doing a `HostArchValues.custom(custom)`. The point was just to convey to the user the option of providing a custom value, right?

apps/opentelemetry_semantic_conventions/include/incubating/aws_attributes.hrl
```diff
@@ -0,0 +1,201 @@
...
+%%  
+-define(AWS_LOG_STREAM_NAMES, 'aws.log.stream.names').
+
+
+%% The AWS request ID as returned in the response headers `x-amz-request-id` or `x-amz-requestid`.
+-define(AWS_REQUESTID, 'aws.request_id').
+
+
+%% The S3 bucket name the request refers to. Corresponds to the `--bucket` parameter of the [S3 API](https://docs.aws.amazon.com/cli/latest/reference/s3api/index.html) operations.
+-define(AWS_S_3_BUCKET, 'aws.s3.bucket').
```
tsloughter, line 181: I noticed Java doesn't have this issue like I mentoined with K_8_S and we have here with `S_3`:

```
  public static final AttributeKey<String> AWS_S3_BUCKET = stringKey("aws.s3.bucket");
```

apps/opentelemetry_semantic_conventions/include/incubating/host_attributes.hrl
bryannaegele: Yes. I was going to suggest the same. The only benefit is it gave some type hints but that's not necessary.

tsloughter (approved):
We discussed the issues, giving pre-approval.

bryannaegele:
We decided to change Erlang enum value definitions to be all screaming snake case while retaining `_VALUES_` in the name. This keeps us aligned with Java's naming approach while not having to create additional files. This should also prevent any future naming collisions.

This should be the last change. We will leave this open for 1 week for comment and additional review before merging and publishing.

apps/opentelemetry_semantic_conventions/include/client_attributes.hrl
lmolkova, line 1: the folder structure seems to be

```
...attributes definition files...
metrics
  ...metrics definition files...
incubating
  ...attributes definition files...
  metrics
    ...metrics definition files...
```

No strong opinion, but it could be easier to navigate if it was

```
attributes
  ...attributes definition files...
metrics
  ...metrics definition files...
events
  ...event definition files...
incubating
  attributes
    ...attributes definition files...
  metrics
    ...metrics definition files...
  events
    ...event definition files...
```


that's what we do in python https://github.com/open-telemetry/opentelemetry-python/tree/main/opentelemetry-semantic-conventions

apps/opentelemetry_semantic_conventions/include/error_attributes.hrl
lmolkova, line 1: please consider generating stable attributes inside incubating part as well. The motivation:

I'm an app developer and use experimental attributes. E.g. `DB_OPERATION_NAME`. 
DB semconv go stable (which I don't have to know about), I update my semconv package and boom, there is a breaking change and compilation failure. 

To minimize this pain, Python and Java generate all attributes in incubating part, but deprecate those that are stable. 

E.g. 

```python
CLIENT_ADDRESS: Final = "client.address"
"""
Deprecated in favor of stable :py:const:`opentelemetry.semconv.attributes.client_attributes.CLIENT_ADDRESS`.
"""
```

https://github.com/open-telemetry/opentelemetry-python/blob/ed8d9507176c33e53f6779ee6647776f629fcd7a/opentelemetry-semantic-conventions/src/opentelemetry/semconv/_incubating/attributes/client_attributes.py#L17

```java
   * @deprecated deprecated in favor of stable {@link
   *     io.opentelemetry.semconv.ClientAttributes#CLIENT_ADDRESS} attribute.
   */
  @Deprecated public static final AttributeKey<String> CLIENT_ADDRESS = stringKey("client.address");
```

https://github.com/open-telemetry/semantic-conventions-java/blob/2cab896b2c95b03d0e359a34a117c235d6ebaa88/semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/ClientIncubatingAttributes.java#L33


It adds a bit of complexity to jinja (check out python and java templates and happy to help), so if you decide to do it, it can totally be done incrementally as a follow up

bryannaegele:
> I might have missed it, but is there a place where schema_url values are defined? Suggesting to keep a list of supported ones e.g. like this https://github.com/open-telemetry/semantic-conventions-java/blob/main/semconv/src/main/java/io/opentelemetry/semconv/SchemaUrls.java
> 
> ```java
>   public static final String V1_26_0 = "https://opentelemetry.io/schemas/1.26.0";
>   public static final String V1_25_0 = "https://opentelemetry.io/schemas/1.25.0";
> ```
> 
> So that instrumentation libs and user apps have a constant which does not update for all their telemetry when they update the semconv artifact version.

What are the schema urls supposed to denote in combination with the codegen? Is it just to have a list for a user to use elsewhere?

tsloughter:
> > I might have missed it, but is there a place where schema_url values are defined? Suggesting to keep a list of supported ones e.g. like this https://github.com/open-telemetry/semantic-conventions-java/blob/main/semconv/src/main/java/io/opentelemetry/semconv/SchemaUrls.java
> > ```java
> >   public static final String V1_26_0 = "https://opentelemetry.io/schemas/1.26.0";
> >   public static final String V1_25_0 = "https://opentelemetry.io/schemas/1.25.0";
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So that instrumentation libs and user apps have a constant which does not update for all their telemetry when they update the semconv artifact version.
> 
> What are the schema urls supposed to denote in combination with the codegen? Is it just to have a list for a user to use elsewhere?

It is for a user to use when creating Tracers/Meters/Loggers.

Or do you just mean including old schema urls instead of just the current?

bryannaegele:
Yeah, the including previous ones. I guess that's for backwards compatibility.

bryannaegele:
@tsloughter you have any thoughts on the other feedback?

apps/opentelemetry_semantic_conventions/include/client_attributes.hrl
tsloughter: Hm, yea, this seems right @bryannaegele. If we are going to use subdirs for metrics we should use them for attributes.

apps/opentelemetry_semantic_conventions/include/error_attributes.hrl
tsloughter: Interesting. What do you think @bryannaegele ?

tsloughter:
I had some with @lmolkova 's comments, aside from those I'm still good.

apps/opentelemetry_semantic_conventions/include/client_attributes.hrl
bryannaegele: I think so. That will make running a clean easier, as well.

apps/opentelemetry_semantic_conventions/include/error_attributes.hrl
bryannaegele: For Erlang the deprecating of stable in incubating I don't think makes sense or work because you're doing an include, so if you include stable and incubating it "just works". If we included both we'd have to rename the macros or else they'd collide.

For Elixir, if a user does an `import`, we get the same behavior as Erlang. I think we could get painless upgrades by just defdelegating anything stable to the stable module. That would also help with usability by removing the need to alias twice.

tsloughter: @bryannaegele that doesn't work because there may be no stable attributes yet for a source. So you can't be including both stable and incubating, only incubating. Then some move to stable and it breaks compilation.

For a user who controls the code including the attribute that is fine, but if its an instrumentation library they are depending on that means they have to fork and get the upstream to update. Then users on old semantic conventions can no longer use the library because it won't compile for them.

When to actually switch a instrumentation library and break people on older versions I don't know...

lmolkova: even when it comes to user applications, requiring users to changes their code is ok for preview library, not not great. Especially when it comes to the attribute that was just stabilized. 

I don't know how dependency version resolution looks in erlang, but in other languages user app may get a higher version of semconv library as transitive dependency and can fail in runtime if such conflict is discovered there.

lmolkova:
> It is for a user to use when creating Tracers/Meters/Loggers.

> Or do you just mean including old schema urls instead of just the current?

I mean including all starting with some.

E.g. my http instrumentation library follows semconv 1.25, my messaging one is slightly out-of-date follows 1.21, and one of my app tracers wants to use 1.26.

In semconv artifact, we can define constants for multiple versions (i.e. every time we update version, we add a constant for a corresponding version). This way,. Http, messaging and app can use 3 different versions and report consistent telemetry.

In the past we used to generate just one, like here 

https://github.com/open-telemetry/opentelemetry-erlang/blob/45102546d2574c6e7900fcfc7e0d99aba10029c1/apps/opentelemetry_semantic_conventions/lib/open_telemetry/semantic_conventions/logs.ex#L10

which means that all components would report just one version, even if they populate telemetry according to different versions.

apps/opentelemetry_semantic_conventions/include/error_attributes.hrl
bryannaegele: I tested this same approach with the erlang header files. I didn't know you can include a header file into another header file, so this approach "just works" :)

bryannaegele: Ok. I modified it to do an `include` in erlang and `defdelegate` in elixir to the stable modules. That should enable backwards/forward compatibility while avoiding conflicts and not defining anything twice. This will make usability more pleasant since a user doesn't have to alias both modules/header files when using. I found that to be quite annoying.

.github/workflows/erlang.yml
```diff
@@ -4,10 +4,12 @@ name: Erlang
 on:
   pull_request:
     branches:
-      - 'main'
+      - "main"
+    types: [opened, reopened, synchronize, labeled]
```
joaothallis, line 8: What label is going to be used?

joaothallis (approved):

apps/opentelemetry_semantic_conventions/templates/registry/elixir/weaver.yaml
```diff
@@ -0,0 +1,58 @@
...
+      })
+      | map(select( .id as $id | any( $excluded[]; . == $id) | not ))
+      | map(select(.attributes | length > 0))
+    application_mode: each
+  - pattern: semantic_metrics.ex.j2
+    filter: >
+      .groups
+      | map(select(.type == "metric"))
+      | map(select(.stability == $stability))
+      | map({ id: .id, group_id: .id | split(".") | .[1], brief, unit, stability, deprecated, instrument, metric_name, note})
```
lmolkova, line 29: there is a small bug in this snippet and in the new helper method - https://github.com/open-telemetry/weaver/issues/289, it's not a problem for 1.26.0, but there will be a tiny issue issue with 1.27.0 - workaround is shared in that issue, but we plan to fix it in weaver in the next release

apps/opentelemetry_semantic_conventions/templates/registry/elixir/semantic_metrics.ex.j2
```diff
@@ -0,0 +1,58 @@
...
+{%- endif -%}
+defmodule {{ module_namespace }}.{{ module_name }} do
+  # This is an auto-generated file
+  @moduledoc """
+  OpenTelemetry Semantic Conventions for {{ ctx.id | title | acronym }} metrics.
+  """
+{%- for metric in ctx.metrics | sort(attribute="metric_name") %}
+  {% if metric is deprecated %}
+  @deprecated """
+  {{ metric.deprecated | replace("\n", "\n  ") }}
```
lmolkova, line 21: jfyi/future: there is a new helper for this - comment formatter https://github.com/open-telemetry/weaver/blob/2e74ab36e6bd6599bd9231b83e9002e40f199712/crates/weaver_forge/README.md#comment-filter

apps/opentelemetry_semantic_conventions/templates/registry/elixir/semantic_metrics.ex.j2
```diff
@@ -0,0 +1,58 @@
...
+  <!-- tabs-open -->
+  ### Elixir
+
+      iex> {{ module_namespace }}.{{ module_name }}.{{ c.func_name(metric.metric_name) }}()
+      {{ c.to_atom(metric.metric_name) }}
+
+  ### Erlang
+
+  ```erlang
+  ?{{ metric.metric_name | snake_case | upper }}.
```
lmolkova, line 45: nit
```suggestion
  ?{{ metric.metric_name | screaming_snake_case }}.
```

apps/opentelemetry_semantic_conventions/templates/registry/elixir/semantic_attributes.ex.j2
```diff
@@ -0,0 +1,126 @@
...
+{%- for attribute in ctx.all_attributes | sort(attribute="name") %}
+  {%- if attribute is stable %}
+  defdelegate {{ c.func_name(attribute.name) }}(), to: {{ stable_module_namespace }}.{{ module_name }}
+  {% if attribute.type is mapping %}
+  defdelegate {{ c.func_name(attribute.name) }}_values(), to: {{ stable_module_namespace }}.{{ module_name }}
+  {% endif -%}
+  {% endif -%}
+{% endfor %}
+{% endif -%}
+{%- for attribute in ctx.attributes | sort(attribute="name") %}
```
lmolkova, line 29: nit: attributes are already sorted
```suggestion
{%- for attribute in ctx.attributes %}
```

apps/opentelemetry_semantic_conventions/templates/registry/elixir/semantic_metrics.ex.j2
```diff
@@ -0,0 +1,58 @@
...
+{%- set module_namespace = "OpenTelemetry.SemConv.Incubating.Metrics" -%}
+{%- else -%}
+{%- set module_namespace = "OpenTelemetry.SemConv.Metrics" -%}
+{%- endif -%}
+defmodule {{ module_namespace }}.{{ module_name }} do
+  # This is an auto-generated file
+  @moduledoc """
+  OpenTelemetry Semantic Conventions for {{ ctx.id | title | acronym }} metrics.
+  """
+{%- for metric in ctx.metrics | sort(attribute="metric_name") %}
```
lmolkova, line 18: nit: metrics are already sorted by name (with the latest weaver)
```suggestion
{%- for metric in ctx.metrics %}
```

lmolkova (approved):
Left some minor suggestions and improvements, but looks great - thanks a lot for doing this :)

apps/opentelemetry_semantic_conventions/include/error_attributes.hrl
bryannaegele: Resolving this as this pattern is working well.

bryannaegele:
> > It is for a user to use when creating Tracers/Meters/Loggers.
> 
> > Or do you just mean including old schema urls instead of just the current?
> 
> I mean including all starting with some.
> 
> E.g. my http instrumentation library follows semconv 1.25, my messaging one is slightly out-of-date follows 1.21, and one of my app tracers wants to use 1.26.
> 
> In semconv artifact, we can define constants for multiple versions (i.e. every time we update version, we add a constant for a corresponding version). This way,. Http, messaging and app can use 3 different versions and report consistent telemetry.
> 
> In the past we used to generate just one, like here
> 
> https://github.com/open-telemetry/opentelemetry-erlang/blob/45102546d2574c6e7900fcfc7e0d99aba10029c1/apps/opentelemetry_semantic_conventions/lib/open_telemetry/semantic_conventions/logs.ex#L10
> 
> which means that all components would report just one version, even if they populate telemetry according to different versions.

Is there a point at which we remove old ones? Is that determined by the WG or some other way?

apps/opentelemetry_semantic_conventions/templates/registry/elixir/semantic_attributes.ex.j2
bryannaegele: I just tried this again and it's still outputting descending instead of ascending without the additional sort. Not sure where it's happening but I've got to leave it in for now. Same for metrics

apps/opentelemetry_semantic_conventions/templates/registry/elixir/semantic_metrics.ex.j2
bryannaegele: Nice! Gonna leave this for a future todo

#771

bryannaegele:
Ok. We're all the way to 1.27.0 now and the http libs are ready to go for beta. There are some weaver updates that can still be done but it's not necessary to get this out.

@tsloughter take a look at this list of excludes before publishing. We can always add something in later if needed.

`["android", "artifact", "aspnetcore", "cicd", "cpu", "device", "disk", "dotnet", "go", "ios", "jvm", "kestrel", "linux", "nodejs", "signalr", "system", "test", "vcs", "veightjs", "v8js", "webengine"]`

https://github.com/open-telemetry/semantic-conventions/tree/v1.27.0/model

tsloughter:
@bryannaegele excludes are semconv we don't care about, right? I'll have to look closer but first thought is that "cicd" we might want.

tsloughter:
And why would we leave out ones like cpu, device, disk, linux, system? I get they are low level but still might want to report some of that from an Erlang application.

Similar for `test`? And `artifact`.

bryannaegele:
For CICD were you thinking of somebody tracing in tests?

tsloughter:
@bryannaegele right.

bryannaegele:
Ok. I limited the excludes to just language/OS/runtime type namespaces, i.e. ios, jvm, v8, dotnet, etc

Also added a step to remove any guides matching those namespaces, as well.

@tsloughter I think that should wrap it up if you want to do one last spot-check of some attrs, docs, and metrics

tsloughter (approved):

PR #737, opened 2024-05-02
fix support for attribute lists of strings to not flatten them

tsloughter:
No longer are attribute values that are lists of strings converted to a single string. This means Erlang charlists will be sent in OTLP as a list of integers and iolists as a list of strings. The user must use characters_to_binary to convert to a binary string if they wish to set the value to a string when they have a value of those types now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## Exporter 
+
+### Fixes
+
+- [BREAKING: Fixes support for attribute values that are lists when the elements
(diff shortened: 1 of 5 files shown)


ferd (approved):
I guess next release requires a major bump?

bryannaegele:
> I guess next release requires a major bump?

It's fixing a long-standing bug introduced [here](https://github.com/open-telemetry/opentelemetry-erlang/pull/468/files#diff-647e41ac745cfba551574cec0d8011c8fea5c0a9f12bc02e9d423af6f6f105cdR72) so I think a patch is technically ok

bryannaegele (approved):

bryannaegele:
Looks like this just needs a fix for the dialyzer error

ferd (approved):

PR #747, opened 2024-05-30
Rename otel_metrics to avoid file clash

f-f:
(for Buck2)


tsloughter:
We can rename this. Though they aren't modules. I was considering it anyway.

PR #772, opened 2024-09-14
Fix warning by replacing charlist with string

zhihuizhang17:
This PR fixes `warning: single-quoted strings represent charlists. Use ~c"" if you indeed want a charlist or use "" instead`


bryannaegele (approved):

zhihuizhang17:
Hi @bryannaegele , could you help make an new release in `hex.pm`?

PR #780, opened 2024-09-24
Hard code semconv 0.2 values

bryannaegele:
Temporarily hard code semconv values from 0.2 so that the api and sdk can get a point release rather than forcing updates to 1.27. I think it will be simpler to do this to keep backwards compatibility for now and not have to do a semconv 1.27 pass. We can delay a minor release for now with this approach.


tsloughter (approved):

bryannaegele:
For future reference, we were also advised by Liudmila to not depend on semconv in the API or SDK and just copy in or hardcode the keys to prevent any version mismatch issues. This isn't a strict requirement, just advised

Issue #784, opened 2024-10-04
undefined function opentelemetry_exporter:export/3 after upgrading opentelemetry


davidjulien:
Hello!

After upgrading `opentelemetry` from version 1.4.0 to 1.4.1  and `opentelemetry_api` from version 1.3.1 to 1.3.2, I have now this error during execution:

```
span exporter threw exception: exporter=:opentelemetry_exporter exception error: undefined function opentelemetry_exporter:export/3
  in function  otel_batch_processor:export/3 (/app/deps/opentelemetry/src/otel_batch_processor.erl, line 444)
  in call from otel_batch_processor:send_spans/3 (/app/deps/opentelemetry/src/otel_batch_processor.erl, line 430)
```

My versions of opentelemetry libs seems up-to-date:
opentelemetry           1.4.1
opentelemetry_api       1.3.2  
opentelemetry_cowboy    0.3.0
opentelemetry_ecto      1.2.0
opentelemetry_exporter  1.7.0
opentelemetry_phoenix   1.2.0

Do I miss something during the upgrade? Changelog doesn't seem to mention a special action before upgrading.

Thanks

David

davidjulien:
Upgrading `opentelemetry_api` only doesn't trigger the bug.

tsloughter:
@davidjulien sorry, yes, this is a bug in the latest packages. I'll publish a new exporter that should resolve it.

We need CI that doesn't test against the entire repo but instead based on the version constraints to catch stuff like this...

tsloughter:
Well that and this should have been minor release bumps and I think mix would have caught it :(

tsloughter:
Retiring those packages to make sure I get this right and making new releases as soon as I can this morning.

davidjulien:
Thank you @tsloughter!

davidjulien:
Tested this morning. No bug detected anymore. Thank you for your reactivity! Have a nice week.

PR #786, opened 2024-10-05
Minor bumps

tsloughter:
I got scared after https://github.com/open-telemetry/opentelemetry-erlang/issues/784 and made these minor version bumps... API probably doesn't need to be at all. SDK now that its fixed to work with older export versions also could probably be a patch. But.. feels safer.

Note I retired the previous patch versions from hex.

### DIFF
--- a/.github/workflows/erlang.yml
+++ b/.github/workflows/erlang.yml
@@ -20,7 +20,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        otp_version: ["26.1.2", "25.3.2.7", "24.3.4.14"]
+        otp_version: ["27.1", "26.1.2", "25.3.2.7", "24.3.4.14"]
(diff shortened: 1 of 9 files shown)


bryannaegele (approved):

shaharz:
We're seeing issues with the zipkin exporter in 1.5.0:
```
span exporter threw exception: exporter=:opentelemetry_zipkin exception error: undefined function opentelemetry_zipkin:export/3
  in function  otel_batch_processor:export/3 (/build/deps/opentelemetry/src/otel_batch_processor.erl, line 444)
  in call from otel_batch_processor:send_spans/3 (/build/deps/opentelemetry/src/otel_batch_processor.erl, line 430)
```

Looks like this change was reverted in a later commit, was that intentional? Would be great to have it back 😄

tsloughter:
Argh, thanks, apparently we don't have integration tests for zipkin and the batch processor :(

Issue #788, opened 2024-10-11
Error after updating to `OpenTelemetry_exporter` to 1.8.0


kevinkirkup:
Here is my configuration:

```
      otel_collector_hostname = System.get_env("OTEL_COLLECTOR_HOSTNAME") || "localhost"

      otel_collector_protocol =
        if System.get_env("OTEL_COLLECTOR_PROTOCOL") == "grpc" do
          :grpc
        else
          :http_protobuf
        end

      otel_collector_port =
        System.get_env("OTEL_COLLECTOR_PORT") |> Utilities.envvar_to_integer_or_default(4318)

      otel_collector_http =
        if System.get_env("OTEL_COLLECTOR_SSL") |> Utilities.enabled?() do
          :https
        else
          :http
        end

      config :opentelemetry,
        span_processor: :batch,
        traces_exporter: :otlp

      config :opentelemetry_exporter,
        otlp_protocol: otel_collector_protocol,
        otlp_traces_protocol: otel_collector_protocol,
        otlp_traces_compression: :gzip,
        otlp_traces_endpoint: {
          otel_collector_http,
          otel_collector_hostname,
          otel_collector_port,
          [verify: :verify_none]
        }
```

Here is the error being reported:

```
{"function":"init/1","line":61,"message":"OTLP exporter failed to initialize: exception error: bad key: exporter_otlp_traces_protocol
 in function otel_configuration:update_config_map/5 (/app/deps/opentelemetry/src/otel_configuration.erl, line 280)
 in call from lists:foldl_1/3 (lists.erl, line 2151)
 in call from otel_exporter_otlp:merge_with_environment/8 (/app/deps/opentelemetry_exporter/src/otel_exporter_otlp.erl, line 363)
 in call from otel_exporter_traces_otlp:init/1 (/app/deps/opentelemetry_exporter/src/otel_exporter_traces_otlp.erl, line 122)
 in call from otel_exporter:init/1 (/app/deps/opentelemetry/src/otel_exporter.erl, line 48)
 in call from otel_batch_processor:init_exporter/2 (/app/deps/opentelemetry/src/otel_batch_processor.erl, line 327)
 in call from otel_batch_processor:idle/3 (/app/deps/opentelemetry/src/otel_batch_processor.erl, line 212)
 in call from gen_statem:loop_state_callback/11 (gen_statem.erl, line 3735)
[...]

PR #789, opened 2024-10-21
fix: key for otlp traces protocol app configuration

tsloughter:
Fix for #788 

This was broken by a find-replace I did when breaking things up across modules. Looking at adding a regression test still.


tsloughter:
Added regression test. Verified it fails with the old code.

ferd (approved):

PR #793, opened 2024-10-25
Fixes #688: Change service.instance.id serialization to string

### LINKED ISSUE #688: service.instance.id is sent as an integer, not as a string in protobuf messages
I am consuming OpenTelemetry events from an Elixir Phoenix application that depends on the hex package :opentelemetry 1.3.1 and exported using the http_protobuf exporter.

The service.instance.id is being received unexpectedly not as a long UUID-like string, but as a large negative integer by my consumer. 

The OpenTelemetry documentation indicates that the (experimental) service.instance.id parameter should be a string, and the examples show some kind of UUID string. 

  https://opentelemetry.io/docs/specs/semconv/resource/#service

To consume this value from BEAM language opentelemetry, it would be useful if it were either a UUID or a long hex string.

I have instrumented the protobuf decoder in my collector, and it appears that service.instance.id is being sent by opentelemetry-erlang as a protobuf integer, rather than a protobuf string or binary. The resource attributes aren't definitively typed in the proto definitions, so the protobuf type is probably being identified from the type of the erlang variable.

Quickly reviewing the code in otel_resource_detector.erl shows this value being generated  from otel_id_generator:generate_trace_id/0 which in other contexts - providing trace.id in spans - delivers a protobuf binary which is easily, consistently and mostly automatically coerced to a long hex string by the consumer because of the typing in the proto definition for spans. 

As a guess (not being somebody comfortable with erlang), I'd suggest stringifying the values in otel_resource_detector.erl in the function add_service_instance/1 and hopefully the protobuf encoder in the exporter will pick the right type.

Annosha:
Fixes #688

This PR aims to update Serialization Logic by:

Converting `service.instance.id ` to a string before it is sent in protobuf messages. Any occurrence of `service.instance.id` in [\otel_resource_detector.erl](https://github.com/open-telemetry/opentelemetry-erlang/blob/main/apps/opentelemetry/src/otel_resource_detector.erl) being assigned an integer value was modified to convert it to a binary string format, ensuring compliance with the expected data type.

Note: As a new contributor, I encountered difficulties in running tests due to a lack of documentation regarding the testing tools in the OpenTelemetry Erlang repository. Consequently, I am relying on CI/CD tests for my initial pull request. I would greatly appreciate any guidance on setting up the testing environment.

For your reference, I am using Windows with the following Erlang version: **Erlang (SMP, ASYNC_THREADS) (BEAM) emulator version 15.1.2**. I am currently facing challenges in setting up **Rebar3**.

### DIFF
--- a/apps/opentelemetry/src/otel_resource_detector.erl
+++ b/apps/opentelemetry/src/otel_resource_detector.erl
@@ -271,7 +271,7 @@ add_service_instance(Resource) ->
                     case erlang:node() of
                         nonode@nohost ->
                             ServiceInstanceId = otel_id_generator:generate_trace_id(),
-                            ServiceInstanceResource = otel_resource:create([{'service.instance.id', unicode:characters_to_binary(ServiceInstanceId)}]),
+                            ServiceInstanceResource = otel_resource:create([{'service.instance.id', integer_to_binary(ServiceInstanceId)}]),
                             otel_resource:merge(ServiceInstanceResource, Resource);
                         ServiceInstance ->
                             ServiceInstance1 = erlang:atom_to_binary(ServiceInstance, utf8),
@@ -290,7 +290,7 @@ add_service_instance(Resource) ->
             end;
         ServiceInstance ->
             ServiceInstanceResource = otel_resource:create([{'service.instance.id',
-                                                            unicode:characters_to_binary(ServiceInstance)}]),
+                                                                        otel_utils:assert_to_binary(ServiceInstance)}]),
             otel_resource:merge(ServiceInstanceResource, Resource)
     end.
 
--- a/apps/opentelemetry/src/otel_resource_detector.erl
+++ b/apps/opentelemetry/src/otel_resource_detector.erl
@@ -271,7 +271,7 @@ add_service_instance(Resource) ->
                     case erlang:node() of
                         nonode@nohost ->
                             ServiceInstanceId = otel_id_generator:generate_trace_id(),
-                            ServiceInstanceResource = otel_resource:create([{'service.instance.id', ServiceInstanceId}]),
+                            ServiceInstanceResource = otel_resource:create([{'service.instance.id', unicode:characters_to_binary(ServiceInstanceId)}]),
                             otel_resource:merge(ServiceInstanceResource, Resource);
                         ServiceInstance ->
                             ServiceInstance1 = erlang:atom_to_binary(ServiceInstance, utf8),
@@ -290,7 +290,7 @@ add_service_instance(Resource) ->
             end;
         ServiceInstance ->
             ServiceInstanceResource = otel_resource:create([{'service.instance.id',
-                                                             otel_utils:assert_to_binary(ServiceInstance)}]),
+                                                            unicode:characters_to_binary(ServiceInstance)}]),
             otel_resource:merge(ServiceInstanceResource, Resource)
     end.
 


apps/opentelemetry/src/otel_resource_detector.erl
```diff
@@ -290,7 +290,7 @@ add_service_instance(Resource) ->
             end;
         ServiceInstance ->
             ServiceInstanceResource = otel_resource:create([{'service.instance.id',
-                                                             otel_utils:assert_to_binary(ServiceInstance)}]),
+                                                            unicode:characters_to_binary(ServiceInstance)}]),
```
tsloughter, line 293: This should remain `otel_utils:assert_to_binary`.

tsloughter:
Sorry about the lack of docs on testing. You mention not being able to setup rebar3, can you give some more details? Running tests should just require running `docker compose up` followed by `rebar3 ct`.

apps/opentelemetry/src/otel_resource_detector.erl
```diff
@@ -271,7 +271,7 @@ add_service_instance(Resource) ->
                     case erlang:node() of
                         nonode@nohost ->
                             ServiceInstanceId = otel_id_generator:generate_trace_id(),
-                            ServiceInstanceResource = otel_resource:create([{'service.instance.id', ServiceInstanceId}]),
+                            ServiceInstanceResource = otel_resource:create([{'service.instance.id', unicode:characters_to_binary(ServiceInstanceId)}]),
```
tsloughter, line 274: This is an integer not characters so it needs to be converted instead with `integer_to_binary`.

tsloughter:
Actually we have service id all wrong. I don't know when this changed but the spec for it is different from when we implemented this and so our implementation needs to change to match https://github.com/open-telemetry/semantic-conventions/blob/d9d1e21a5bbb37fe8facea323491d59342a5f810/docs/attributes-registry/service.md#service-instance-id

Would you like to continue the work? Otherwise I can hop on this right away as its a bug and major difference to the spec.

Thanks for bringing it to our attention!

Annosha:
@tsloughter I am not highly experienced in Erlang and am focusing on beginner-friendly issues for my contributions. Please feel free to proceed with the work. I would greatly appreciate any assistance in ensuring this PR is accurate and ready for merging.

Annosha:
@tsloughter I have set up Rebar3; a system restart was required to update the environment variables. I’ve implemented the fixes in the latest commit, but I'm encountering errors when running` rebar3 ct` on my machine. I’m attaching the error log—could you please review it and advise if the issues are related to the changes I made?
[Uploading error-log.txt…]()

tsloughter:
@Annosha hey, sorry, that link to uploaded error log doesn't work. Could you just push the changes so we can see if they pass in github CI?

Annosha:
@tsloughter sorry about the file, here is the error log. And I have pushed changes here in this PR tho. 
[error-log.txt](https://github.com/user-attachments/files/17607445/error-log.txt)

tsloughter:
@Annosha oh, sorry for not getting back to you! To run the full test suite you need to have a collector running. You can do that with docker compose in the project. If you run that do things pass?

I haven't gotten around to fixing up `service.id` to match the spec but am looking at it again now. I may accept this first to at least make it a string if it'll take me a while to implement.

Sorry again for dropping the ball, and thanks!

PR #796, opened 2024-11-04
Fix `OpenTelemetry.Tracer.end_span` return type spec

RudolfMan:
Hi there!

Based on the specs of `otel_span:end_span/0,1` this `Tracer.end_span/0,1` is supposed to always return span context (`:undefined` is not an option for the return)

See here:
https://github.com/open-telemetry/opentelemetry-erlang/blob/0a38bd4765c9f3efbf5e5b40e95783b9df97e2a7/apps/opentelemetry_api/src/otel_span.erl#L291-L306

### DIFF
--- a/apps/opentelemetry_api/src/otel_span.erl
+++ b/apps/opentelemetry_api/src/otel_span.erl
@@ -289,7 +289,7 @@ update_name(_, _) ->
 %% If `SpanCtx' is not recording, this function doesn't do anything.
 %% Returns the updated span context.
 -spec end_span(SpanCtx) -> SpanCtx when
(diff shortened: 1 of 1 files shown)


tsloughter:
I believe the `otel_tracer.erl` may be the incorrect one as `SpanCtx` could be `undefined`. Not positive yet, but that is my guess here.

tsloughter:
@RudolfMan any update? I think you actually need to update `otel_tracer` to have `undefined` as a valid return of `end_span`.

RudolfMan:
Just updated, @tsloughter!

PR #802, opened 2024-11-22
bump experimental api and sdk versions (0.5.2 and 0.5.1)

tsloughter:


CHANGELOG.md
```diff
@@ -7,6 +7,31 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## Experimental API 0.5.2 - 2024-11-22
+
+### Added
+
+- [Hard code semconv values](https://github.com/open-telemetry/opentelemetry-erlang/pull/780)
```
ferd, line 14: you bump it as 0.5.2 here but it is 0.6 in code

tsloughter: agh, thanks. Fixed, hopefully. I wanted SDK to be a minor bump since its actual changes. API just replaces the same semconv values with hardcoded values, no actual changes to what the user sees, so just did a patch.

apps/opentelemetry_experimental/src/opentelemetry_experimental.app.src
```diff
@@ -1,6 +1,6 @@
 {application, opentelemetry_experimental,
  [{description, "Implementation of unstable OpenTelemetry signals"},
-  {vsn, "0.5.1"},
+  {vsn, "0.6.0"},
```
ferd, line 3: this still being 0.6.0 seems like an oversight too?

tsloughter: na, experimental sdk should be 0.6.0. I believe it is everywhere except the title of this PR?

ferd (approved):

PR #804, opened 2024-11-27
Inspect Elixir exception module name in `record_exception`

GregMefford:
The main change in this PR is to change the way we report the `exception.type` attribute when `OpenTelemetry.Span.record_exception` is called. Currently, it is using `to_string`, which results in the struct name having an `Elixir.` prefix on it, which is technically correct, but not what an Elixir developer would expect, since Elixir hides that most of the time. By using `inspect` instead, we get that functionality here as well.

The bulk of the PR, though, is adding some testing infrastructure that allows me to be able to assert that we're doing the right thing there, since this change is in the `opentelemetry_api` package, which does not depend on the `opentelemetry` SDK, where all the span-processing happens, leaving us with only `otel_tracer_noop`, which just ignores the data we want to assert on. I tried to make it so that it'll be compatible with Erlang tests as well in the future, but I'm not really familiar with how to organize the code there. In Elixir, there's a convention to put this kind of module that's only used in tests in the `test/support` directory and configure the compiler paths (which I've also done in this PR even though I didn't end up using it because it doesn't "just work" for Erlang files) when we're running in test mode so that it can find them. That way, they're not included in the distributed Hex package because they're not in the main library code. I wasn't sure how to do the same thing for Erlang.

### DIFF
--- a/apps/opentelemetry_api/lib/open_telemetry/span.ex
+++ b/apps/opentelemetry_api/lib/open_telemetry/span.ex
@@ -135,10 +135,8 @@ defmodule OpenTelemetry.Span do
   def record_exception(span_ctx, exception, trace \\ nil, attributes \\ [])
 
   def record_exception(span_ctx, exception, trace, attributes) when is_exception?(exception) do
-    exception_type = to_string(exception.__struct__)
-
     exception_attributes = [
-      {:"exception.type", exception_type},
+      {:"exception.type", inspect(exception.__struct__)},
       {:"exception.message", Exception.message(exception)},
       {:"exception.stacktrace", Exception.format_stacktrace(trace)}
     ]
--- a/apps/opentelemetry_api/lib/open_telemetry/span.ex
+++ b/apps/opentelemetry_api/lib/open_telemetry/span.ex
@@ -121,7 +121,7 @@ defmodule OpenTelemetry.Span do
   defdelegate add_events(span_ctx, events), to: :otel_span
 
   @doc """
-  Record an exception as an event, following the semantics convetions for exceptions.
+  Record an exception as an event, following the semantic conventions for exceptions.
 
   If trace is not provided, the stacktrace is retrieved from `Process.info/2`
   """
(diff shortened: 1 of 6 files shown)


GregMefford:
Heh, I missed that there were already some higher-level tests at the top level of the git repo until I saw it failing in CI. 🤦
Oh well, now we _also_ have some infrastructure in place that we can use at the lower level. 😅

ferd:
Does this risk changing what you'd see for an Erlang-provided exception when using the Elixir macros compared to what happened prior to this fix?

GregMefford:
> Does this risk changing what you'd see for an Erlang-provided exception when using the Elixir macros compared to what happened prior to this fix?

It's technically a change regardless since it's intentionally formatting the exception differently, but I'd argue it's fixing a bug compared to what would be expected on the Elixir side. But the only change in this PR is if you call the Elixir version of the API (`OpenTelemetry.Span.record_exception`) rather than the Erlang one (`otel_span:record_exception`), and passed in an Elixir Exception struct (it's pattern-matching only the ones that have that shape and silently ignoring others). The Erlang version of the API [does something different in terms of formatting](https://github.com/open-telemetry/opentelemetry-erlang/blob/a56e1345bf6750d29abb90843ae9103ecf16c7ca/apps/opentelemetry_api/src/otel_span.erl#L222), and I'd argue that make we should figure out how to converge them, given that libraries could end up calling one or the other depending on how they integrate, and IMO it's really more about how you want to format exceptions / stacktraces, perhaps based on a configuration option, even.

ferd:
Yeah to be clear my concern here was whether dropping `Elixir` from the elixir exception would also change the non-Elixir exception handled at the same call-site but it comes from an Erlang lib.

GregMefford:
> I have an implementation in Erlang in the cowboy instrumentation.

Nice, I did see that my Cowboy error spans had a different error formatting in prod, but hadn't dug into how it's working yet. That'll be a good starting point for converging on a standard format.

PR #808, opened 2024-12-05
feat: setup code attributes in Elixir macros

btkostner:
This replaces #550. It adds code attributes to all Elixir macro calls. This information is gathered at compile time, and then merged with the start opt attributes at runtime. I created a new public function to format these attributes in case down the line we want to format them differently (like removing the "Elixir" prefix from namespace) and in case other libraries need them while not using the macros.

The end result is you get these attributes for the Elixir `Tracer` macros: `code.filepath`, `code.function`, `code.lineno`, `code.namespace`.

### DIFF
--- a/apps/opentelemetry_api/lib/open_telemetry/attributes.ex
+++ b/apps/opentelemetry_api/lib/open_telemetry/attributes.ex
@@ -1,0 +1,74 @@
+defmodule OpenTelemetry.Attributes do
+  @moduledoc """
+  This module contains utility functions for span attributes.
+
+  Elixir has built in variables like `__ENV__` and `__CALLER__` that can be used to generate
+  span attributes like `code.function`, `code.lineno`, and `code.namespace` either during runtime
+  or compile time. This module provides a function to generate these attributes from a `t:Macro.Env`
+  struct.
+
+  For more information, view the [OpenTelemetry Semantic Conventions](OSC).
+
+  [OSC]: https://opentelemetry.io/docs/specs/semconv/attributes-registry
+  """
+
+  @code_filepath :"code.filepath"
+  @code_function :"code.function"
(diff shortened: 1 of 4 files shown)


apps/opentelemetry_api/lib/open_telemetry/tracer.ex
```diff
@@ -221,4 +241,17 @@ defmodule OpenTelemetry.Tracer do
   def update_name(name) do
     :otel_span.update_name(:otel_tracer.current_span_ctx(), name)
   end
+
+  @doc false
+  @spec merge_start_opts(OpenTelemetry.Span.start_opts(), OpenTelemetry.attributes_map()) ::
+          OpenTelemetry.Span.start_opts()
+  def merge_start_opts(start_opts, builtin_attributes) do
```
btkostner, line 248: I left this public but undocumented. Not sure if we want to document and make it part of the public API

tsloughter:
Sorry, I missed that there was a new PR. I'll get this reviewed soon.

apps/opentelemetry_api/lib/open_telemetry/attributes.ex
```diff
@@ -0,0 +1,74 @@
+defmodule OpenTelemetry.Attributes do
```
tsloughter, line 1: There is an `otel_attributes` module that something named `OpenTelemetry.Attributes` should be the Elixir equivalent of. While that isn't needed I think this should be named different anyway. Maybe `CodeAttributes`?

btkostner:
Let me see if I can pull some examples from other ecosystems and see what they do 👀

tsloughter:
@btkostner thanks!

tsloughter:
Oh and last I checked, I think months ago, no one else implemented these attributes at all :)

Issue #810, opened 2024-12-20
rebar3_ewqalizer was removed and CI is broken.


ferd:
The changeset, as of https://github.com/WhatsApp/eqwalizer/commit/786f0dad8a9b7b2f361b71947ea8d05368b1c16c removes the rebar3 plugin.

It seemingly now requires [ELP](https://whatsapp.github.io/erlang-language-platform/docs/get-started/) to be installed and running and to use a distinct plugin to work.

Since we were pulling from `main` on the plugin, all CI for these libs are now broken.

Do we want to try and fix this or do we just rip it out to get going faster at this point?

bryannaegele:
I vote for fastest fix. It's been a "nice to have" but hasn't proven critical.

tsloughter:
it is critical. but the plugin isn't needed, I haven't used it in ages in other projects and just forgot to do the removal here.

tsloughter:
To be clear, I'll try to clean this up today. So PRs to come.

PR #813, opened 2024-12-29
remove unneeded eqwalizer_rebar3 plugin

tsloughter:
Oops forgot to send this fix for #810

### DIFF
--- a/.github/workflows/erlang.yml
+++ b/.github/workflows/erlang.yml
@@ -137,8 +137,8 @@ jobs:
             ${{ runner.os }}-eqWAlizer-${{ matrix.otp_version }}-5-
       - name: Download elp
         run: |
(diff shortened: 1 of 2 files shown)


bryannaegele:
@tsloughter do you know where meck is getting pulled in? Seems like some dep needs to have a version set.
image

tsloughter:
Hm, I'll check. But what did we decide on version support? is it current + 2 or current + 3? If its current + 2 then 24 wouldn't be supported anyway and can be dropped from CI.

tsloughter:
Its a top level dep that I for some reason did `{deps, [{meck, ">= 0.0.0"}]}`...

bryannaegele:
Good point.

bryannaegele:
https://github.com/open-telemetry/opentelemetry-erlang/pull/814

bryannaegele (approved):

PR #814, opened 2025-01-02
Update OTP/Elixir versions and drop OTP 23 & 24

bryannaegele:


tsloughter (approved):

tsloughter:
guess I also need to update elp

Will do.

PR #816, opened 2025-01-04
remove semconv from docs.sh

tsloughter:
We should maybe do this for all of them and move to using mix for docs for the api packages and rebar3_ex_doc for docs in the other packages?


bryannaegele:
Semconv should already be converted so sounds good.

ferd (approved):

PR #821, opened 2025-01-20
chore: add relevant links to spec in doc blocks

msutkowski:
Just a suggestion. The other language libraries tend to link directly to the spec to see what the underlying operations map to and I figured we could do the same.


tsloughter (approved):
Thanks, this is great!

Is there a reason you picked 1.8.0 to link to? Are we that behind on what spec is followed?

msutkowski:
> Thanks, this is great!
> 
> Is there a reason you picked 1.8.0 to link to? Are we that behind on what spec is followed?

I chose it to align specifically with what's mentioned in the README. I can go ahead and look at the latest version of the spec and see if there are any major discrepancies that we'd have to account for 🕵️

msutkowski:
> > Thanks, this is great!
> > Is there a reason you picked 1.8.0 to link to? Are we that behind on what spec is followed?
> 
> I chose it to align specifically with what's mentioned in the README. I can go ahead and look at the latest version of the spec and see if there are any major discrepancies that we'd have to account for 🕵️

@tsloughter I compared the links on 1.8 to whats on 1.41 and it seems like we can use the latest and update things. The link paths change in some cases (e.g. https://github.com/open-telemetry/opentelemetry-specification/blob/v1.8.0/specification/context/context.md -> https://github.com/open-telemetry/opentelemetry-specification/blob/v1.41.0/specification/context/README.md), but it wouldn't be a huge effort to swap everything. I'd ultimately want to spend quite a bit more time and make sure I didn't miss anything though 😅

tsloughter:
Aaah, I gotcha. It is up to you then. I'd be happy to have the latest links but its ok if you want to have it merged as it is.

PR #837, opened 2025-03-02
simplify metrics API and add metrics READMEs

tsloughter:
I wanted to avoid mistakes from the earlier API's that are now stable and just remove anything not absolutely necessary since adding is easy to do, removing is impossible once stable without a 2.0.

Sorry @bryannaegele and @ferd for the large PR. The gist of it is that I removed the API for passing an Instrument record. I realized that all it was doing was extracting the name and using that to look up streams to update! So why pass the Instrument! Ugh, but just realized this also has a change to how docs are generated for the API that I never pulled out.

Note that Instrument names are namespaced by the Meter so they aren't global and won't conflict if two different Application's use the same name.

There are 2 reasons an API for passing something besides the name might come in the future:

- It is decided to add to the SDK modules like `otel_sdk_counter`. Then we'd store that name of the module (since SDK's can be technically swapped out) in the Instrument record and calling the API's `otel_counter:add` would just call `otel_sdk_counter:add`. I briefly considered this as it would make some logic clearer and more easily extendable. An example of this is how you can't record a negative value for a Counter. This can not be restricted in the API and has to be in the SDK. Having `otel_sdk_counter:add` would allow for:

```
add(_Ctx, Name, Number, _Attributes) when Number < 0 ->
    ?LOG_INFO("Discarding negative value for instrument ~s of type counter", [Name]),
    false;
add(Ctx, Instrument, Number, Attributes) ->
    otel_meter_default:record(Ctx, Instrument, Number, Attributes).
```

While currently this is done in `otel_meter_server`:

```
maybe_init_aggregate(_, _Meter, Value, #instrument{kind=Kind} = Instrument, _Stream, _Attributes)
        when Value < 0, Kind == ?KIND_COUNTER orelse Kind == ?KIND_HISTOGRAM ->
    ?LOG_INFO("Discarding negative value for instrument ~s of type ~s", [Instrument#instrument.name, Kind]),
    ok;
```

Which isn't extendable with third party Instruments since it would require updating `otel_meter server`.

- If Bound Instruments are added to the spec there would be an API for that. But that is different since it is bound to a set of attributes as well as the Instrument.


tsloughter:
@open-telemetry/erlang-approvers ping. Just trying to get this in so I can do some more and push on getting GA verified.

apps/opentelemetry_api/src/otel_tracestate.erl
```diff
@@ -115,7 +115,7 @@ update(Key, Value, Tracestate=#tracestate{members=TracestateList}) ->
             Tracestate
     end.
 
--spec decode_header(string() | undefined) -> t().
+-spec decode_header(unicode:latin1_binary() | undefined) -> t().
```
ferd, line 118: the code below uses `string:lexemes/2`; that's a Unicode-aware function, and without converting to Unicode first, there's a chance that some unicode characters could get corrupted, maybe?

I guess it hasn't been a problem anywhere else so we're probably fine ignoring this.

apps/opentelemetry_api_experimental/README.md
```diff
@@ -1,9 +1,126 @@
...
+- observable updown counter
+- observable gauge
+
+#### Macros
+
+The header `otel_meter.hrl` contains macros for working with Instruments. This
+includes creation in the form `create_<instrument type>` and recording
+measurements like `counter_add` and `histogram_record`.
+
+Below is example creation found in the [`dice_roll_elli` example]():
```
ferd, line 37: TODO: link somewhere

apps/opentelemetry_api_experimental/README.md
```diff
@@ -1,9 +1,126 @@
...
+
+Measurements are taken on an instrument through recordings. Each type of
+instrument has functions and macros specific to its type for recording
+measurements. Counters can be added to so have the `counter_add` macro, UpDown
+counters can also be added to (but with the addition of accepting negative
+numbers) so have `updown_counter_add` and Histograms are passed recordings so
+have the macro `histogram_record`.
+
+
+An example from [`dice_roll_elli` example]() of recording an addition to a
```
ferd, line 55: TODO: link somewhere

apps/opentelemetry_api_experimental/README.md
```diff
@@ -1,9 +1,126 @@
...
+
+See the Experimental SDK's `README.md` for how to setup Views for aggregation
+and then the exporting of metrics.
+
+## Details
+
+There are a number of pieces that work together to make Metrics work in
+OpenTelemetry. There are defaults you can use for everything except the Reader
+and Exporter. Without configuring a Reader (which takes an Exporter) there will
+be no metrics exported from the node.
```
ferd, line 70: This paragraph is a bit ambiguous. It might be better to go for something like 

"Metrics require multiple parts to work. Most are configured with functional default values, except for a Reader, which must be passed an Exporter. Other components include [list of components]"

but there should ideally be links that guide to what the parts are and what roles they play. As a reader I possibly do not know what a Reader or an Exporter are here, though I'll know to keep an eye on them. The rest of the docs cover Providers, Meters, Measurements, Instrumetns. Are they related to the Reader and Exporter? Where do I go from here?

apps/opentelemetry_experimental/README.md
```diff
@@ -1,9 +1,144 @@
...
 
-Build
------
+# Metrics
 
-    $ rebar3 compile
+There are a number of pieces that work together to make Metrics work in
+OpenTelemetry. There are defaults you can use for everything except the Reader
+and Exporter. Without configuring a Reader (which takes an Exporter) there will
+be no metrics exported from the node.
```
ferd, line 11: Paragraph seems to be a dupe from the API readme, so same comment here

apps/opentelemetry_experimental/README.md
```diff
@@ -1,9 +1,144 @@
...
+an exact (like a sum) or a statistical metric. Each Instrument has a default
+aggregation that is used if none is specified through a View by the user or
+through configuration to the Reader to override the default. 
+
+Supported aggregators:
+
+- sum: `otel_aggregation_sum`
+- drop: `otel_aggregation_drop`
+- last value: `otel_aggregation_last_value`
+- explicit bucket histogram: `otel_aggregation_histogram_explicit`
```
ferd, line 25: What do they do? From the name:

- sum adds all values I put
- drop I'm unsure; my first reaction would be like a negative sum, but looking at the [overall docs](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#drop-aggregation) (which maybe we should link to) seems to be more of a no-op thing that ignores all values
- last value: gauge-like
- explicit bucket histogram: histogram for which I specify all bucket boundaries?

apps/opentelemetry_experimental/README.md
```diff
@@ -1,9 +1,144 @@
...
+Default aggregators for instruments:
+
+- counter: sum
+- updown counter: sum
+- histogram: explicit bucket histogram
+- observable counter: sum
+- observable updown counter: sum
+- observable gauge: last value
+
+## View
```
ferd, line 36: ah here are the definitions the other README didn't contain

apps/opentelemetry_experimental/src/otel_metrics_tables.erl
```diff
@@ -93,6 +93,7 @@ foreach_instrument(InstrumentsTab, Fun) ->
 -spec lookup_instrument(ets:table(), #meter{}, otel_instrument:name())
                        -> otel_instrument:t() | undefined.
 lookup_instrument(InstrumentsTab, Meter, Name) ->
+    ct:pal("TAB ~p ~p", [{Meter, Name}, ets:tab2list(InstrumentsTab)]),
```
ferd, line 96:
```suggestion
```

ferd (commented):
The refactoring seems to make sense to me, but I had a few comments, mostly minor stuff.

apps/opentelemetry_api_experimental/README.md
tsloughter: I feel like maybe this paragraph isn't even needed. I think my intention originally was to setup the read for the fact that they would learn about a lot of pieces but its possible to use metrics without even interacting with many of them directly. But to do that I should also just spell out that the one you do interact with is an Instrument since that is the one listed below.

apps/opentelemetry_experimental/src/otel_metrics_tables.erl
tsloughter: Done locally, will push up with the rest of the changes.

apps/opentelemetry_experimental/README.md
tsloughter: Yea... this documentation is particularly hard because of the API/SDK split. I should probably just give light coverage of aggregators in the API readme to be complete and they can come to this one for more.

apps/opentelemetry_experimental/README.md
tsloughter: REmoved the paragraph.

apps/opentelemetry_experimental/README.md
ferd: Either that or just pointing to the other doc as the place to find more about some topic.

ferd (approved):

PR #855, opened 2025-05-09
SemConv 1.33.0

bryannaegele:
This is surely a nightmare for some but this version marks DB semantics as stable, so best to get it over with for Ecto before releasing that. Code attributes are also stable now.

Releases should be much easier now thanks to all the work the weaver team did.

Please pay special attention to the new namespaces and acronyms for acceptability. The `AZ` one not being in Azure is kinda weird tbh but that's the spec. 🤷🏻 

Android namespace was excluded in erlang but not elixir. Not sure which way to go on that one so just added it back to erlang for now for completeness.


trask:
> The `AZ` one not being in Azure is kinda weird

are you referring to `az.namespace`? that was renamed to `azure.namespace` (and `az.namespace` was deprecated)

bryannaegele:
Yeah, that was a weird one. I'm guessing we should just add that to the ignored namespaces since it effectively never existed in our releases? Seems like the easiest course of action.

trask:
> it effectively never existed in our releases

I'm curious why it never existed? `az.namespace` was around for a while, or are you updating from a pretty old semconv release?

bryannaegele:
From 1.27.0, so almost a year since the last release. I'm on my phone but I think they're new files which would mean we didn't have it before

tsloughter (approved):

PR #874, opened 2025-07-23
Precompute hex encodings of SpanID and TraceID

kzemek:
`io_lib:format/2` has very significant overhead over `binary:encode_hex/2`. This has shown up prominently in our profiling of small spanned functions (think a happy-path `Cache.get()`).

## Benchmark results

```
# io_lib:format/2
Name                ips        average  deviation         median         99th %
test_span      159.15 K        6.28 μs   ±321.29%        4.40 μs       52.21 μs

# binary:to_hex/2
Name                ips        average  deviation         median         99th %
test_span      374.44 K        2.67 μs  ±1132.93%        1.70 μs       17.35 μs
```

## Benchee code
```elixir
require OpenTelemetry.Tracer

defmodule TestSpan do
  def test(_) do
    OpenTelemetry.Tracer.with_span "test" do
      :ok
    end
  end
end

Benchee.run(%{test_span: &TestSpan.test/1},
  time: 10,
  before_scenario: fn _ ->
    "scenario"
    |> OpenTelemetry.Tracer.start_span()
    |> OpenTelemetry.Tracer.set_current_span()
  end
)
```

## Flamegraph overview
image


kzemek:
Given how dominating `encode_hex` still is in `otel_ctx:attach/1` and `otel_ctx:detach/1`, it might make sense to go a step further and precompute the encodings into `#span_ctx`.

```
# precomputed
Name                ips        average  deviation         median         99th %
test_span      519.81 K        1.92 μs  ±1355.00%        1.34 μs       12.64 μs
```

image

tsloughter:
I'd be curious if this also saves you any time: https://github.com/open-telemetry/opentelemetry-erlang/pull/590

tsloughter:
Yes, I think we can add to span_ctx without a major version bump. Maybe theoretically we can't but I don't think its stopped us before :), and those should only be used through functions on a local node, not passed between nodes or accessed with `element(N, SpanCtx)`.

kzemek:
Added precomputing code. Here's a quick benchmark between >=OTP26, >=OTP24, <OTP24 code
```
# >=OTP26
Name                ips        average  deviation         median         99th %
test_span      496.58 K        2.01 μs  ±1867.46%        1.35 μs       13.15 μs

# >=OTP24
Name                ips        average  deviation         median         99th %
test_span      420.95 K        2.38 μs  ±1526.01%        1.63 μs       16.49 μs

# <OTP24
Name                ips        average  deviation         median         99th %
test_span      369.14 K        2.71 μs  ±1639.21%        1.93 μs       17.33 μs
```

bryannaegele:
> I'd be curious if this also saves you any time: #590

This is still really needed for logs. It's a nightmare dealing with it encoded.

kzemek:
Thanks for the run, will get the `:undefined` oversight fixed today

kzemek:
Done.

There's an ordering problem with `otel_span_ets` tests. Since `opentelemetry` depends on `opentelemetry_api ~> 1.4`, the tests pull the current version which doesn't define the new fields (yet).

```
src/otel_span_ets.erl:105:15: field hex_trace_id undefined in record span_ctx
src/otel_span_ets.erl:107:15: field hex_span_id undefined in record span_ctx
```

tsloughter:
@kzemek sorry about dropping the ball on this. Can you rebase so we can merge?

PR #894, opened 2025-09-07
[Feature]: Add is_remote flag in exporter for spans and span links

### LINKED ISSUE #893: [Feature]: Add is_remote flag in exporter for spans and span links
After updating protobuf schema, we can set the is_remote_parent flag when exporting spans and span links with OTLP.

Describe the solution you'd like:
Update OTLP exporter to fill the relevant bits in SpanFlags.

Considered Alternatives
None

Additional Context
Example:
Proto schema PR: https://github.com/open-telemetry/opentelemetry-proto/pull/484
Go implementation: https://github.com/open-telemetry/opentelemetry-go/pull/5194/files
Python implementation: https://github.com/open-telemetry/opentelemetry-python/pull/3881/files

nikhilmantri0902:
After updating protobuf schema, we can set the is_remote_parent flag when exporting spans and span links with OTLP.

Closes #893

Describe the solution you'd like:
Update OTLP exporter to fill the relevant bits in SpanFlags.

Considered Alternatives
None

Additional Context
Example:
Proto schema PR: https://github.com/open-telemetry/opentelemetry-proto/pull/484
Go implementation: https://github.com/open-telemetry/opentelemetry-go/pull/5194/files
Python implementation: https://github.com/open-telemetry/opentelemetry-python/pull/3881/files

### DIFF
--- a/apps/opentelemetry/include/otel_span.hrl
+++ b/apps/opentelemetry/include/otel_span.hrl
@@ -28,6 +28,9 @@
           %% 64 bit int parent span
           parent_span_id          :: opentelemetry:span_id() | undefined | '_',
 
+          %% true if the parent span context came from a remote process
+          parent_span_is_remote   :: boolean() | undefined | '_',
+
           %% name of the span
           name                    :: unicode:unicode_binary() | atom() | '_',
 
--- a/apps/opentelemetry_exporter/src/otel_otlp_traces.erl
+++ b/apps/opentelemetry_exporter/src/otel_otlp_traces.erl
@@ -22,7 +22,8 @@
 %% for testing
 -ifdef(TEST).
 -export([to_proto_by_instrumentation_scope/1,
-         to_proto/1]).
+         to_proto/1,
+         build_span_flags/2]).
 -endif.
 
 -include_lib("opentelemetry_api/include/opentelemetry.hrl").
@@ -67,6 +68,7 @@ to_proto(#span{trace_id=TraceId,
                span_id=SpanId,
                tracestate=TraceState,
                parent_span_id=MaybeParentSpanId,
+               parent_span_is_remote=ParentIsRemote,
                name=Name,
                kind=Kind,
                start_time=StartTime,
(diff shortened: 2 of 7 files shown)


apps/opentelemetry/include/otel_span.hrl
```diff
@@ -28,6 +28,9 @@
           %% 64 bit int parent span
           parent_span_id          :: opentelemetry:span_id() | undefined | '_',
 
+          %% true if the parent span context came from a remote process
+          parent_span_is_remote   :: boolean() | undefined | '_',
```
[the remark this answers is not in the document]
nikhilmantri0902: Thank you for the thoughtful question! However, I believe there's a conceptual distinction that needs clarification:
trace_flags vs span.flags are different concepts:
**trace_flags (8-bit)**: W3C Trace Context specification flags
Only contains sampling information (0x01 bit)
Used for propagation in trace headers
Should NOT include isRemote information
**span.flags (32-bit)**: OpenTelemetry protocol span flags
Bits 0-7: Trace flags (same as above)
Bits 8-9: isRemote information (0x100, 0x200)
Only used in OTLP export, not propagated
Our current approach is correct because:
We store parent_span_is_remote as a separate boolean field
We compute span.flags at export time by combining trace_flags + isRemote info
This follows the OpenTelemetry specification exactly
It matches the implementation in Go and Python SDKs
The "sync" concern is not applicable because:
trace_flags is for propagation (W3C headers)
span.flags is for export (OTLP protocol)
They serve different purposes and are computed at different times
This separation of concerns is intentional and follows the OpenTelemetry specification. The isRemote information should not be mixed into trace_flags as that would violate the W3C Trace Context specification.

apps/opentelemetry_exporter/src/otel_otlp_traces.erl
```diff
@@ -144,3 +148,21 @@ to_otlp_status(?OTEL_STATUS_OK) ->
...
+%% This follows the OTLP specification for span flags.
+%% @param ParentIsRemote boolean() | undefined - whether the parent span context is remote
+%% @param BaseFlags integer() - base flags (typically trace_flags)
+%% @return integer() - computed span flags
+-spec build_span_flags(boolean() | undefined, integer()) -> integer().
+build_span_flags(ParentIsRemote, BaseFlags) ->
+    %% SPAN_FLAGS_CONTEXT_HAS_IS_REMOTE_MASK = 0x100
+    %% SPAN_FLAGS_CONTEXT_IS_REMOTE_MASK = 0x200
+    HasIsRemoteMask = 16#100,
+    IsRemoteMask = 16#200,
```
tsloughter, line 162: I take it we have no good way of using the values from the generated proto files because they aren't exposed?

nikhilmantri0902: right

PR #923, opened 2025-10-17
simplify cliff.toml usage and bump versions

tsloughter:


tsloughter:
Argh, merging this way messed up cliff generation... great.

PR #939, opened 2025-11-30
chore: Add clo monitor exemption for artifact hub

thompson-tomo:
This defines a .clomonitor.yml file which is used by [clo monitor](https://clomonitor.io/projects/cncf/open-telemetry#opentelemetry-network).

This file contains an exemption for artifact hub as artifact's produced by this library are not supported by artifact hub. By including the exemption we can improve the score for this project and the organisation. File contents have been copied from other repos where the exemption has been applied.


thompson-tomo:
No worries @tsloughter commit message has been updated.

PR #954, opened 2026-02-10
fix: preserve span table across otel_span_ets crashes

QuinnWilton:
The span ETS table is created without {heir, Pid, Data}. When the
otel_span_ets process crashes, the table is destroyed and all in-flight
spans are silently lost. The process restarts and creates a fresh empty
table, but spans that were active between crash and restart are gone.
The try/catch in storage_insert/1 prevents cascading badarg errors but
masks the data loss.

Use the supervisor (otel_span_sup) as heir. On crash, ownership
transfers atomically to the supervisor. The restarted otel_span_ets
sees the table already exists via ets:info/2, skips ets:new, and
resumes operating on the preserved data. The table is public, so all
read/write operations continue to work regardless of which process
owns it.

### DIFF
--- a/apps/opentelemetry/src/otel_span_ets.erl
+++ b/apps/opentelemetry/src/otel_span_ets.erl
@@ -197,6 +197,7 @@ init(_Opts) ->
         undefined ->
             ets:new(?SPAN_TAB, [named_table, public,
                                 {write_concurrency, true},
+                                {heir, whereis(otel_span_sup), undefined},
                                 {keypos, #span.span_id}]);
         _ ->
             ok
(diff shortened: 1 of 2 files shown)


tsloughter:
Really sorry I dropped the ball on this. If you've moved on I can move this forward but I think it is just off to make the sup an heir and never take it back. In this case there is no reason to take it back, so it works fine, but it may as well be started in `start_link` and not bother with the heir in that case.

QuinnWilton:
You're totally okay! I agree with that assessment :) Not taking the heir back felt a little bit strange to me too, and just starting things in `start_link` sounds reasonable.

I'm a little bit swamped this week preparing for my Goatmire talk next week, so feel free to grab this if you'd like, but I'm more than happy to make those changes once I have some time again

Issue #769, opened 2024-09-12
Configuration file support and how to integrate with sys.config


tsloughter:
OpenTelemetry has a new configuration file specification, example: https://github.com/open-telemetry/opentelemetry-configuration/blob/main/examples/sdk-config.yaml that we need to support.

It is defined that if used by the user that environment variables (which can be substituted into the config) are ignored.

I think for easiest integration we have to also ignore sys.config but this may be contentious and confusing. Instead support a path to the file in the configuration but then if that is given we ignore the rest.

jamescarr:
I think being intentrional is the best approach. If I am providing a configuration file in other applications I kind of presume it is going to ignore any environment variables unless they are used in string interpolation.

PR #984, opened 2026-05-08
fix: processor-level config takes precedence over top-level settings

### LINKED ISSUE #976: Batch processor app env config: user-provided exporter overridden by BATCH_PROCESSOR_DEFAULTS merge
## Observed Behavior

When configuring the batch processor via application environment with a custom exporter (e.g. `otel_exporter_pid` for testing), the user-provided `exporter` value is silently replaced by the default `{opentelemetry_exporter, #{}}` after `otel_configuration:merge_processor_config/4` runs.

```erlang
application:set_env(opentelemetry, processors, [
    {otel_batch_processor, #{scheduled_delay_ms => 1, exporter => {otel_exporter_pid, self()}}}
]),
application:start(opentelemetry).
```

After startup, the batch processor state shows `opentelemetry_exporter` as the active exporter, not `otel_exporter_pid`. Spans are never delivered to the test process.

## Root Cause

In `otel_configuration.erl`, `merge_processor_config/4` for `otel_batch_processor`:

```erlang
merge_processor_config(otel_batch_processor, Opts, ConfigMap, AppEnv) ->
    BatchEnvMapping = [{bsp_scheduled_delay_ms, scheduled_delay_ms},
                       {bsp_exporting_timeout_ms, exporting_timeout_ms},
                       {bsp_max_queue_size, max_queue_size},
                       {traces_exporter, exporter}],
    maps:merge(?BATCH_PROCESSOR_DEFAULTS,
               merge_processor_config_(BatchEnvMapping, Opts, ConfigMap, AppEnv)).
```

The `{traces_exporter, exporter}` entry in `BatchEnvMapping` causes `merge_processor_config_` to check `ConfigMap` for a `traces_exporter` key. The default `ConfigMap` contains `traces_exporter => {opentelemetry_exporter, #{}}`. When `is_default(traces_exporter, AppEnv, Mappings)` returns `false` (because the default `ConfigMap` has it set), the fold overwrites the user's `exporter` key in `Opts` with `{opentelemetry_exporter, #{}}`.

The final `maps:merge(?BATCH_PROCESSOR_DEFAULTS, MergedOpts)` then has the OTLP exporter in both maps, so the user's `otel_exporter_pid` is lost.

## Expected Behavior

User-provided `exporter` in the processor options map should take precedence over the default `traces_exporter` from `ConfigMap`. If the user explicitly sets `exporter => {otel_exporter_pid, Pid}` in the processor config, it should not be overridden.

## Workaround

Call `otel_batch_processor:set_exporter/2` after application start:

```erlang
application:start(opentelemetry),
otel_batch_processor:set_exporter(otel_exporter_pid, self()).
```

This bypasses `otel_configuration` entirely and sets the exporter directly on the running processor.

Discovered while fixing our test suite: https://github.com/straw-hat-team/commanded/pull/90

## Environment

- Erlang/OTP: 28
- Elixir: 1.19.0
- opentelemetry: 1.7.0
- opentelemetry_exporter: 1.10.0

## Notes

This is similar to #850 (resource config inconsistency in the same `merge_processor_config` area) — both stem from processor-level config being unintentionally overridden by global defaults during the merge.

The upstream test suite (`otel_batch_processor_SUITE.erl`) does not hit this path because it calls `otel_batch_processor:start_link/1` directly, bypassing `otel_configuration`.

lazureykis:
## Summary

Fixes #976.

When both `traces_exporter` (or `bsp_*` settings) and processor-level config are set, the processor-level config was silently overridden by the top-level value. For example:

```erlang
application:set_env(opentelemetry, processors, [
    {otel_batch_processor, #{exporter => {my_exporter, #{}}}}
]),
application:set_env(opentelemetry, traces_exporter, {other_exporter, #{}}),
```

The batch processor would end up using `other_exporter` instead of `my_exporter`.

### Changes

- **`merge_processor_config_/4`**: Simplified to always respect processor-level config — if a key already exists in processor opts, top-level/env values don't override it.
- **`transform(span_processor, batch|simple)`**: Returns empty `#{}` instead of `BATCH_PROCESSOR_DEFAULTS`/`SIMPLE_PROCESSOR_DEFAULTS`. Defaults are already applied via `maps:merge` in `merge_processor_config/4`, so pre-populating was redundant and prevented the merge logic from distinguishing user-set values from defaults.
- Removed the now-unused `is_default/3` helper.

### New precedence order

**processor opts > top-level/env config > BATCH_PROCESSOR_DEFAULTS**

### Tests

Added `processor_exporter_not_overridden` (12 assertions) and `processor_config_precedence_with_os_env` (4 assertions) covering:

- Processor exporter not overridden by top-level `traces_exporter`
- Processor `none` exporter preserved
- `bsp_*`/`ssp_*` settings don't override processor-level equivalents
- `bsp_*`/`ssp_*` still apply when keys are absent from processor opts
- OS env (`OTEL_TRACES_EXPORTER`) doesn't override processor-level exporter
- Multiple processors get independent config resolution
- Reporter's exact scenario from #976

### DIFF
--- a/apps/opentelemetry/src/otel_configuration.erl
+++ b/apps/opentelemetry/src/otel_configuration.erl
@@ -199,31 +199,22 @@ merge_processor_config(otel_simple_processor, Opts, ConfigMap, AppEnv) ->
 merge_processor_config(_, Opts, _, _) ->
     Opts.
 
-merge_processor_config_(EnvMapping, Opts, ConfigMap, AppEnv) ->
-    Mappings = config_mappings(general_sdk),
-
+merge_processor_config_(EnvMapping, Opts, ConfigMap, _AppEnv) ->
     lists:foldl(fun({K, V}, Acc) ->
                         case maps:get(K, ConfigMap, undefined) of
                             undefined ->
                                 Acc;
                             Value ->
-                                %% use default only if the config isn't found in Opts
-                                %% but if the value in ConfigMap isn't the default use it
-                                IsDefault = is_default(K, AppEnv, Mappings),
-                                case (IsDefault andalso not maps:is_key(V, Opts)) orelse not IsDefault of
+                                %% apply top-level/env config value only if the
+                                %% key isn't already explicitly set in processor opts
+                                case maps:is_key(V, Opts) of
                                     true ->
-                                        Acc#{V => Value};
+                                        Acc;
                                     false ->
-                                        Acc
+                                        Acc#{V => Value}
                                 end
                         end
                 end, Opts, EnvMapping).
-
-%% return true if the user (through application or os environment) configured
-%% a certain setting
-is_default(Key, AppEnv, Mappings) ->
-    {OSVarName, Key, _TransformType} = lists:keyfind(Key, 2, Mappings),
-    not lists:keymember(Key, 1, AppEnv) andalso os:getenv(OSVarName) =:= false.
 
 %% sampler configuration is unique since it has the _ARG that is a sort of
 %% sub-configuration of the sampler config, and isn't a list.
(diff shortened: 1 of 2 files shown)


lazureykis:
@tsloughter

tsloughter:
Sorry for the delay and thanks for this. I'm trying to be certain about this. Another issue is sadly another way of configuring things needs to get in soon and will complicate this even more. But I'll likely merge this today because off the top of my head it seems right.

lazureykis:
@tsloughter Friendly ping — I think this is safe to merge independently of #769: declarative config ignores env/`sys.config` when a file is present, so it sits as a separate tier and doesn't touch the precedence this PR fixes. Happy to rebase if #769 lands first. Rebased on latest `main`.

tsloughter (approved):

PR #988, opened 2026-05-18
Use erl_error:format_exception for exception.stacktrace (closes #248)

### LINKED ISSUE #248: Use new erl_error format_exception for record_exception?
We likely want to use http://erlang.org/documentation/doc-12.0-rc1/lib/stdlib-3.15/doc/html/erl_error.html on OTP-24+ on `otel_span:record_exception`.

arseniycodes:
Closes #248.

## Problem

`otel_span:record_exception/5` and `record_exception/6` format the `Stacktrace` argument via `io_lib`'s term printer with `chars_limit, 50`:

```erlang
{ok, StacktraceString} = otel_utils:format_binary_string("~0tP", [Stacktrace, 10], [{chars_limit, 50}]),
```

50 characters might be too little to capture a meaningful stacktrace.

## Fix

Issue #248 — opened in 2021 by @tsloughter — already noted that `otel_span:record_exception` should use `erl_error:format_exception/3` on OTP-24+. The helper for that already exists in this repo at [`otel_utils:format_exception/3`](apps/opentelemetry_api/src/otel_utils.erl) and is unused.

- **`exception.stacktrace`**: switch from the truncated term-printer call to `otel_utils:unicode_to_binary(otel_utils:format_exception(Class, Term, Stacktrace))`. This produces the natural multi-line Erlang representation, e.g.:
  ```
  ** exception throw: my_error
       in function  opentelemetry_SUITE:record_exception_works/1
       ...
  ```
- **`exception.type`**: raise `chars_limit` from 50 → 1024. Happy to discuss the limit

CI matrix is OTP 26.2 / 27.2 / 28.1 (all ≥ 24), so `otel_utils:format_exception/3`'s `erl_error` branch is always live; no conditional code is needed.

## Tests

`record_exception_works` and `record_exception_with_message_works` in `apps/opentelemetry/test/opentelemetry_SUITE.erl` previously baked the `chars_limit, 50` formatting into their expected values. Updated both tests to compute the expected stacktrace via `otel_utils:format_exception`.

## Impact

- **For users**: `exception.stacktrace` can capture more stacktrace information.
- **For backends**: attribute size goes up. For most exception sites the new output is ~few hundred bytes to a few KB.
- **Backward compatibility**: the `exception.type` / `exception.stacktrace` / `exception.message` attribute schema is unchanged; only the string content shape changes.

## How I noticed

I work at an OTel vendor and a customer complained about truncated stacktraces.

### DIFF
--- a/apps/opentelemetry/test/opentelemetry_SUITE.erl
+++ b/apps/opentelemetry/test/opentelemetry_SUITE.erl
@@ -956,10 +956,16 @@ record_exception_works(Config) ->
             [Span] = assert_exported(Tid, SpanCtx),
             [Event] = otel_events:list(Span#span.events),
             ?assertEqual(exception, Event#event.name),
+            {ok, ExpectedStack} = otel_utils:unicode_to_binary(otel_utils:format_exception(Class, Term, Stacktrace)),
             ?assertEqual(#{'exception.type' => <<"throw:my_error">>,
-                           'exception.stacktrace' => list_to_binary(io_lib:format("~p", [Stacktrace], [{chars_limit, 50}])),
+                           'exception.stacktrace' => ExpectedStack,
                            <<"some-attribute">> => <<"value">>},
                          otel_attributes:map(Event#event.attributes)),
+            %% Regression guard: the stacktrace must not be the legacy
+            %% 50-char truncated form. It must contain at least one resolved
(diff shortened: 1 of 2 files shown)


tsloughter (approved):
Thanks!

PR #991, opened 2026-05-24
fix: preserve boolean OTLP attribute encoding

### LINKED ISSUE #990: OTLP exporter encodes boolean attributes as strings
## Summary

Boolean OTLP attributes are currently exported as `string_value` instead of `bool_value`.

The cause is that `otel_otlp_common:to_any_value/1` matches `is_atom(Value)` before `is_boolean(Value)`. In Erlang, `true` and `false` are atoms, so the boolean clause is unreachable for real boolean values.

Current code:

```erlang
to_any_value(Value) when is_atom(Value) ->
    #{value => {string_value, to_binary(Value)}};
...
to_any_value(Value) when is_boolean(Value) ->
    #{value => {bool_value, Value}};
```

## Repro

```erlang
otel_otlp_common:to_any_value(true).
%% => #{value => {string_value, <<"true">>}}

otel_otlp_common:to_any_value(false).
%% => #{value => {string_value, <<"false">>}}
```

Expected:

```erlang
#{value => {bool_value, true}}
#{value => {bool_value, false}}
```

## Impact

This affects any OTLP-exported boolean attribute that goes through `otel_otlp_common`, so traces, metrics, and logs can all be affected.

Backends and collectors will see `"true"` / `"false"` as strings instead of booleans.

## Why this looks longstanding

This does not appear to be a recent regression.

The same clause ordering already existed in the older monolithic exporter implementation before `otel_otlp_common` was split out:

- `0d69c0fbbc54431e3f23a982daf4575735ac285e` (`exporter: update trace protos to v0.11.0`)

It was then copied into `otel_otlp_common.erl` in:

- `2f004d22757b5f5ccd839c02916405556b54321a` (`exporter: breakout trace and metric specific code to own modules`)

So this looks like an old bug that survived the refactor, not something newly introduced by the helper split.

## Testing gap

There is already exporter test coverage that builds a span with a boolean attribute, but it does not assert the encoded `AnyValue` variant, so this slips through.

## Suggested fix

Move the boolean clause above the atom clause and add a focused regression test, e.g.:

```erlang
to_any_value(Value) when is_boolean(Value) ->
    #{value => {bool_value, Value}};
to_any_value(Value) when is_atom(Value) ->
    #{value => {string_value, to_binary(Value)}}.
```

And a test like:

```erlang
?assertEqual(#{value => {bool_value, true}},
             otel_otlp_common:to_any_value(true)),
?assertEqual(#{value => {bool_value, false}},
             otel_otlp_common:to_any_value(false)).
```

fallintoplace:
## Summary

- preserve boolean OTLP attributes as `bool_value` instead of converting them to strings
- add a focused regression test for `otel_otlp_common:to_any_value/1`

Fixes #990.

## Root cause

`otel_otlp_common:to_any_value/1` matched `is_atom(Value)` before `is_boolean(Value)`.
In Erlang, `true` and `false` are atoms, so boolean attributes always took the string branch and were exported as `string_value`.

## Validation

- added a common test case covering `true` and `false`
- did not run the Erlang test suite locally because this environment does not have `erl`, `rebar3`, `mix`, or `elixir` installed

### DIFF
--- a/apps/opentelemetry_exporter/src/otel_otlp_common.erl
+++ b/apps/opentelemetry_exporter/src/otel_otlp_common.erl
@@ -60,14 +60,14 @@ to_any_value(Value) when is_binary(Value) ->
         String ->
             #{value => {string_value, String}}
     end;
+to_any_value(Value) when is_boolean(Value) ->
+    #{value => {bool_value, Value}};
 to_any_value(Value) when is_atom(Value) ->
     #{value => {string_value, to_binary(Value)}};
 to_any_value(Value) when is_integer(Value) ->
     #{value => {int_value, Value}};
 to_any_value(Value) when is_float(Value) ->
     #{value => {double_value, Value}};
-to_any_value(Value) when is_boolean(Value) ->
-    #{value => {bool_value, Value}};
 to_any_value(Value) when is_map(Value) ->
     #{value => {kvlist_value, to_key_value_list(maps:to_list(Value))}};
 to_any_value(Value) when is_tuple(Value) ->
--- a/apps/opentelemetry_exporter/test/opentelemetry_exporter_SUITE.erl
+++ b/apps/opentelemetry_exporter/test/opentelemetry_exporter_SUITE.erl
@@ -14,7 +14,7 @@ all() ->
      {group, grpc}, {group, grpc_gzip}].
 
 groups() ->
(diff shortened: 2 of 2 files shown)


tsloughter (approved):

tsloughter:
Thanks!

PR #993, opened 2026-05-27
fix: cap W3C Baggage extract at 8192 bytes and 180 entries

### LINKED ISSUE #992: W3C Baggage extract has no size or entry-count cap
## Summary

`otel_propagator_baggage:extract/5` walks the inbound `baggage` HTTP header through `string:lexemes(String, [$,])` and `lists:foldl/3` without any byte or entry-count cap. Each entry is percent-decoded into a fresh binary and inserted into the accumulator map. The W3C Baggage specification recommends 8192 bytes and 180 entries; this propagator enforces neither.

This is a spec-compliance bug and a BEAM hardening item: an oversized `baggage` header (legitimately malformed or not) causes the calling process to spend an unbounded amount of scheduler time and per-process heap on the parse. The default propagator configuration `{text_map_propagators, [trace_context, baggage]}` registers `otel_propagator_baggage` on every instrumented HTTP entry point, so any Cowboy/Elli/Mochiweb service with OpenTelemetry trace propagation enabled inherits the same shape.

The maintainers reviewed this off-list as GHSA-64w2-whjg-q7q7 and declined to treat it as a security advisory, but asked for it to be opened as a regular issue/PR (https://github.com/open-telemetry/opentelemetry-erlang/security/advisories/GHSA-64w2-whjg-q7q7).

## Reference: other OpenTelemetry SDKs already enforce these caps

| SDK | File | Byte cap | Entry cap |
|---|---|---|---|
| Java SDK 1.62.0 | `W3CBaggagePropagator` | 8192 | 180 |
| Go | `propagation/baggage.go` | 8192 (`maxBytes`) | 64 (`maxMembers`) |
| .NET | `BaggagePropagator.cs` | 8192 (`MaxBaggageLength`) | 180 (`MaxBaggageItems`) |
| C++ | `baggage.h` `FromHeader` | 8192 (`kMaxSize`) | 180 (`kMaxKeyValuePairs`) |
| Erlang | `otel_propagator_baggage.erl` | none | none |

## Current code

[`apps/opentelemetry_api/src/otel_propagator_baggage.erl`](https://github.com/open-telemetry/opentelemetry-erlang/blob/8fa66fdb448f16bc3d22af93d6c7ed3ea4b647b6/apps/opentelemetry_api/src/otel_propagator_baggage.erl#L81-L94) lines 81-94:

```erlang
extract(Ctx, Carrier, _CarrierKeysFun, CarrierGet, _Options) ->
    case CarrierGet(?BAGGAGE_HEADER, Carrier) of
        undefined ->
            Ctx;
        String ->
            Pairs = string:lexemes(String, [$,]),
            DecodedBaggage =
                lists:foldl(fun(Pair, Acc) ->
                                    [Key, Value] = string:split(Pair, "="),
                                    Acc#{decode_key(Key) => decode_value(Value)}
                            end, #{}, Pairs),
            otel_baggage:set_to(Ctx, DecodedBaggage)
    end.
```

`string:lexemes(String, [$,])` allocates the full list of pair-binaries up front, `lists:foldl` then walks every pair and calls `decode_key` / `decode_value` (each builds a fresh binary), and the accumulator map keeps growing. None of `byte_size(String)`, `length(Pairs)`, or accumulator size is bounded. The `[Key, Value] = string:split(...)` match is also non-exhaustive: a single malformed pair (no `=`) crashes the extract.

## Observed behavior

Tested against `opentelemetry_api 1.4.0` installed from Hex.pm into a fresh rebar3 project on Erlang/OTP 27 erts-15.2.7.8:

```
N=1000     (9.8 KB header)      extract: 35.7 ms       mem 426 KB
N=50000    (660 KB header)      extract: 33.6 seconds  mem 22 MB
N=100000   (1.4 MB header)      no output before 120s timeout
```

A single header thus consumes tens of seconds of scheduler time and tens of megabytes of per-process heap; repeated requests scale that linearly. Wrapping the call in a separate process does not help because the work is CPU-bound inside `lists:foldl`.

## Suggested fix

Mirror the Java SDK 1.62.0 caps (`MAX_BAGGAGE_BYTES = 8192`, `MAX_BAGGAGE_ENTRIES = 180`):

- Reject headers larger than 8192 bytes before `string:lexemes/2` walks them.
- Stop decoding after 180 entries regardless of how many pairs `lexemes` produced.
- While there, replace the non-exhaustive `[Key, Value] = string:split(Pair, "=")` match with a `case` that skips malformed pairs instead of crashing the extract.

PR with that change and a `otel_propagator_baggage_SUITE` covering both caps is filed alongside this issue.

tonghuaroot:
Fixes #992.

`otel_propagator_baggage:extract/5` currently walks the inbound `baggage` HTTP header through `string:lexemes(String, [$,])` and `lists:foldl/3` without any byte or entry-count cap. The W3C Baggage specification recommends 8192 bytes and 180 entries, and the other OpenTelemetry SDKs (Java SDK 1.62.0 `W3CBaggagePropagator`, Go `propagation/baggage.go` `maxBytes`/`maxMembers`, .NET `BaggagePropagator.cs` `MaxBaggageLength`/`MaxBaggageItems`, C++ `baggage.h` `FromHeader` `kMaxSize`/`kMaxKeyValuePairs`) all enforce equivalent caps. This brings the Erlang implementation in line.

The maintainers reviewed this off-list as GHSA-64w2-whjg-q7q7 and asked for it to land as a regular issue/PR (https://github.com/open-telemetry/opentelemetry-erlang/security/advisories/GHSA-64w2-whjg-q7q7).

## Changes

- `apps/opentelemetry_api/src/otel_propagator_baggage.erl`
  - Add `MAX_BAGGAGE_BYTES = 8192` and `MAX_BAGGAGE_ENTRIES = 180` defines.
  - `extract/5` rejects headers larger than `MAX_BAGGAGE_BYTES` before `string:lexemes/2` walks them.
  - Decode loop is rewritten as `decode_pairs/3`, which stops after `MAX_BAGGAGE_ENTRIES` regardless of how many pairs `lexemes` produced.
  - `decode_pairs/3` matches `string:split(Pair, "=")` via `case` and skips malformed pairs instead of crashing on `[Key, Value] = ...` like the previous code did.
- `apps/opentelemetry_api/test/otel_propagator_baggage_SUITE.erl`
  - New CT suite with cases for simple extract, byte cap, entry cap, headers within both caps, malformed-pair skipping, and missing header.

## Verification

```
$ rebar3 ct --suite=apps/opentelemetry_api/test/otel_baggage_SUITE,apps/opentelemetry_api/test/otel_propagators_SUITE,apps/opentelemetry_api/test/otel_propagator_baggage_SUITE
%%% otel_baggage_SUITE: ..
%%% otel_propagators_SUITE: .....
%%% otel_propagator_baggage_SUITE: ......
All 13 tests passed.

$ rebar3 xref
===> Running cross reference analysis...
(clean)
```

Run on `erlang:27-alpine` (Erlang/OTP 27) with rebar3 3.24.0, against the worktree at this PR's head.

### DIFF
--- a/apps/opentelemetry_api/src/otel_propagator_baggage.erl
+++ b/apps/opentelemetry_api/src/otel_propagator_baggage.erl
@@ -50,6 +50,13 @@
 
 -define(BAGGAGE_HEADER, <<"baggage">>).
 
+%% Limits recommended by the W3C Baggage specification and applied by the
+%% other OpenTelemetry SDK implementations (Java SDK 1.62.0
+%% `W3CBaggagePropagator`, Go `propagation/baggage.go`, .NET
+%% `BaggagePropagator.cs`, C++ `baggage.h`).
+-define(MAX_BAGGAGE_BYTES, 8192).
+-define(MAX_BAGGAGE_ENTRIES, 180).
+
 %% @private
 fields(_) ->
     [?BAGGAGE_HEADER].
@@ -83,14 +90,33 @@ extract(Ctx, Carrier, _CarrierKeysFun, CarrierGet, _Options) ->
         undefined ->
             Ctx;
         String ->
-            Pairs = string:lexemes(String, [$,]),
-            DecodedBaggage =
-                lists:foldl(fun(Pair, Acc) ->
-                                    [Key, Value] = string:split(Pair, "="),
-                                    Acc#{decode_key(Key) => decode_value(Value)}
+            case header_size(String) of
+                Size when Size > ?MAX_BAGGAGE_BYTES ->
+                    Ctx;
+                _ ->
+                    Pairs = string:lexemes(String, [$,]),
+                    DecodedBaggage = decode_pairs(Pairs, #{}, 0),
+                    otel_baggage:set_to(Ctx, DecodedBaggage)
+            end
+    end.
 
-                            end, #{}, Pairs),
-            otel_baggage:set_to(Ctx, DecodedBaggage)
+header_size(String) when is_binary(String) ->
+    byte_size(String);
+header_size(String) when is_list(String) ->
+    iolist_size(String);
+header_size(_) ->
+    0.
+
+decode_pairs([], Acc, _N) ->
+    Acc;
+decode_pairs(_Pairs, Acc, N) when N >= ?MAX_BAGGAGE_ENTRIES ->
+    Acc;
+decode_pairs([Pair | Rest], Acc, N) ->
+    case string:split(Pair, "=") of
+        [Key, Value] ->
+            decode_pairs(Rest, Acc#{decode_key(Key) => decode_value(Value)}, N + 1);
+        _ ->
+            decode_pairs(Rest, Acc, N)
     end.
 
 %%
--- a/apps/opentelemetry_api/test/otel_propagator_baggage_SUITE.erl
+++ b/apps/opentelemetry_api/test/otel_propagator_baggage_SUITE.erl
@@ -1,0 +1,105 @@
+-module(otel_propagator_baggage_SUITE).
+
+-compile(export_all).
+
+-include_lib("stdlib/include/assert.hrl").
+-include_lib("common_test/include/ct.hrl").
+
+-include("opentelemetry.hrl").
+
+-define(BAGGAGE_HEADER, <<"baggage">>).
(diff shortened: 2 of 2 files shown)


tsloughter:
Thanks! But closing in favor of #1031

PR #994, opened 2026-05-27
Optional gRPC dep

peaceful-james:
The gRPC has module naming conflicts with `h2` which is used by latest `hackney`.

This makes it optional (opt-in) to avoid that conflict.


hassox:
Would love to see this land

tsloughter:
I have changes to make the module names not conflict. Is that sufficient or do you still want changes like this?

tsloughter:
If I remove the name conflict is it worth having these changes? I got all the grpcbox/chatterbox changes published but forgot to update otel, I can do that soon.

peaceful-james:
Totally your call. 👍🏻

PR #1001, opened 2026-06-11
fix: decode hex trace and span ids in OTLP log records

### LINKED ISSUE #1000: OTLP logs exporter puts hex-encoded trace/span ids into protobuf bytes fields, breaking log-trace correlation
# OTLP logs exporter sends hex-encoded trace_id/span_id instead of raw bytes, breaking trace↔log correlation

## Summary

`otel_otlp_logs:log_record/2` (in `apps/opentelemetry_experimental/src/otel_otlp_logs.erl`)
copies the `otel_trace_id` / `otel_span_id` logger metadata **verbatim** into the
protobuf `trace_id` / `span_id` fields of the OTLP `LogRecord`.

That metadata is produced by `otel_span:hex_span_ctx/1`
(`apps/opentelemetry_api/src/otel_span.erl` — its doc comment notes it exists
"because the main use of this function is logger metadata"), and is a 32-char /
16-char **lowercase hex binary**. The OTLP wire format expects the **raw
16 / 8 bytes**.

The result: every OTLP backend receives 32 ASCII bytes as the trace id (16 as the
span id). Backends that hex-encode `bytes` ids for display/storage then show
**64-character "double-hexed" trace ids**, which never match the real trace ids
exported by the traces pipeline — so logs can never be joined to their traces.

Notably, the **same block of code already hex-decodes `otel_trace_flags`**
(`erlang:binary_to_integer(TraceFlagsHex, 16)`) before putting it in the
LogRecord — the ids were simply missed:

```erlang
%% otel_otlp_logs.erl (main @ 1fcc1301000c8c75d8b298187a3e8d6717ec03f1)
LogRecord = case Metadata of
    #{otel_trace_id := TraceId,
      otel_span_id := SpanId,
      otel_trace_flags := TraceFlagsHex} ->
        TraceFlags = case TraceFlagsHex of
            <<_:0, _/binary>> when byte_size(TraceFlagsHex) == 2 ->
                erlang:binary_to_integer(TraceFlagsHex, 16);   % <-- flags ARE decoded
            _ -> 0
        end,
        #{trace_id => TraceId,    % <-- 32-byte hex binary, NOT decoded
          span_id => SpanId,      % <-- 16-byte hex binary, NOT decoded
          trace_flags => TraceFlags};
    ...
```

## Spec

The OTLP protobuf definition declares these fields as raw bytes
([opentelemetry-proto `logs/v1/logs.proto`](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/logs/v1/logs.proto)):

> ```proto
> // A unique identifier for a trace. All logs from the same trace share
> // the same `trace_id`. ... is considered invalid (empty) if all bytes are zeros.
> bytes trace_id = 9;
> // A unique identifier for a span within a trace ...
> bytes span_id = 10;
> ```

(16 bytes for `trace_id`, 8 bytes for `span_id`; only OTLP/JSON uses hex
encoding, OTLP/protobuf and OTLP/gRPC carry raw bytes.)

The repo's own traces exporter does the right thing —
`apps/opentelemetry_exporter/src/otel_otlp_traces.erl` encodes
`trace_id => <<TraceId:128>>` and `span_id => <<SpanId:64>>` (raw bytes).

## Affected versions

- `opentelemetry_experimental` 0.5.1 (released logs exporter used via the
  Elixir/Erlang OTLP log handler)
- `main` as of `1fcc1301000c8c75d8b298187a3e8d6717ec03f1` (file
  `apps/opentelemetry_experimental/src/otel_otlp_logs.erl`, ~lines 92–111)
- Any version since the logs exporter started reading
  `otel_trace_id`/`otel_span_id` from logger metadata.

## Reproduction sketch

1. Elixir (or Erlang) app with the OpenTelemetry SDK and the experimental OTLP
   log handler/exporter configured.
2. Inside an active span, attach the span context to logger metadata the
   documented way:

   ```elixir
   ctx = OpenTelemetry.Tracer.current_span_ctx()
   Logger.metadata(:otel_span.hex_span_ctx(ctx) |> Map.to_list())
   Logger.info("hello")
   ```

3. Export over OTLP (protobuf or gRPC) to any backend (e.g. the OTel Collector
   with a `debug` exporter, ClickHouse, Tempo/Loki, etc.).
4. Observe: the backend receives a 32-byte `trace_id` whose bytes are the ASCII
   characters `0-9a-f`. Backends hex-encode it again for display and show a
   64-char trace id, e.g.:

   - actual trace id: `abcdef0123456789abcdef0123456789`
   - what the backend stores for the log record:
     `6162636465663031323334353637383961626364656630313233343536373839`

   No query for the real trace id will ever match the log records.

## Real-world impact

ServiceRadar hit this in production: every log line emitted by their Elixir
applications was unjoinable to its trace — the OTLP consumer received 64-char
trace ids on the log records while the trace pipeline carried the correct
32-char (16-byte) ids. Trace↔log correlation was silently broken across the
entire fleet until the double-hex pattern was spotted in the stored data.

## Suggested fix

Hex-decode `otel_trace_id`/`otel_span_id` to raw bytes before placing them in
the LogRecord map (exactly as the adjacent `otel_trace_flags` handling already
does), tolerating the charlist form returned by older versions of
`otel_span:hex_span_ctx/1`, passing through already-raw ids, and omitting the
correlation fields entirely when an id cannot be decoded. PR to follow.

mfreeman451:
# fix: decode hex trace and span ids in OTLP log records

Fixes #<issue-number>.

## What / Why

The OTLP `LogRecord.trace_id` / `LogRecord.span_id` fields are raw bytes
(16 / 8) on the wire, but `otel_otlp_logs:log_record/2` copied the
`otel_trace_id` / `otel_span_id` logger metadata — the 32/16-char lowercase hex
binaries produced by `otel_span:hex_span_ctx/1` — verbatim into the proto map.
Every OTLP backend therefore received ASCII hex as the id bytes and stored
"double-hexed" 64-char trace ids, silently breaking trace↔log correlation for
all Erlang/Elixir applications using the experimental OTLP log exporter.

The same block already hex-decodes `otel_trace_flags`
(`erlang:binary_to_integer(TraceFlagsHex, 16)`); the ids were simply missed.
The traces exporter shows the expected representation:
`otel_otlp_traces.erl` encodes `trace_id => <<TraceId:128>>`,
`span_id => <<SpanId:64>>`.

## The fix

`log_record/2` now decodes the id metadata to raw bytes before building the
LogRecord, via a new private `decode_id/2`:

- lowercase **or** uppercase hex binary of exactly 2N chars → N raw bytes;
- hex charlist (what older versions of `otel_span:hex_span_ctx/1` returned —
  the existing code comment in this function already notes the
  charlist→binary change) → converted to binary, then decoded;
- already-raw N-byte binary → passed through unchanged;
- anything else / decode failure → `trace_id`, `span_id` **and** `trace_flags`
  are omitted from the LogRecord entirely, rather than exporting garbage ids
  (an absent id is "invalid/unset" per the proto docs; a wrong-length or
  ASCII id is corrupt data).

The hex decode is hand-rolled (a 5-line nibble comprehension) instead of using
`binary:decode_hex/1` because that BIF requires OTP 24 and only handles mixed
case from OTP 26.2, while this repo still documents OTP 23+ support
(README "Erlang/OTP 23+") — same approach as the version-conditional fallbacks
in `otel_utils:encode_hex/1`.

## Test coverage

New `apps/opentelemetry_experimental/test/otel_otlp_logs_SUITE.erl`
(9 cases, exercised through `otel_otlp_logs:to_proto/3`):

- lowercase hex binary ids → exact raw bytes (+ `trace_flags` 1 from `<<"01">>`)
- uppercase hex binary ids → exact raw bytes
- charlist ids → exact raw bytes
- already-raw binary ids → passed through
- ids present without `otel_trace_flags` → ids decoded, no `trace_flags` key
- non-hex ids of the correct length → all three fields omitted
- wrong-length (odd) ids → all three fields omitted
- valid trace id + invalid span id → both ids (and flags) omitted
- no span-ctx metadata → no id fields, rest of the LogRecord unchanged

`rebar3 ct --suite apps/opentelemetry_experimental/test/otel_otlp_logs_SUITE`:
all 9 pass. The pre-existing experimental metric suites and `rebar3 xref` are
unaffected.

## Backward compatibility

- Anyone who worked around this locally by putting **raw** 16/8-byte ids into
  `otel_trace_id`/`otel_span_id` metadata is unaffected: raw ids pass through
  unchanged.
- Consumers that adapted to the buggy double-hex ids on the backend side will
  see correctly correlated ids after upgrading (i.e. the bug stops being
  observable — that is the point of the fix).
- Records whose metadata cannot be decoded now omit the correlation fields
  instead of carrying corrupt ids; the rest of the LogRecord is exported as
  before.
- No public API changes; the new functions are private to `otel_otlp_logs`.

Fixes #1000

### DIFF
--- a/apps/opentelemetry_experimental/src/otel_otlp_logs.erl
+++ b/apps/opentelemetry_experimental/src/otel_otlp_logs.erl
@@ -90,22 +90,15 @@ log_record(#{level := Level,
     Flags = 0,
 
     %% Note: otel_trace_id and otel_span_id from hex_span_ctx are now binaries, not charlists.
+    %% They are hex encoded, but the OTLP trace_id and span_id fields are the
+    %% raw 16/8 bytes, so they must be decoded before being put in the
+    %% LogRecord. If either id can't be decoded then the correlation fields
+    %% are omitted entirely rather than exporting garbage ids.
     LogRecord = case Metadata of
-        #{otel_trace_id := TraceId,
-          otel_span_id := SpanId,
-          otel_trace_flags := TraceFlagsHex} ->
-            TraceFlags = case TraceFlagsHex of
-                <<_:0, _/binary>> when byte_size(TraceFlagsHex) == 2 ->
-                    erlang:binary_to_integer(TraceFlagsHex, 16);
-                _ -> 0
-            end,
-            #{trace_id => TraceId,
-              span_id => SpanId,
-              trace_flags => TraceFlags};
-        #{otel_trace_id := TraceId,
-          otel_span_id := SpanId} ->
-            #{trace_id => TraceId,
-              span_id => SpanId};
+        #{otel_trace_id := MetaTraceId,
+          otel_span_id := MetaSpanId} ->
+            correlation_fields(MetaTraceId, MetaSpanId,
+                               maps:get(otel_trace_flags, Metadata, undefined));
         _ ->
             #{}
     end,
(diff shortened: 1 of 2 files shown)


mfreeman451:
Addressed both review comments in the amended commit:

1. `decode_id/2` charlist branch now wraps `unicode:characters_to_binary/1` in try/catch — non-chardata metadata lists (atoms in lists, etc.) omit the correlation fields instead of raising. Regression test: `non_chardata_list_ids_dropped`.
2. The duplicated decode branching is extracted into `correlation_fields/3`, which decodes both ids once and adds `trace_flags` only when present. While restructuring, also hardened the pre-existing `binary_to_integer/2` call: present-but-non-hex flags (e.g. `<<"zz">>`) now fall back to 0 instead of raising. Regression test: `non_hex_trace_flags_default_to_zero`.

Suite: 11/11 passing.

One out-of-scope observation while testing: an *improper list* metadata value (e.g. `[97|98]`) crashes log export in `otel_otlp_common:to_array_value/1` during attribute encoding — that path predates this PR and affects any metadata key, so I left it alone here; happy to file it separately.

mfreeman451 (commented):
lgtm

tsloughter:
Hm, this is good, my only thought is then maybe we shouldn't be storing the id's this way in the metadata in the first place. But that change would break people who currently rely on outputting this in their logs to stdout since they'd have to change their formatters. And since I did the brilliant thing of putting the metadata update in the stable SDK I guess I can't do that change until 2.0....

So going to accept this :), thanks!

tsloughter (approved):

PR #1005, opened 2026-06-30
fix(experimental): keep otel_log_handler exporting after an idle interval

anagrius:
## Problem

`otel_log_handler` stops exporting logs after the first export interval that elapses with an empty batch.

The handler is a `gen_statem`. Only the `idle` state arms the `export_logs` timer (in its `enter` clause). On a tick it transitions `idle -> exporting` with `{next_event, internal, export}`. The `exporting` state handles the export like this:

```erlang
exporting(internal, export, Data=#data{exporter=Exporter, ... batch=Batch}) when map_size(Batch) =/= 0 ->
    _ = export(Exporter, Resource, Batch, Config),
    {next_state, idle, Data#data{batch=#{}}};
exporting(EventType, EventContent, Data) ->
    handle_event(EventType, EventContent, Data).
```

When the batch is **empty**, the guarded clause doesn't match and the event falls through to `handle_event/3`, which returns `keep_state_and_data`. The state machine therefore stays in `exporting` and never returns to `idle` — so the timer is never re-armed. From that point on, `handle_event(cast, {log, ...})` keeps appending to the batch, but the `internal, export` event that drains it is never generated again. **All subsequent logs are buffered forever and never exported.**

## Impact

Any service that isn't continuously logging faster than `scheduled_delay_ms` hits this almost immediately: the first idle interval wedges the handler and log export silently stops. We hit this in production (an Elixir/Phoenix service): logs flowed for the first interval, then stopped.

## Fix

Add an `exporting/3` clause for the empty-batch case that returns to `idle`, re-arming the timer:

```erlang
exporting(internal, export, Data=#data{batch=Batch}) when map_size(Batch) =:= 0 ->
    {next_state, idle, Data};
```

## Test

Adds `otel_log_handler_SUITE` with a regression test: it lets several export intervals elapse with no logs, then emits one log and asserts it is exported within a timeout. Without the fix the handler is wedged and the test times out.

## Validation

Verified against the released `opentelemetry_experimental` 0.5.1 in a downstream Elixir app: a probe test that logs *after* an idle period failed (log never exported) before the fix and passes after applying this exact clause.

## Relationship to #671

#671 rewrites the batch/handler machinery wholesale (via `otel_batch_olp`) and would also resolve this, but it has been open since 2023. This is a minimal, self-contained fix that is safe to land independently and ship in a patch release while #671 is pending.

### DIFF
--- a/apps/opentelemetry_experimental/src/otel_log_handler.erl
+++ b/apps/opentelemetry_experimental/src/otel_log_handler.erl
@@ -189,6 +189,12 @@ exporting(internal, export, Data=#data{exporter=Exporter,
                                        batch=Batch}) when map_size(Batch) =/= 0 ->
     _ = export(Exporter, Resource, Batch, Config),
     {next_state, idle, Data#data{batch=#{}}};
+exporting(internal, export, Data=#data{batch=Batch}) when map_size(Batch) =:= 0 ->
+    %% Nothing accumulated this interval. Return to `idle` so its `enter` clause
+    %% re-arms the export timer. Without this clause an empty tick falls through
+    %% to `handle_event/3` (`keep_state_and_data`) and the handler stays in
+    %% `exporting` forever, never exporting again after the first idle interval.
+    {next_state, idle, Data};
 exporting(EventType, EventContent, Data) ->
     handle_event(EventType, EventContent, Data).
 
--- a/apps/opentelemetry_experimental/test/otel_log_handler_SUITE.erl
+++ b/apps/opentelemetry_experimental/test/otel_log_handler_SUITE.erl
@@ -1,0 +1,72 @@
+-module(otel_log_handler_SUITE).
+
+-compile(export_all).
+
+-include_lib("common_test/include/ct.hrl").
+
+all() ->
+    [exports_logs_after_idle_interval].
+
+init_per_suite(Config) ->
+    {ok, _} = application:ensure_all_started(opentelemetry_experimental),
+    OldPrimary = logger:get_primary_config(),
+    ok = logger:update_primary_config(#{level => info}),
+    [{old_primary, OldPrimary} | Config].
+
+end_per_suite(Config) ->
+    ok = logger:set_primary_config(?config(old_primary, Config)),
+    _ = application:stop(opentelemetry_experimental),
+    ok.
+
+%% Regression test for the empty-batch wedge: an export interval that elapses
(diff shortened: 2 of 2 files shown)


ferd (approved):

anagrius:
Update: CI has since gone red, but for pre-existing reasons unrelated to this change (hex.pm certificate breakage and a W3C interop test suite update). Details in my comment below; the tracestate gap is tracked in #1010 with a fix in #1011.

anagrius:
Note on the failing checks: they are pre-existing CI breakage unrelated to this change. The same jobs fail for any PR against current `main` (last green `main` run was May 24, before both regressions below). The Erlang test jobs that exercise this fix (OTP 26/27/28, Dialyzer, EqWAlizer) all pass.

**1. Elixir jobs (`format`, `Dialyze`, `Test API/SDK`)** fail during `erlef/setup-beam`'s `mix local.rebar` step, before any project code runs:

```
{:tls_alert, {:unsupported_certificate, ... {key_usage_mismatch, ...}}}
Could not mix rebar from any hex.pm mirror
```

Since around June 12, `builds.hex.pm` serves a certificate whose key-usage extensions are rejected by the OTP `ssl` versions pinned in `elixir.yml` (25.3.2.16, 27.2), see hexpm/hex#1175. Possible fixes in `elixir.yml`: add a `hexpm-mirrors` fallback (`https://cdn.jsdelivr.net/hex`) to the `setup-beam` steps, or bump the pinned OTP versions to patched releases (27.3+). Happy to send a `ci:` PR for either variant if maintainers want.

**2. `Run W3C Trace Context Interop Tests`** fails on two tracestate tests (`test_tracestate_key_illegal_vendor_format`, `test_tracestate_key_length_limit`). The workflow checks out `w3c/trace-context` at HEAD, and w3c/trace-context#584 (merged June 29) rewrote those tests to match the Trace Context Level 2 key grammar, where `@` is an ordinary key character. The Erlang tracestate parser still implements the Level 1 `tenant@vendor` grammar, so it discards those headers. That is a spec-compliance gap exposed by the updated test suite, independent of this PR.

Filed as #1010, fix in #1011 (verified locally against the interop suite at HEAD: all 41 tests pass; its workflows are awaiting first-run approval from a maintainer).

Once those two land, a re-run here should be green.

anagrius:
Any chance getting this merged?

ferd:
I'd need a maintainer (@tsloughter) to do this. As a reviewer, I have no merge permissions.

PR #1016, opened 2026-07-12
chore: add AGENTS.md for AI coding agent guidance

bryannaegele:
## Summary
- Adds `AGENTS.md` documenting the multi-app rebar3/mix layout, per-app build/test/lint commands, PR title conventions, and other repo-specific context for AI coding agents working in this repo.
- Adds `CLAUDE.md` as a pointer to `AGENTS.md` for Claude Code.

## Test plan
- N/A (documentation only)

Assisted-by: Claude Sonnet 5

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,120 @@
+# AGENTS.md
+
+Guidance for AI coding agents working in this repository.
+
+## What this is
+
+OpenTelemetry SDK/API implementation for Erlang and Elixir. It's a multi-app
+rebar3 project (not a single OTP app) — each directory under `apps/` is an
+independent, separately-versioned, separately-released Hex/rebar package:
+
+- `opentelemetry` — the SDK
+- `opentelemetry_api` — the API, the main entry point for Elixir users
+- `opentelemetry_api_experimental` / `opentelemetry_experimental` — experimental
+  API/SDK surface (metrics, logs) not yet stable
+- `opentelemetry_exporter` — OTLP exporter
+- `opentelemetry_zipkin` — Zipkin exporter
+- `opentelemetry_semantic_conventions` — generated semantic convention constants
+
+`opentelemetry_api`, `opentelemetry_api_experimental`, and
+`opentelemetry_semantic_conventions` are dual-published to Hex for both
+rebar3 and mix, so each of those three also has its own `mix.exs`.
+
+Changes are usually scoped to one `apps/*` directory. Check which app(s) a
+change touches before assuming a repo-wide build/test is needed.
+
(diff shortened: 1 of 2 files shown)


yordis:
I am not sure if adding such files is a good idea; you will be affecting everyone's workflow. That file injects information into most hardness, and such a file depends on each individual's workflow and style.

Eventually, it may become stale or add information that the hardness doesn't need.

They are plenty of study proving that such practices aren't as useful and actually a net negative.

I would suggest moving the information to the README or documentation directory.

bryannaegele:
Yeah, we can definitely trim down on the workflow bits with the exception of the commit/PR formatting which is in the [community guidelines](https://github.com/open-telemetry/community/blob/main/policies/genai.md).

`What this is` section I think can be updates/fixes to the readme and the top-level notes in `Tests`, as well.

In the `Notes for agents` there's a mix of things that can be pushed to the respective apps. And I agree on adding a top-level docs for this repo but I don't want to scaffold all that out in this.

Part of the exercise of this is to surface all the things that need to be fixed up which would cover some of these topics which it is doing. I meant to mark this as a draft. Similar PR will go up in contrib.

lmk what you think but there is a base requirement of adding the file with per the community guidelines. What else (if anything) goes in it is up to us.

yordis:
`CONTRIBUTING.md` file and normal documentation of the project which is why I created https://github.com/open-telemetry/opentelemetry-erlang-contrib/tree/main/docs is far more effective, given the fact that we will no be quantifying the usage of such file, most people don't, and the ones that do, discover that are just not that effective.

Every model release and every hardness change make these static files ineffective. These are just proven situations.

AGENTS.md
```diff
@@ -0,0 +1,122 @@
...
+  Titles with other types (e.g. `docs`, `refactor`, `test`) fail CI.
+- Changelogs per app are generated via `git cliff` (`justfile`, `cliff.toml`)
+  — don't hand-edit `apps/*/CHANGELOG.md`.
+
+### Commit formatting
+
+We appreciate it if users disclose the use of AI tools when the significant part of a commit is
+taken from a tool. When making a commit this should be disclosed through an
+Assisted-by: commit message trailer and as a trailer in PR Summaries.
+
```
yordis, line 95: Is this a CNCF policy or is there a direct objective to request this?

I am in the opposite camp unless we will be giving credits to built-in hardness and tools as well, models aren't the final answer, and in some cases, the list is long enough, given the fact that, sometimes, we could spawn multiple harnesses to work on thing (included proprietary hardnesses and models).

bryannaegele: I linked it above. It's in the community guidelines

Issue #753, opened 2024-06-30
otel_tracer_noop:with_span/5 doesn't restore context properly


DenysGonchar:
## description:
[otel_tracer_noop:with_span/5](https://github.com/open-telemetry/opentelemetry-erlang/blob/fa43476/apps/opentelemetry_api/src/otel_tracer_noop.erl#L61) doesn't restore context properly

## here is a code snippet that demonstrates the problem:
```Elixir
require OpenTelemetry.Tracer, as: Tracer

ctx = Tracer.with_span :parent_span1 do
  OpenTelemetry.Ctx.get_current()
end |> IO.inspect(label: "parent_span1 context"); :ok

OpenTelemetry.set_default_tracer({:otel_tracer_noop, []})

Tracer.with_span :parent_span2 do
  OpenTelemetry.Ctx.get_current() |> IO.inspect(label: "parent_span2 context")
  Tracer.with_span ctx, :nested_span, %{} do
    OpenTelemetry.Ctx.get_current() |> IO.inspect(label: "nested_span context")
  end
  OpenTelemetry.Ctx.get_current() |> IO.inspect(label: "restored parent_span2 context")
end; :ok
```

## and the output is:
```
parent_span1 context: %{
  {:otel_tracer, :span_ctx} => {:span_ctx,
   226093672235243343240248130524572830431, 14684213315684633963, 1,
   {:tracestate, []}, true, false, true,
   {:otel_span_ets, #Function<2.55784961/1 in :otel_tracer_server.on_end/1>}}
}

parent_span2 context: %{
  {:otel_tracer, :span_ctx} => {:span_ctx, 0, 0, 0, {:tracestate, []}, false,
   false, false, :undefined}
}
nested_span context: %{
  {:otel_tracer, :span_ctx} => {:span_ctx,
   226093672235243343240248130524572830431, 14684213315684633963, 1,
   {:tracestate, []}, true, false, true,
   {:otel_span_ets, #Function<2.55784961/1 in :otel_tracer_server.on_end/1>}}
}
restored parent_span2 context: %{
  {:otel_tracer, :span_ctx} => {:span_ctx,
   226093672235243343240248130524572830431, 14684213315684633963, 1,
   {:tracestate, []}, true, false, true,
   {:otel_span_ets, #Function<2.55784961/1 in :otel_tracer_server.on_end/1>}}
}
```
## expected behaviour:
the context is properly restored after execution of `otel_trace:with_span/5`

[...]

PR #1019, opened 2026-07-19
restore context properly in noop tracer if span function crashes

DenysGonchar:
this PR resolves issue #753

If the span function fails, the span must be properly ended and the previous context must be restored. This behavior is identical to otel_tracer_default.


apps/opentelemetry_api/src/otel_tracer_noop.erl
```diff
@@ -56,11 +56,15 @@ start_span(Ctx, _, _SpanName, _) ->
...
-    otel_ctx:attach(Ctx1),
+    Token = otel_ctx:attach(Ctx1),
     try
         Fun(SpanCtx)
     after
-        otel_ctx:attach(Ctx)
+        %% passing SpanCtx directly ensures that this `end_span' ends the span started
+        %% in this function. If spans in `Fun()' were started and not finished properly
+        %% they will be abandoned and it be up to the `otel_span_sweeper' to eventually remove them.
+        _ = otel_span_ets:end_span(SpanCtx),
```
tsloughter, line 66: Since it is a noop it will not be in ETS. This should be removed.

tsloughter (changes requested):

PR #1031, opened 2026-08-18
fix(api): enforce W3C Baggage limits on the extract path

serhiy-bzhezytskyy:
The inject path builds the baggage header, but `extract` parsed the inbound header with no limits — every comma-separated pair was folded into the map, so an unbounded inbound `baggage` header was decoded in full. The tracestate propagator already enforces its 32-member cap, so this was the one uncapped extract path.

`extract` now applies the W3C Baggage limits (180 entries / 4096 bytes per entry / 8192 bytes total): over-limit members are dropped once a limit is reached and the earlier ones are kept. Every other OpenTelemetry SDK enforces these on extract — opentelemetry-java did it as CVE-2026-45292 (fixed 1.62.0), and its Jaeger propagator borrows the same W3C limits.

**Update:** the byte accounting summed each pair's own bytes against the 8192-byte total but never charged the `,` that joins entries on the wire, so a header could pass the check while its real wire size was already over the limit. Fixed to charge one separator byte per already-accepted entry, mirroring how `inject` already builds the header.

Adds `otel_propagator_baggage_SUITE` with a case per limit (including a byte-vs-grapheme case and the separator-byte boundary); each fails without the change.

Assisted-By: Claude Fable 5

### DIFF
--- a/apps/opentelemetry_api/src/otel_propagator_baggage.erl
+++ b/apps/opentelemetry_api/src/otel_propagator_baggage.erl
@@ -50,6 +50,11 @@
 
 -define(BAGGAGE_HEADER, <<"baggage">>).
 
+%% W3C Baggage size limits, enforced on extract (https://www.w3.org/TR/baggage/#limits).
+-define(MAX_ENTRIES, 180).
+-define(MAX_ENTRY_BYTES, 4096).
+-define(MAX_TOTAL_BYTES, 8192).
+
 %% @private
 fields(_) ->
     [?BAGGAGE_HEADER].
@@ -84,13 +89,22 @@ extract(Ctx, Carrier, _CarrierKeysFun, CarrierGet, _Options) ->
             Ctx;
         String ->
             Pairs = string:lexemes(String, [$,]),
-            DecodedBaggage =
-                lists:foldl(fun(Pair, Acc) ->
-                                    [Key, Value] = string:split(Pair, "="),
-                                    Acc#{decode_key(Key) => decode_value(Value)}
-
-                            end, #{}, Pairs),
+            {DecodedBaggage, _, _} =
+                lists:foldl(fun decode_within_limits/2, {#{}, 0, 0}, Pairs),
             otel_baggage:set_to(Ctx, DecodedBaggage)
+    end.
+
+decode_within_limits(_Pair, {Acc, Count, Bytes}) when Count >= ?MAX_ENTRIES ->
+    {Acc, Count, Bytes};
+decode_within_limits(Pair, {Acc, Count, Bytes}) ->
+    PairBytes = byte_size(Pair),
+    NewBytes = Bytes + PairBytes,
+    case PairBytes =< ?MAX_ENTRY_BYTES andalso NewBytes =< ?MAX_TOTAL_BYTES of
+        false ->
+            {Acc, Count, Bytes};
+        true ->
+            [Key, Value] = string:split(Pair, "="),
+            {Acc#{decode_key(Key) => decode_value(Value)}, Count + 1, NewBytes}
     end.
 
 %%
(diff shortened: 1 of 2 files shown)


ferd (commented):
From the link in comments:

> If either of the above conditions is not met, a platform MAY drop list-members until both conditions are met. The selection of which list-members to drop and their order is unspecified and left to the implementer. **Note that the above limits are minimum requirements to comply with the specification.** An implementor or platform MAY define higher limits and SHOULD propagate as much baggage information as is reasonable within their requirements. If a platform cannot propagate all baggage, it MUST NOT propagate any partial list-members. If there are multiple baggage headers, all limits apply to the combination of all baggage headers and not each header individually.

Emphasis mine. These appear to be _minimum_ limits, and we're treating them as a maximum. I think this is okay and the most restrictive interpretation possible, but the value is not configurable and lower than the previous [absence of] values.

I think the implementation is fine, but I'm curious whether the choice of clamping to the limits represents a) a breaking change and b) a need to make the values configurable with these entries as a lower bound.

serhiy-bzhezytskyy:
Right, the spec sets minimums and this clamps to them.

Configurable: no SDK I checked exposes these (Go, .NET, JS, Python, C++, PHP and Ruby all hardcode the equivalents) and the spec defines no `OTEL_BAGGAGE_*` variable. The in-repo precedent is stricter than this PR: `otel_tracestate:parse_pairs/1` returns `[]` for a list past 32, dropping the header whole.

Breaking: the API is unchanged and only headers already past the minimums behave differently. Making the values configurable with these as a floor is a bigger change, better done separately.

Not addressed here: `inject/4` still folds the whole baggage map into the header with no cap.

serhiy-bzhezytskyy:
Pushed a fix for a separate accounting bug found while re-reading this: decode_within_limits/2 summed each pair's bytes but never charged the comma between entries, so the 8192-byte check could pass while the real wire size was already over. Independent of the minimums-vs-configurability question above, which is still open.

ferd:
> Right, the spec sets minimums and this clamps to them.
> 
> Configurable: no SDK I checked exposes these (Go, .NET, JS, Python, C++, PHP and Ruby all hardcode the equivalents) and the spec defines no `OTEL_BAGGAGE_*` variable. The in-repo precedent is stricter than this PR: `otel_tracestate:parse_pairs/1` returns `[]` for a list past 32, dropping the header whole.
> 
> Breaking: the API is unchanged and only headers already past the minimums behave differently. Making the values configurable with these as a floor is a bigger change, better done separately.
> 
> Not addressed here: `inject/4` still folds the whole baggage map into the header with no cap.

Since we have 
```
-define(MAX_ENTRIES, 180).
-define(MAX_ENTRY_BYTES, 4096).
-define(MAX_TOTAL_BYTES, 8192).
```
I think that if we only enforced the 32 limits, we're now able to be limited to 2 entries if they are hitting the MAX_TOTAL_BYTES between the two of them. However I can see how this is more of a "bug fix" than a change of behavior that would be breaking, am willing to approve.

ferd (approved):

serhiy-bzhezytskyy:
Merged main into the branch (no conflicts, pushed). One thing worth flagging: Codecov reports 0% patch coverage on `otel_propagator_baggage.erl`, but that's a CI-side artifact -- ran `cover:analyse(otel_propagator_baggage, calls, line)` against the local coverdata and every line of `decode_within_limits/2` is hit hundreds of times by the existing 5 tests. Is there anything else you'd want changed here, or does this need a maintainer rather than another round from you?

serhiy-bzhezytskyy:
@tsloughter this has ferd's approval and his spec question answered, branch is up to date with main, and the only red check (Codecov patch coverage) is a CI-side artifact -- verified locally that every line of the new function is covered. Could you take a look when you have a moment?

tsloughter:
@serhiy-bzhezytskyy thanks! Once CI passes I'll merge.

PR #1032, opened 2026-09-11
feat: support declarative configuration through json file

tsloughter:
Only a json file, which must already have been validated against
the jsonschema, is accepted. It should be written in yaml and
converted to json. The opentelemetry-configuration repo provides
a tool for doing this conversion and verification.

If OTEL_CONFIG_FILE is set then all other configuration is
ignored, including the application environment and OS env.


tsloughter:
Opening as a draft because this is mostly codex written. Once I've reviewed each line and made any necessary changes I will take it out of draft. But the overall idea of accepting a json file with `OTEL_CONFIG_FILE` and ignoring all other configuration will stay.

I will close my other PR about declarative configuration. I originally started this with the plan of supporting a matching configuration format in `sys.config` and internally in each component. That was too much to do at once. Breaking it down into support the declarative config by converting it to the current format and then worrying about improving the existing `sys.config` based configuration is much simpler -- though I still have no idea how to do that without breaking backwards compatibility without being very hacky.

tsloughter:
I should also make clear this is NOT full support. It supports what we already supported in `sys.config` essentially, but in yaml/json format as defined by the configuration schema. Things like config providers and custom definitions of components is not yet supported and likely will wait for a later PR.

tsloughter:
I'm scrapping this. I want to do a 2.0 instead.

